PR #3, opened 2020-03-25
Fix typo in identify() docstring

casio:
Just had a first look into the package and stumbled upon this lil typo.


mariusandra:
Thanks @casio ! Merged :)

PR #8, opened 2020-09-09
Fix tests and clean up integration

timgl:
- Fixes tests
- Adds tests as github action
- removes 'screen'
- removes 'type'


yakkomajuri:
@timgl I'm wondering if we're even using `simulator.py` and if there's any intention to do so, so that we can either properly clean it up or get rid of it.

timgl:
it's a tool to debug, cleaned it up

PR #15, opened 2020-11-23
Do not start Feature Flag polling if no API Key

mariusandra:
## Changes

- To complement https://github.com/PostHog/posthog/issues/2150#issuecomment-732174233 and https://github.com/PostHog/posthog/pull/2473
- Moves the "no personal api key found" warning and `return` statement one level higher, thus never starting the poller if no api key found.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ env
 flake8.out
 pylint.out
 posthog-analytics
+.idea
+
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -228,11 +228,6 @@ class Client(object):
         self.join()
 
     def _load_feature_flags(self):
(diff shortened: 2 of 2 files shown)


Twixes:
Hmm, why does this use a personal API key instead of a project key? @timgl

timgl (approved):

timgl:
@Twixes b/c project keys don't have access to the feature flags (and we wouldn't want it too as that could be confidential experiments etc)

mariusandra:
I merged it... but now I have no idea how to release a new version onto pip. @timgl can you help here? I see @paolodamico is also a maintainer on pip.

paolodamico:
I can help with the release

PR #18, opened 2020-12-04
Add project_api_key, improve errors

yakkomajuri:
- Add a `project_api_key` config param to be used instead of `api_key` help minimize confusion on the type of key to use. This is backwards compatible with `api_key` and the docs will be updated accordingly
- Consolidate repeated code
- Handle a case when response is equal to `None`, leading to errors that aren't helpful such as:

```
error uploading: 'NoneType' object has no attribute 'status_code'
```


Twixes (approved):
This looks sensible to me.

PR #20, opened 2020-12-30
Fix post() returning None on non-200 response

### LINKED ISSUE #19: Client exit when /batch response return none 200 status code.
Client throw exception and exit with error when response status code is not 200. 

`Traceback (most recent call last):
  File "/usr/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/usr/lib/python3.8/threading.py", line 870, in run
    self._target(*self._args, **self._kwargs)
  File "/root/posthog/posthog-python/prime/generator/gen_specific.py", line 138, in _thread_send
    send_event(user_id, time_choose[randint(0, len(time_choose) - 1)], event_name)
  File "/root/posthog/posthog-python/prime/generator/gen_specific.py", line 127, in send_event
    posthog.capture(id_user, event_name, data, timestamp=datetime.fromtimestamp(timestamp))
  File "/root/posthog/posthog-python/posthog/__init__.py", line 46, in capture
    _proxy('capture', distinct_id=distinct_id, event=event, properties=properties, context=context, timestamp=timestamp, message_id=message_id)
  File "/root/posthog/posthog-python/posthog/__init__.py", line 161, in _proxy
    return fn(*args, **kwargs)
  File "/root/posthog/posthog-python/posthog/client.py", line 117, in capture
    return self._enqueue(msg)
  File "/root/posthog/posthog-python/posthog/client.py", line 192, in _enqueue
    batch_post(self.api_key, self.host, gzip=self.gzip,
  File "/root/posthog/posthog-python/posthog/request.py", line 69, in batch_post
    if res.status_code == 200:
AttributeError: 'NoneType' object has no attribute 'status_code'`

Hungsiro506:
Resolves #19.


Twixes:
Thanks for catching this! At the very least a code editor could have alerted us about this mismatch here when writing the module… If it were typed. It wasn't, so I just added that and hopefully we should be immune to such bugs as we sprinkle the codebase with types.

Twixes (approved):

PR #21, opened 2021-01-03
Black and isort all the things

Twixes:
This formats the code with black and isort with [pre-commit](https://pre-commit.com/), in accordance with the rest of our Python code base. `pip3 install pre-commit && pre-commit install` to enable hooks.

### DIFF
--- a/posthog/test/consumer.py
+++ b/posthog/test/consumer.py
@@ -1,20 +1,20 @@
-import unittest
-import mock
-import time
 import json
+import time
+import unittest
+
+import mock
 
 try:
     from queue import Queue
 except ImportError:
     from Queue import Queue
 
-from posthog.consumer import Consumer, MAX_MSG_SIZE
+from posthog.consumer import MAX_MSG_SIZE, Consumer
 from posthog.request import APIError
 from posthog.test.utils import TEST_API_KEY
 
 
 class TestConsumer(unittest.TestCase):
-
     def test_next(self):
         q = Queue()
         consumer = Consumer(q, '')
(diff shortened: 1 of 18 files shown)


posthog/test/consumer.py
```diff
@@ -131,8 +107,8 @@ def mock_post(*args, **kwargs):
                     self.assertEqual(exc, expected_exception)
                 else:
                     self.fail(
-                        "request() should raise an exception if still failing "
-                        "after %d retries" % consumer.retries)
+                        "request() should raise an exception if still failing " "after %d retries" % consumer.retries
```
mariusandra, line 110: this line seems a bit suspicious

posthog/utils.py
```diff
@@ -72,9 +70,10 @@ def _clean_dict(dict_):
             data[k] = clean(v)
         except TypeError:
             log.warning(
-                'Dictionary values must be serializeable to '
-                'JSON "%s" value %s of type %s is unsupported.',
-                k, v, type(v),
+                'Dictionary values must be serializeable to ' 'JSON "%s" value %s of type %s is unsupported.',
```
mariusandra, line 73: same with this double string situation

mariusandra (approved):
I guess it's fine! Found some funny/confusing string concatenation in the blacked output though. Not a blocker, just might be better to convert them to one string for clarity.

posthog/test/consumer.py
Twixes: Weird indeed! Melted them into one.

PR #22, opened 2021-01-26
Add condition for rollout-percentage

EDsCODE:
Quick patch for [this sentry error](https://sentry.io/organizations/posthog/issues/2178357897/?project=1899813) so the request will just be sent to the decide endpoint.
Need further investigation as to why that field isn't populated

### DIFF
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -296,7 +296,7 @@ class Client(object):
             except IndexError:
                 return default
 
-            if feature_flag.get("is_simple_flag"):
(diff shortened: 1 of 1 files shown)


Twixes (commented):
I think we may not need this since your changes @macobo?

macobo:
While this doesn't fix the original problem this is still a valid fix, since a flag can return simple and not have a rollout_percentage.

Twixes:
Alright, merging then.

PR #23, opened 2021-02-08
Add $set and $set_once support

paolodamico:
Adds `$set` and `$set_once` support to the library. Related to https://github.com/PostHog/posthog/issues/2684.

Further, makes things easier to test & run locally. Using `setup.py test` is deprecated, and so is `test_requires`. Thus, I changed the test file names and added support for pytest so we can still test things. I *didn't* change any CI configuration, which hopefully gives confidence that the new and old testing methods are on par.

If you're okay with this change, I can "upgrade" `ci.yaml` as well.


paolodamico:
First time taking a stab at this library, so a thorough review would be appreciated.

paolodamico:
Ping @macobo should we move this forward?

yakkomajuri:
If we're doing `$set_once`, shouldn't we also do `$set`?

Twixes:
Definitely agree with support `set` the same way.

posthog/test/test_client.py
```diff
@@ -125,7 +185,7 @@ def test_basic_page(self):
 
     def test_basic_page_distinct_uuid(self):
         client = self.client
-        distinct_id = uuid4()
+        distinct_id = str(uuid4())
```
neilkakkar, line 188: Not sure how this worked before O.o

simulator.py
```diff
@@ -71,6 +71,14 @@ def identify():
...
 
 
+def set_once():
+    posthog.set_once(
+        options.distinct_id,
+        properties=json_hash(options.traits),
+        context=json_hash(options.context),
+    )
+
+
```
mariusandra, line 81: Just a nit, but might be good to add "set" here for completion as well.

neilkakkar: Gotcha, makes sense

mariusandra (approved):

mariusandra:
Can you please still fix the black error... and then feel free to merge it into master :).

PR #26, opened 2021-04-01
Add UUID to ID_TYPES

### LINKED ISSUE #25: Library does not accept `uuid.UUID` as `distinct_id`, but Posthog uses UUIDs internally
This threw me for a loop for a second. I had to `str(myUUID)` to be able to use it as the `distinct_id` in all calls.

Since the documentation often shows UUIDs and doesn't explain that `distinct_id` should be a Number or `string_types`, it might be worth adding `uuid.UUID` to `ID_TYPES` at https://github.com/PostHog/posthog-python/blob/master/posthog/client.py#L24

Of course, one would need to cast the UUID to `str()` as well.

Twixes:
Makes `UUID` an accepted ID type. No changed in processing code needed, as `stringify_id` already `str()`-casts IDs. Resolves #25.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@ dist/
 MANIFEST
 build/
 .eggs/
-.coverage/
+.coverage
 .vscode/
 env/
 venv/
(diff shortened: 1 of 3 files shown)


paolodamico (approved):
lgtm

PR #29, opened 2021-05-13
Resolve polling issues with feature flags

### LINKED ISSUE #16: Feature flag loading blocks python scripts from finishing
See https://github.com/PostHog/posthog/issues/2150 for more info

neilkakkar:
Fixes #16 

Tested using `example.py`: it blocked when a valid personal API key was set, but doesn't anymore.

### DIFF
--- a/posthog/poller.py
+++ b/posthog/poller.py
@@ -4,7 +4,7 @@ import threading
 class Poller(threading.Thread):
     def __init__(self, interval, execute, *args, **kwargs):
         threading.Thread.__init__(self)
-        self.daemon = True # Make daemon to not interfere with program exit
+        self.daemon = True  # Make daemon to not interfere with program exit
         self.stopped = threading.Event()
         self.interval = interval
         self.execute = execute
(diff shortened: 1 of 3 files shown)


posthog/poller.py
```diff
@@ -4,7 +4,7 @@
 class Poller(threading.Thread):
     def __init__(self, interval, execute, *args, **kwargs):
         threading.Thread.__init__(self)
-        self.daemon = False
+        self.daemon = True  # Make daemon to not interfere with program exit
```
yakkomajuri, line 7: Do we need to enable this given we're now stopping the poller on exit?

yakkomajuri (commented):
Haven't worked in the codebase for a while, left one comment about flipping the `daemon` toggle

posthog/poller.py
neilkakkar: Indeed.

It's only when we manually shut down the runner, that the onExit part is called.

In other cases, given nothing is actively running, we don't want poll to actively block exit (which is what it does when it's not a daemon thread).
The program automatically exits when the only things running are daemon threads.

yakkomajuri (approved):

PR #32, opened 2021-05-18
Add support for project_api_key. 

### LINKED ISSUE #31: posthog.project_api_key doesn't work
According to the docs (https://posthog.com/docs/libraries/python#installation) `posthog.api_key` has been deprecated in favour of `posthog.project_api_key`.

However, when I use `project_api_key` with the current version, 1.3.1, it does not work, while `api_key` does.

neilkakkar:
Fixes #31

This doesn't remove `api_key` yet, but supports both `project_api_key` and `api_key`.

### DIFF
--- a/example.py
+++ b/example.py
@@ -6,8 +6,8 @@ import time
 import posthog
 
 # You can find this key on the /setup page in PostHog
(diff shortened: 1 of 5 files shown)


mariusandra (approved):
LGTM

Twixes (approved):
Should fix the reported issue, though still not great how we recommend the new param to users, but internally confusingly use the old one. Created #33 to keep track of this.

PR #36, opened 2021-05-24
Bump backoff dependency to 1.10.0

mands:
Bump backoff dependency as per analytics-python PR https://github.com/segmentio/analytics-python/pull/173.

Have fixed to 1.10.0 as per other PR - however may be better to keep as `>=1.6.0` or so for backwards compatibility.


Twixes (approved):
Looks good to me

PR #38, opened 2021-05-28
Remove pinning of backoff dependency

mands:
Have modified the backoff dependency so that it handles any future (minor & patch) releases - this helps with merging the library into conda/conda-forge - see https://github.com/conda-forge/staged-recipes/pull/15012


yakkomajuri (approved):
Seems fair to me.

Thanks for this and all the related work!

PR #40, opened 2021-06-04
Better handling on feature_enabled

yakkomajuri:
Was looking at the feature flags and came across a pretty bad pattern where we use an error for path handling in logic we have full control over.

### DIFF
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -330,9 +330,11 @@ class Client(object):
         if not self.feature_flags:
             response = default
         else:
-            try:
-                feature_flag = [flag for flag in self.feature_flags if flag["key"] == key][0]
-            except IndexError:
+            for flag in self.feature_flags:
+                if flag["key"] == key:
+                    feature_flag = flag
+                    break
+            else:
                 return default
 
             if feature_flag.get("is_simple_flag"):


posthog/client.py
```diff
@@ -330,9 +330,11 @@ def feature_enabled(self, key, distinct_id, default=False):
...
             response = default
         else:
-            try:
-                feature_flag = [flag for flag in self.feature_flags if flag["key"] == key][0]
-            except IndexError:
+            for flag in self.feature_flags:
+                if flag["key"] == key:
+                    feature_flag = flag
+                    break
+            else:
```
neilkakkar, line 337: Not sure if it's Github display, but this `else` seems indented incorrectly? 👀

yakkomajuri: Actually, this is correct, it's a Python for-else loop: https://docs.python.org/3/tutorial/controlflow.html#break-and-continue-statements-and-else-clauses-on-loop

yakkomajuri: Else will run if break doesn't

neilkakkar: Ah, fascinating! Wasn't aware this was possible :o

neilkakkar (approved):

PR #42, opened 2021-06-22
Add release workflow

Twixes:
This way any PostHog member will be able to release a new version of `posthog-python`, instead of having to become a PyPI collaborator (much more cumbersome).
CC @yakkomajuri


yakkomajuri:
Beautiful!

Twixes:
I had to make a series of tweaks but it works smoothly now.

Screen Shot 2021-06-22 at 14 46 35

PR #52, opened 2022-03-23
Avoid unclosed SSLSocket warnings

chamini2:
When using posthog, even when manually calling `posthog.shutdown()`, we always get a warning at the end:

```py
sys:1: ResourceWarning: unclosed <ssl.SSLSocket fd=10, family=AddressFamily.AF_INET, type=SocketKind.SOCK_STREAM, proto=0, laddr=('192.168.1.110', 59478), raddr=('34.227.250.33', 443)>
```

The warning only happens if we did send some messages (try calling `flush`, which is called by `shutdown` anyways, to make it show up).

### DIFF
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -10,7 +10,7 @@ from six import string_types
 
 from posthog.consumer import Consumer
 from posthog.poller import Poller
-from posthog.request import APIError, batch_post, decide, get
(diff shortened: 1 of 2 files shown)


chamini2:
Hello! Do you guys not get this behavior?

pauldambra:
sorry nobody replied here...

i'll close this since it's been nearly 3 years 🙈

PR #54, opened 2022-03-30
Remove "context"

mariusandra:
We ignore it and never use it, so let's remove. This is a remnant from the early life of this library.

### DIFF
--- a/posthog/__init__.py
+++ b/posthog/__init__.py
@@ -23,7 +23,6 @@ def capture(
     distinct_id,  # type: str
     event,  # type: str
     properties=None,  # type: Optional[Dict]
-    context=None,  # type: Optional[Dict]
     timestamp=None,  # type: Optional[datetime.datetime]
     uuid=None,  # type: Optional[str]
     groups=None,  # type: Optional[Dict]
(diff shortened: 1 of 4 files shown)


neilkakkar:
lol, was wondering about this today, I guess safe to remove, will do so 👌

haacked:
I resolved the merge conflicts. 

I support this change, but I do have a concern that it might break existing customer code that calls `capture`. For example, if someone were to have code like this:

```python
client.capture("some_distinct_id", "some-event", None, None, datetime.now(tzutc()), None, None, True)
```

When we remove the `context` argument, this call becomes invalid. Should we consider adding a deprecation warning log message before yanking this argument? Or is it very unlikely someone is calling this method with this many positional arguments. Hopefully folks are using named arguments once they get toward the end of the argument list. 😄

mariusandra:
Deprecating might indeed be safer, as we don't want to break people doing a patch or minor update. It'll remain a no-op like now, but at least it'll be clear that it's not to be used. Then we can remove it altogether after a few months or so...

haacked:
I can make that change if you'd like. I'll see if I can find other examples for how we deprecate things.

haacked:
I opened a PR https://github.com/PostHog/posthog-python/pull/182 to deprecate context. I'll close this PR and log an issue to remove context some time later.

haacked:
Issue created here: https://github.com/PostHog/posthog-python/issues/183

PR #56, opened 2022-04-25
fix: Personal api key not required

timgl:
Most feature flags require a call to decide anyway, so we might as well allow feature flags without first grabbing all active feature flags.

### DIFF
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -364,10 +364,8 @@ class Client(object):
                 if flag["key"] == key:
                     feature_flag = flag
                     if feature_flag.get("is_simple_flag"):
-                        response = _hash(key, distinct_id) <= (
-                            (feature_flag.get("rollout_percentage", 100) or 100) / 100
-                        )
-        if not response:
+                        response = _hash(key, distinct_id) <= (feature_flag.get("rollout_percentage", 100) / 100)
+        if response == None:
             try:
                 request_data = {
                     "distinct_id": distinct_id,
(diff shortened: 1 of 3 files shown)


posthog/client.py
```diff
@@ -354,39 +354,38 @@ def feature_enabled(self, key, distinct_id, default=False, *, groups={}):
...
-            else:
-                return default
-
-            if feature_flag.get("is_simple_flag"):
-                response = _hash(key, distinct_id) <= ((feature_flag.get("rollout_percentage", 100) or 100) / 100)
+                    if feature_flag.get("is_simple_flag"):
+                        response = _hash(key, distinct_id) <= (
+                            (feature_flag.get("rollout_percentage", 100) or 100) / 100
+                        )
+        if not response:
```
mariusandra, line 370: If it's a simple flag, yet doesn't match (response is `False`), won't we make a needless `decide` call? Shouldn't this compare with `None` instead?

posthog/client.py
```diff
@@ -354,39 +354,38 @@ def feature_enabled(self, key, distinct_id, default=False, *, groups={}):
         require("distinct_id", distinct_id, ID_TYPES)
         require("groups", groups, dict)
 
-        if not self.personal_api_key:
-            self.log.warning("[FEATURE FLAGS] You have to specify a personal_api_key to use feature flags.")
```
mariusandra, line 358: This error was helpful. It'll now come from the API?

mariusandra (commented):
two comments

posthog/client.py
timgl: Yep good point, will test

mariusandra (approved):

PR #60, opened 2022-05-12
Feature: Enable multi variate feature flags for Python library

utkuzih:
Using the v2 decide endpoint. Should be merged after https://github.com/PostHog/posthog-python/pull/57 as that pull request fixes how feature flags are captured on Posthog.


utkuzih:
For what its worth, I have accidentally created the #59 on the wrong account. Closed that one and re-created on the account which owns the forked repository.

PR #64, opened 2022-05-24
Have an option to send feature variants with the .capture(...) calls.

utkuzih:
Implemented `send_feature_flags=False` flag for the `posthog.capture(...)` call. Also added a new public function similar to the Javascript library: `posthog.get_feature_variants(distinct_id, groups=None)`.

### DIFF
--- a/posthog/test/test_client.py
+++ b/posthog/test/test_client.py
@@ -417,7 +417,7 @@ class TestClient(unittest.TestCase):
     @mock.patch("posthog.client.decide")
     @mock.patch("posthog.client.get")
     def test_feature_enabled_simple_is_false(self, patch_get, patch_decide):
-        client = Client(TEST_API_KEY, personal_api_key="test")
+        client = Client(TEST_API_KEY)
         client.feature_flags = [
             {"id": 1, "name": "Beta Feature", "key": "beta-feature", "is_simple_flag": True, "rollout_percentage": 0}
         ]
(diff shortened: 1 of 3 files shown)


posthog/test/test_client.py
```diff
@@ -417,7 +417,7 @@ def test_feature_enabled_simple(self, patch_get, patch_decide):
     @mock.patch("posthog.client.decide")
     @mock.patch("posthog.client.get")
     def test_feature_enabled_simple_is_false(self, patch_get, patch_decide):
-        client = Client(TEST_API_KEY)
+        client = Client(TEST_API_KEY, personal_api_key="test")
```
neilkakkar, line 420: why'd you have to add this here? 🤔

utkuzih: Ah ha, I was blindly adding the `personal_api_key` to all tests which mocked `client.decide`. No need in here.

neilkakkar:
Hey @utkuzih , apologies for taking a while. I wanted to confirm that the solution for https://github.com/PostHog/posthog/issues/9547 wouldn't impact this in any way. Doesn't seem like it anymore, so I'll get this merged first thing Monday :)

neilkakkar:
It would be nice to have a test or two for the new functionality as well. I'll add these myself on Monday if you don't get to it by then.

neilkakkar:
I couldn't push to your fork, so created a new PR instead 😅 . Thanks for working on this!

utkuzih:
@neilkakkar Thank you!

PR #67, opened 2022-06-27
Add get feature flag method

liyiy:
Add get feature flag method similar to https://github.com/PostHog/posthog-js/blob/4338bb740fda9b31770834ed3eff2a47d910b798/src/posthog-featureflags.js#L145


neilkakkar (approved):
Needs an add in `__init__.py` but otherwise LGTM!

PR #68, opened 2022-07-25
feat(feature-flags): Enable local evaluation of flags

neilkakkar:
Following spec from https://github.com/PostHog/posthog/issues/10459 

The PR isn't really that big, ~3,000 LoC are tests, 2,000 of which are autogenerated consistency tests.

(but still big enough, sigh). Wanted to figure out exactly what all we need for the new library to be feature complete, and use this as a template for other libraries to follow.


EDsCODE:
Update:

- Applied black and isort so the line diffs exploded but it's just because of the result set in the test
- Adjust simple flag handling so that existing tests still pass. Feel free to change this as necessary
- Adjusted return of "get_feature_flag" so that it still returns variants and not just booleans
- Added group property evaluation
- Added tests for person property eval and group property eval

neilkakkar:
Thanks!

I was planning to not have any deprecated options in the library, since we control the API which returns flag definitions, and we can transform all deprecated ones into new ones. Effectively, I want to deprecate `is_simple_flag` evaluation here, since and only have one code path that's exactly the same as the main logic on our servers.

neilkakkar:
That would probs imply a major version update, so self-hosted know not to upgrade _unless_ they're on PostHog version 1.38, but I think that's a good trade-off to make, since it means a little less code in every client library, and a clean cut w.r.t the API endpoints used to support local eval. (see main PostHog PR, very much WIP: https://github.com/PostHog/posthog/pull/10957 )

posthog/client.py
```diff
@@ -398,44 +414,39 @@ def feature_enabled(self, key, distinct_id, default=False, *, groups={}):
...
-                            if feature_flag.get("rollout_percentage") is not None
-                            else 100
-                        )
-                        response = _hash(key, distinct_id) <= (rollout_percentage / 100)
-        if response == None:
+                    try:
+                        flag_filters = feature_flag.get("filters") or {}
+                        aggregation_group_type_index = flag_filters.get("aggregation_group_type_index")
+                        if aggregation_group_type_index is not None:
+                            focused_group_properties = group_properties[
```
neilkakkar, line 421: This is tricky. We should hash out how we expect users to add properties here.

I think what you're doing makes sense, specially when different groups can have the same `name` property. However at the same time, feature flags support a single group at a time, so we'd expect only properties from a single group per call..

This is more future proof, but also a bit annoying for current user needs.

posthog/test/test_client.py
```diff
@@ -439,156 +450,23 @@ def mock_post_fn(*args, **kwargs):
...
 
     def test_default_timeout_15(self):
-        client = Client(TEST_API_KEY)
+        client = Client(FAKE_TEST_API_KEY)
         for consumer in client.consumers:
             self.assertEqual(consumer.timeout, 15)
 
-    @mock.patch("posthog.client.Poller")
-    @mock.patch("posthog.client.get")
-    def test_load_feature_flags(self, patch_get, patch_poll):
```
neilkakkar, line 453: these moved to `test_feature_flags.py`

neilkakkar:
A better handle on what the tests are doing: https://github.com/PostHog/posthog/issues/10459#issuecomment-1193842228 (scroll to the bottom) - I've added tests we'd need in common between all libraries

posthog/client.py
EDsCODE: EIther works for me. When only one group is in focus, how does this function know which group is being used?

neilkakkar: `aggregation_group_type_index` ? Not sure I get the question

EDsCODE: Oh, so upon calling posthog.get_feature_flag() you're assumed to already know what group type the flag in question is, so you only pass in the group properties related to that group?

neilkakkar: Yeah pretty much, since they're the ones who defined the flag

neilkakkar: Ah, thought about this a bit more, and I think your format makes sense, since we'll want to send these properties to `/decide` as well, when we do fallback, so that these props can be overridden.

neilkakkar: And /decide evaluates multiple flags, and we don't know which of those have which groups

EDsCODE (approved):
great work!

posthog/test/test_feature_flags.py
```diff
@@ -0,0 +1,3078 @@
...
+        self.assertEqual(patch_poll.call_count, 1)
+
+    def test_load_feature_flags_wrong_key(self):
+        client = Client(FAKE_TEST_API_KEY, personal_api_key=FAKE_TEST_API_KEY)
+        with freeze_time("2020-01-01T12:01:00.0000Z"):
+            self.assertRaises(APIError, client.load_feature_flags)
+
+    @mock.patch("posthog.client.decide")
+    @mock.patch("posthog.client.get")
+    def test_feature_enabled_simple(self, patch_get, patch_decide):
```
neilkakkar, line 540: this and following tests were copied over from the client, so not necessary to write again in every other lib (these libs should have their own old tests 🤷).

PR #75, opened 2022-09-14
fix: datetime comparison issues

neilkakkar:
Have gone for a naive datetime approach here, so we disregard timezone information. Not sure if we want to handle this / if there's valid usecases for this yet.

### DIFF
--- a/posthog/feature_flags.py
+++ b/posthog/feature_flags.py
@@ -4,7 +4,7 @@ import re
 
 from dateutil import parser
 
-from posthog.utils import is_valid_regex
+from posthog.utils import convert_to_datetime_aware, is_valid_regex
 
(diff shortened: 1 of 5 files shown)


posthog/feature_flags.py
```diff
@@ -131,18 +131,24 @@ def match_property(property, property_values) -> bool:
 
     if operator in ["is_date_before", "is_date_after"]:
         try:
-            parsed_date = parser.parse(value)
+            parsed_date = parser.parse(value, ignoretz=True)
```
EDsCODE, line 134: Hmmm I think this could be confusing because you're not restricted to setting date properties in a certain way on the app so you wouldn't expect a timezone you set to just be removed.

neilkakkar: Fair. We could make everything tz aware and default to utc if none known?

PR #81, opened 2023-01-19
feat(feature-flags): JSON payload function

EDsCODE:
- implements https://github.com/PostHog/posthog/issues/13627 for posthog-python
     - get_feature_flag_payload
     - get_all_flags_and_payloads
     - change api to v=3

### DIFF
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -368,9 +368,6 @@ class Client(object):
 
     def _load_feature_flags(self):
         try:
-            # Don't try endpoint at all if personal_api_key is None
-            if self.personal_api_key is None:
-                raise APIError(status=401)
 
             response = get(
                 self.personal_api_key,
@@ -378,12 +375,12 @@ class Client(object):
                 self.host,
                 timeout=10,
             )
-            if response.status_code == 200:
-                self.feature_flags = response["flags"] or []
-                self.feature_flags_by_key = {
-                    flag["key"]: flag for flag in self.feature_flags if flag.get("key") is not None
-                }
-                self.group_type_mapping = response["group_type_mapping"] or {}
+
+            self.feature_flags = response["flags"] or []
+            self.feature_flags_by_key = {
+                flag["key"]: flag for flag in self.feature_flags if flag.get("key") is not None
+            }
+            self.group_type_mapping = response["group_type_mapping"] or {}
 
         except APIError as e:
             if e.status == 401:
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -368,14 +368,20 @@ class Client(object):
 
     def _load_feature_flags(self):
         try:
+            # Don't try endpoint at all if personal_api_key is None
+            if self.personal_api_key is None:
+                raise APIError(status=401)
+
             response = get(
-                self.personal_api_key, f"/api/feature_flag/local_evaluation/?token={self.api_key}", self.host
+                self.personal_api_key, f"/api/feature_flag/local_evaluation/?token={self.api_key}", self.host,
+                timeout=10
             )
-            self.feature_flags = response["flags"] or []
-            self.feature_flags_by_key = {
-                flag["key"]: flag for flag in self.feature_flags if flag.get("key") is not None
-            }
-            self.group_type_mapping = response["group_type_mapping"] or {}
+            if response.status_code == 200:
+                self.feature_flags = response["flags"] or []
+                self.feature_flags_by_key = {
+                    flag["key"]: flag for flag in self.feature_flags if flag.get("key") is not None
+                }
+                self.group_type_mapping = response["group_type_mapping"] or {}
 
         except APIError as e:
             if e.status == 401:
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -561,48 +561,25 @@ class Client(object):
         response = None
 
         if match_value is not None and self.feature_flags_by_key:
-            flag_definition = self.feature_flags_by_key.get(key) or {}
-            flag_filters = flag_definition.get("filters") or {}
-            flag_payloads = flag_filters.get("payloads") or {}
-            response = flag_payloads.get(str(match_value).lower(), None)
+            response = self._get_feature_flag_payload(key, match_value)
 
         if response is None and not only_evaluate_locally:
             decide_payloads = self.get_feature_payloads(distinct_id, groups, person_properties, group_properties)
             response = decide_payloads.get(str(key).lower(), None)
 
         return response
+    
+    def _get_feature_flag_payload(self, key, match_value):
+        flag_definition = self.feature_flags_by_key.get(key) or {}
+        flag_filters = flag_definition.get("filters") or {}
+        flag_payloads = flag_filters.get("payloads") or {}
+        payload = flag_payloads.get(str(match_value).lower(), None)
+        return payload
 
     def get_all_flags(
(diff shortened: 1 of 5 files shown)


posthog/client.py
```diff
@@ -127,6 +129,14 @@ def identify(self, distinct_id=None, properties=None, context=None, timestamp=No
...
 
     def get_feature_variants(self, distinct_id, groups=None, person_properties=None, group_properties=None):
+        resp_data = self.get_feature(distinct_id, groups, person_properties, group_properties)
+        return resp_data["featureFlags"]
+
+    def get_feature_payloads(self, distinct_id, groups=None, person_properties=None, group_properties=None):
+        resp_data = self.get_feature(distinct_id, groups, person_properties, group_properties)
+        return resp_data["featureFlagPayloads"]
+
+    def get_feature(self, distinct_id, groups=None, person_properties=None, group_properties=None):
```
neilkakkar, line 139: nit: I'd call this `get_decide`

posthog/client.py
```diff
@@ -522,48 +535,121 @@ def get_feature_flag(
             self.distinct_ids_feature_flags_reported[distinct_id].add(feature_flag_reported_key)
         return response
 
+    def get_feature_flag_payload(
```
neilkakkar, line 538: this needs to go in `__init__.py` for the interface to be available.

Also add to `example.py` to test locally that everything works :)

posthog/client.py
```diff
@@ -522,48 +535,121 @@ def get_feature_flag(
...
+                person_properties=person_properties,
+                group_properties=group_properties,
+                send_feature_flag_events=send_feature_flag_events,
+                only_evaluate_locally=True,
+            )
+
+        response = None
+
+        if match_value is not None:
+            response = self._get_feature_flag_payload(key, match_value)
```
neilkakkar, line 564: this is a confusing name (differs by main with just an underscore). Can we call this `compute_feature_flag_payload_locally` instead?

neilkakkar: or jsut `compute_payload_locally`

posthog/request.py
```diff
@@ -63,7 +63,7 @@ def _process_response(
 
 def decide(api_key: str, host: Optional[str] = None, gzip: bool = False, timeout: int = 15, **kwargs) -> Any:
     """Post the `kwargs to the decide API endpoint"""
-    res = post(api_key, host, "/decide/?v=2", gzip, timeout, **kwargs)
+    res = post(api_key, host, "/decide/?v=3", gzip, timeout, **kwargs)
```
neilkakkar, line 66: is the resiliency stuff in a separate pr? 😅 (I'm hoping you're handling that too 😬 )

posthog/client.py
```diff
@@ -522,48 +535,121 @@ def get_feature_flag(
...
+        flag_definition = self.feature_flags_by_key.get(key) or {}
+        flag_filters = flag_definition.get("filters") or {}
+        flag_payloads = flag_filters.get("payloads") or {}
+        payload = flag_payloads.get(str(match_value).lower(), None)
+        return payload
+
     def get_all_flags(
         self, distinct_id, *, groups={}, person_properties={}, group_properties={}, only_evaluate_locally=False
     ):
+        response, _, fallback_to_decide = self._get_all_flags_and_payloads_locally(
```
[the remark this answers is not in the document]
liyiy: simplify this how? i'm about to just copy pasta for ruby so..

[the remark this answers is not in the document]
liyiy: you mean this one? https://github.com/PostHog/posthog-js-lite/pull/59 there's a million PRs just for payloads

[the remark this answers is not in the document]
liyiy: and none of them indicate that it's meant to simplify and improve the interface, just what it does code wise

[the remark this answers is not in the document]
neilkakkar: Now you know why I couldn't point to a specific PR 😂 .

Let me try and remember what the issue was here lol.....

Ah yes:

1. You don't need to duplicate code between `get_all_flags` and `get_all_payloads`. Instead, the interface will be `get_all_flags` and `get_all_flags_and_payloads`, and get_all_flags will simply use the result of get_all_flags_and_payloads.
2. You can get flags and payloads in one decide call, doesn't make sense to have 2 decide calls.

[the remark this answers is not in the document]
liyiy: :P TY

posthog/request.py
EDsCODE: not anymore! 😈

neilkakkar: Sounds good to me!

liyiy (approved):
seems good to me, i'll let @neilkakkar be the final reviewer on this 🤗

posthog/client.py
```diff
@@ -522,48 +540,117 @@ def get_feature_flag(
...
+            distinct_id, groups=groups, person_properties=person_properties, group_properties=group_properties
+        )
+
+        if fallback_to_decide and not only_evaluate_locally:
+            try:
+                flags_and_payloads = self.get_decide(
+                    distinct_id, groups=groups, person_properties=person_properties, group_properties=group_properties
+                )
+                flags = {**flags, **flags_and_payloads["featureFlags"]}
+                payloads = {**payloads, **flags_and_payloads["featureFlagPayloads"]}
```
liyiy, line 614: if we're strictly falling back to decide here why are we merging payloads and not just using `flags_and_payloads`?

posthog/client.py
```diff
@@ -64,6 +64,8 @@ def __init__(
         self.gzip = gzip
         self.timeout = timeout
         self.feature_flags = None
+        self.feature_flags_by_key = None
+        self.feature_flag_payloads = None
```
liyiy, line 68: This doesn't look like it's being used anywhere? :o

posthog/test/test_feature_flags.py
```diff
@@ -1149,6 +1373,122 @@ def test_flag_with_multiple_variant_overrides(self, patch_decide):
...
+            300,
+        )
+
+        self.assertEqual(
+            self.client.get_feature_flag_payload(
+                "person-flag", "some-distinct-id", match_value=True, person_properties={"region": "USA"}
+            ),
+            300,
+        )
+        self.assertEqual(patch_decide.call_count, 2)
```
liyiy, line 1435: how is this calling decide twice? It doesn't in my spec 🤔

neilkakkar: because there are no saved flags / local evaluation is off / no personal API key.

So, we don't know what match_value=True corresponds to, flag wise and need to call decide.

Maybe your test has local eval on? In which case it ought to be 0 calls.

posthog/client.py
neilkakkar: we needed that earlier when decide didn't return all things (just enabled flags), but yeah, agreed, now we can just return what decide returns :)

posthog/test/test_feature_flags.py
liyiy: Another q: What about a test when decide returns multivariate payloads?

EDsCODE: That code path should be identical because a decide-based return for payloads reads from `featureFlagPayloads` key which should contains <flag-key>: payload mapping. There's nothing that distinguishes multivariate flag payloads here. The primary difference exists on local evaluation return because of how the payloads are defined

PR #82, opened 2023-01-24
Update alias docs

lharries:
Update the alias call to match how aliasing actually works. It was the wrong order


posthog/__init__.py
```diff
@@ -214,18 +214,25 @@ def alias(
...
     For example:
     ```python
-    posthog.alias('anonymous session id', 'distinct id')
+    posthog.alias('distinct id', 'anonymous session id')
+    ```
+
+    or
+
+    ```python
+    posthog.alias('distinct id', 'users-email@posthog.com')
```
tiina303, line 229: this looks like a bad example. The `users-email@posthog.com` doesn't look like an anonymous session ID to me

posthog/client.py
```diff
@@ -241,21 +241,21 @@ def group_identify(self, group_type=None, group_key=None, properties=None, conte
...
-                "distinct_id": previous_id,
-                "alias": distinct_id,
+                "distinct_id": distinct_id,
+                "alias": previous_id,
             },
             "timestamp": timestamp,
             "context": context,
             "event": "$create_alias",
-            "distinct_id": previous_id,
+            "distinct_id": distinct_id,
```
tiina303, line 258: why are we swapping the order here?

example.py
```diff
@@ -41,9 +41,9 @@
 
 print(posthog.feature_enabled("beta-feature", "distinct_id"))
 
-# # Alias a previous distinct id with a new one
+# Add an alias_id to a distinct_id
```
tiina303, line 44: This is not clear to me what alias_id restrictions are.

posthog/client.py
lharries: In our talks we tell people to use the ordering posthog.alias(distinct_id, alias). However, the parameter names are the wrong way round for this so I've updated the parameter names. Even in the current version we're actually sending it the right way (but in a fairly hacky way where we send the distinct_id as the alias because the user actually gives the alias in place of the distinct_id)

example.py
lharries: I thought we are removing the alias restrictions with the new updates of persons on events enabling person merging?

pauldambra:
let's assume this is out of date enough that we can close it 

sorry @lharries 😅 (i'm guessing you don't mind after all this time)

PR #96, opened 2023-04-04
chore: change package to be an instantiable client

EDsCODE:
Using global client is setting up for failure if there are overlapping posthogs in a dependency and the project using the depency

Changes:
- wrap the function headers in __init__.py to be an object that can be instantiated

Note: This is a major version change as it breaks the current pattern

### DIFF
--- a/posthog/__init__.py
+++ b/posthog/__init__.py
@@ -9,6 +9,7 @@ __version__ = VERSION
 
 default_client = None
 
+
 class Posthog(object):
     log = logging.getLogger("posthog")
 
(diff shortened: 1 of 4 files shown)


neilkakkar:
I was thinking if it makes sense to support both & be backwards compatible? Since any new lib using the instantiable version would mean there are no clashes, and old ones can keep working as is.

No strong feelings for either

posthog/__init__.py
```diff
@@ -3,423 +3,427 @@
...
-        return None
-    if not default_client:
-        default_client = Client(
-            api_key,
+class Posthog(object):
+    log = logging.getLogger("posthog")
+
+    """Settings."""
+    disabled = False  # type: bool
+    _client: Client = None  # type: Client
```
neilkakkar, line 18: We don't need a Client proxy anymore, just inherit from Client

neilkakkar: Right now it's a bit annoying because we _still_ have to remember to add new functions here for it to work

EDsCODE: Fair point. I left it to start because it seems to be a good way to bound the functions we want exposed

EDsCODE:
> I was thinking if it makes sense to support both & be backwards compatible? Since any new lib using the instantiable version would mean there are no clashes, and old ones can keep working as is.
> 
> No strong feelings for either

How do we leave a global while still solving the issue of multiple conflicting libraries? Seems like the only option would be to create a `Client` everytime `_proxy` is run

posthog/__init__.py
```diff
@@ -423,3 +423,6 @@ def _proxy(method, *args, **kwargs):
 
     fn = getattr(default_client, method)
     return fn(*args, **kwargs)
+
+class Posthog(Client):
```
EDsCODE, line 427: @neilkakkar Or maybe something like this where the global client situation still exists but we'll update docs to use this new pattern so new usage won't cause collisions

neilkakkar: yep exactly what I meant!

posthog/__init__.py
```diff
@@ -423,3 +423,7 @@ def _proxy(method, *args, **kwargs):
 
     fn = getattr(default_client, method)
     return fn(*args, **kwargs)
+
+
+class Posthog(Client):
+    pass
```
neilkakkar, line 429: As it stands, I think the `disabled` setting wouldn't work, rest seem good!

Fine with keeping it like so, as long as we mention in the docs

neilkakkar (approved):

posthog/__init__.py
```diff
@@ -406,8 +406,6 @@ def shutdown():
 def _proxy(method, *args, **kwargs):
     """Create an analytics client if one doesn't exist and send to it."""
     global default_client
-    if disabled:
```
neilkakkar, line 409: Breaking behaviour change here:

Earlier, if you did `posthog.capture()`, then `posthog.disabled = True`, then `posthog.capture()`, the second capture wouldn't fire, but now it would, since the initialised client has disabled = False.

Can do something like:

```
if disabled:
   default_client.disable()
else:
    default_client.enable()
```

and move it below the null check for `default_client`

EDsCODE: done

PR #98, opened 2023-04-12
Disable geoip for capture and decide calls by default

liyiy:
https://github.com/PostHog/posthog/issues/14351

The `$geoip` property on server libraries tends to default to incorrect locations (mainly based from where the server is located and not the user) and causes issues such as incorrect event/user data for analytics and problems with feature flag rollouts based on location.

This fix prevents the `advanced geoip plugin` and the `decide` call in https://github.com/PostHog/posthog/blob/master/posthog/api/decide.py#L144 from overriding user properties with incorrect server library geoip properties by default.

Users who are manually passing in geoip values will have to set `geoip_disable` to `False` to ensure their values are not discarded

related PR: https://github.com/PostHog/posthog/pull/15039

### DIFF
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -252,7 +252,16 @@ class Client(object):
 
         return self._enqueue(msg, geoip_disable)
 
-    def group_identify(self, group_type=None, group_key=None, properties=None, context=None, timestamp=None, uuid=None, geoip_disable=None):
+    def group_identify(
+        self,
+        group_type=None,
+        group_key=None,
+        properties=None,
+        context=None,
+        timestamp=None,
+        uuid=None,
+        geoip_disable=None,
+    ):
         properties = properties or {}
         context = context or {}
         require("group_type", group_type, ID_TYPES)
@@ -293,7 +302,9 @@ class Client(object):
 
         return self._enqueue(msg, geoip_disable)
 
-    def page(self, distinct_id=None, url=None, properties=None, context=None, timestamp=None, uuid=None, geoip_disable=None):
+    def page(
+        self, distinct_id=None, url=None, properties=None, context=None, timestamp=None, uuid=None, geoip_disable=None
+    ):
         properties = properties or {}
         context = context or {}
 
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -171,6 +171,7 @@ class Client(object):
         uuid=None,
         groups=None,
         send_feature_flags=False,
+        geoip_disable=True,
     ):
         properties = properties or {}
         context = context or {}
@@ -200,6 +201,9 @@ class Client(object):
                 for feature, variant in feature_variants.items():
                     msg["properties"]["$feature/{}".format(feature)] = variant
                 msg["properties"]["$active_feature_flags"] = list(feature_variants.keys())
+
+        if geoip_disable:
+            msg["properties"]["$geoip_disable"] = True
 
         return self._enqueue(msg)
 
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -48,7 +48,7 @@ class Client(object):
         personal_api_key=None,
         project_api_key=None,
         disabled=False,
-        geoip_disable=True,
+        geoip_disable=None,
     ):
         self.queue = queue.Queue(max_queue_size)
 
@@ -114,7 +114,7 @@ class Client(object):
                 if send:
                     consumer.start()
 
-    def identify(self, distinct_id=None, properties=None, context=None, timestamp=None, uuid=None):
+    def identify(self, distinct_id=None, properties=None, context=None, timestamp=None, uuid=None, geoip_disable=None):
         properties = properties or {}
         context = context or {}
         require("distinct_id", distinct_id, ID_TYPES)
@@ -129,16 +129,16 @@ class Client(object):
             "uuid": uuid,
         }
 
-        return self._enqueue(msg)
+        return self._enqueue(msg, geoip_disable)
 
     def get_feature_variants(
-        self, distinct_id, groups=None, person_properties=None, group_properties=None, geoip_disable=True
+        self, distinct_id, groups=None, person_properties=None, group_properties=None, geoip_disable=None
     ):
         resp_data = self.get_decide(distinct_id, groups, person_properties, group_properties, geoip_disable)
         return resp_data["featureFlags"]
 
     def _get_active_feature_variants(
-        self, distinct_id, groups=None, person_properties=None, group_properties=None, geoip_disable=True
+        self, distinct_id, groups=None, person_properties=None, group_properties=None, geoip_disable=None
     ):
         feature_variants = self.get_feature_variants(
             distinct_id, groups, person_properties, group_properties, geoip_disable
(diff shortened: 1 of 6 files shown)


posthog/client.py
```diff
@@ -201,6 +212,9 @@ def capture(
                     msg["properties"]["$feature/{}".format(feature)] = variant
                 msg["properties"]["$active_feature_flags"] = list(feature_variants.keys())
 
+        if self.geoip_disable:
```
neilkakkar, line 215: Hmm, this doesn't quite work because other SDK functions, like `identify`, `alias`, `set`, don't call capture but call `_enqueue` directly.

Also, this overrides if the user sent in a property like `{"$geoip_disable": False}` which isn't what we want.

So, I'd say, move this into logic `enqueue`, and add the property if it doesn't already exist.

posthog/test/test_client.py
```diff
@@ -494,6 +494,11 @@ def test_enabled_to_disabled(self):
 
         self.assertEqual(msg, "disabled")
 
+    def test_geoip_disable(self):
+        client = Client(FAKE_TEST_API_KEY, on_error=self.set_fail, geoip_disable=True)
+        _, msg = client.capture("distinct_id", "python test event")
```
neilkakkar, line 499: test where user sends in the property too

posthog/client.py
```diff
@@ -129,21 +131,29 @@ def identify(self, distinct_id=None, properties=None, context=None, timestamp=No
...
-        resp_data = self.get_decide(distinct_id, groups, person_properties, group_properties)
+    def get_feature_variants(
+        self, distinct_id, groups=None, person_properties=None, group_properties=None, geoip_disable=True
+    ):
+        resp_data = self.get_decide(distinct_id, groups, person_properties, group_properties, geoip_disable)
         return resp_data["featureFlags"]
 
-    def _get_active_feature_variants(self, distinct_id, groups=None, person_properties=None, group_properties=None):
-        feature_variants = self.get_feature_variants(distinct_id, groups, person_properties, group_properties)
+    def _get_active_feature_variants(
```
neilkakkar, line 140: this one should not default to true when the SDK level setting is false, which it does right now because we didn't add geoip_disable as a parameter to the call site.

I think it might be easier to work with these here if you default to `False` everywhere, explicitly input true when SDK level setting is set, and then test that this message is added to all capture/flag calls. Same for `get_feature_variants` and `get_feature_payloads` and `get_decide`

posthog/test/test_client.py
```diff
@@ -112,7 +112,7 @@ def test_get_active_feature_flags(self, patch_decide):
         }
 
         client = Client(FAKE_TEST_API_KEY, on_error=self.set_fail, personal_api_key=FAKE_TEST_API_KEY)
-        variants = client._get_active_feature_variants("some_id", None, None, None)
+        variants = client._get_active_feature_variants("some_id", None, None, None, True)
```
neilkakkar, line 115: Here, assert decide was called with the 'right' geoip_disable value.

Also add a similar test for capture, where you're not manually calling `client._get_active_feature_variants` but let capture call it

posthog/client.py
```diff
@@ -470,6 +484,7 @@ def feature_enabled(
         group_properties={},
         only_evaluate_locally=False,
         send_feature_flag_events=True,
+        geoip_disable=True,
```
neilkakkar, line 487: These defaults should also come from the SDK at the top level?

So, should be `None` default here I'd say, so a user can override it. If it's not `None`, pass on given value below. If it's `None`, use the SDK default (`self.geoip_disable` )

neilkakkar:
> Users who are manually passing in geoip values will have to set geoip_disable to False to ensure their values are not discarded

This makes for an annoying user experience. I don't think we should have to force users to remember to set this setting, to make sure that their properties behave as expected.

Rather, we have very clear priority ordering here: if the user inputs anything, use that; if not, use the defaults. My previous comments in code are working towards this ideology.

liyiy:
> > Users who are manually passing in geoip values will have to set geoip_disable to False to ensure their values are not discarded
> 
> This makes for an annoying user experience. I don't think we should have to force users to remember to set this setting, to make sure that their properties behave as expected.
> 
> Rather, we have very clear priority ordering here: if the user inputs anything, use that; if not, use the defaults. My previous comments in code are working towards this ideology.

Right, but the default you're asking for is that it disables geoip by default.. so users will still have to manually set geoip_disable to `False` no?

posthog/client.py
```diff
@@ -327,6 +353,11 @@ def _enqueue(self, msg):
         msg["properties"]["$lib"] = "posthog-python"
         msg["properties"]["$lib_version"] = VERSION
 
+        if geoip_disable is None:
+            geoip_disable = self.geoip_disable
+            if geoip_disable is True:
```
neilkakkar, line 358:
```suggestion
            if geoip_disable:
```

posthog/test/test_client.py
```diff
@@ -494,6 +506,22 @@ def test_enabled_to_disabled(self):
...
+
+        _, identify_msg = client.identify("distinct_id", {"trait": "value"})
+        client.flush()
+        self.assertEqual(identify_msg["properties"]["$geoip_disable"], True)
+
+    def test_geoip_disable_method_overrides_init_on_events(self):
+        client = Client(FAKE_TEST_API_KEY, on_error=self.set_fail, geoip_disable=True)
+        _, msg = client.capture("distinct_id", "python test event", geoip_disable=False)
+        client.flush()
+        self.assertTrue("$geoip_disable" not in msg["properties"])
```
neilkakkar, line 523: we need tests for the flag functions too!

And one for `_active_feature_flag` on capture (i.e. when all flags are attached to an event)

neilkakkar (commented):
Looking good!

posthog/test/test_client.py
neilkakkar: oh and one more case in this current test here where we initialise with False, pass in geoip_disable=True and see it in properties.

posthog/client.py
liyiy: i thought we want to test just in case someone passes in a string of "True" or something

posthog/test/test_client.py
liyiy: > we need tests for the flag functions too!

all of them?! 😱

neilkakkar: smallest possible that covers all code paths, so nope probably not all, anything that goes to `get_decide` should work

posthog/client.py
neilkakkar: The property being set is `True`, so if they make a mistake and pass in a string here, doesn't affect us, ok to accept.

But I guess the reverse argument applies to the `"False"` case. Upto you

posthog/client.py
```diff
@@ -127,25 +129,36 @@ def identify(self, distinct_id=None, properties=None, context=None, timestamp=No
...
-        resp_data = self.get_decide(distinct_id, groups, person_properties, group_properties)
+    def get_feature_variants(
+        self, distinct_id, groups=None, person_properties=None, group_properties=None, geoip_disable=None
+    ):
+        resp_data = self.get_decide(distinct_id, groups, person_properties, group_properties, geoip_disable)
         return resp_data["featureFlags"]
 
-    def _get_active_feature_variants(self, distinct_id, groups=None, person_properties=None, group_properties=None):
-        feature_variants = self.get_feature_variants(distinct_id, groups, person_properties, group_properties)
+    def _get_active_feature_variants(
```
neilkakkar, line 140: Don't think this works for the case when someone calls `posthog.capture(send_feature_flags=True, geoip_disable=False)`, or is tested yet?

liyiy: huh? what exactly would that test? If we're checking to see if setting a method `geoip_disable` param overrides the init one then the test in `test_feature_flags` should cover that, right?

liyiy: What is the problem/issue here of why `posthog.capture(send_feature_flags=True, geoip_disable=False)` won't work?

neilkakkar: https://github.com/PostHog/posthog-python/blob/geoip-disable/posthog/client.py#L211

This is separate from `test_feature_flags`, it's for `test_capture`. In the line above^, we're not passing in `geoip_disable` coming from capture, which is why anything the user passes in there is discarded when it comes to `_get_active_feature_variants`

liyiy: oh I thought that would've been fine since it's a capture call that still hits the logic in enqueue for it

liyiy: but I guess what you're saying here is the `feature flags` that returns will be incorrect because `geoip_disable` isn't passed there, yes?

neilkakkar: correct, so it will rightly not add geoip properties to the captured event, but wrongly add feature flags.

liyiy: Can we lead with that next time? 😁 I'm better at working backwards from the initial problem than building up to the problem, if that makes sense?

neilkakkar: Not quite sure I get it. Could you give me another example of what this means?

PR #99, opened 2023-04-21
fix: disable behaviour for feature flags

neilkakkar:
Disable used to return none for feature flags as well, but this behaviour broke with the client refactor.

Noticed while trying to update client to latest version


EDsCODE (commented):
✅

PR #105, opened 2023-11-06
feat(flags): Add relative date operators, fix numeric ops

neilkakkar:
Copy of https://github.com/PostHog/posthog/pull/18359/  https://github.com/PostHog/posthog/pull/18349 and https://github.com/PostHog/posthog/pull/18321/ but for SDKs. All other SDKs to follow

### DIFF
--- a/posthog/feature_flags.py
+++ b/posthog/feature_flags.py
@@ -311,7 +311,6 @@ def relative_date_parse_for_feature_flag_matching(value: str) -> Optional[dateti
             # Guard against overflow, disallow numbers greater than 10_000
             return None
 
-        print(number)
         interval = match.group("interval")
         if interval == "h":
             parsed_dt = parsed_dt - relativedelta(hours=number)
--- a/posthog/feature_flags.py
+++ b/posthog/feature_flags.py
@@ -212,7 +212,8 @@ def match_property(property, property_values) -> bool:
         else:
             raise InconclusiveMatchError("The date provided must be a string or date object")
 
-    return False
+    # if we get here, we don't know how to handle the operator
+    raise InconclusiveMatchError(f"Unknown operator {operator}")
 
 
 def match_cohort(property, property_values, cohort_properties) -> bool:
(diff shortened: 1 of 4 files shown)


posthog/feature_flags.py
```diff
@@ -142,41 +145,64 @@ def match_property(property, property_values) -> bool:
...
-        return type(override_value) is type(value) and override_value < value
+        parsed_value = None
+        try:
+            parsed_value = float(value)  # type: ignore
+        except Exception:
+            pass
 
-    if operator == "lte":
-        return type(override_value) is type(value) and override_value <= value
+        if parsed_value is not None:
```
neilkakkar, line 169:
```suggestion
        if parsed_value is not None and override_value is not None:
```

jurajmajerik (approved):

posthog/test/test_feature_flags.py
```diff
@@ -1801,6 +1803,21 @@ def test_match_properties_math_operators(self):
 
         self.assertFalse(match_property(property_d, {"key": "44"}))
         self.assertFalse(match_property(property_d, {"key": 44}))
+        self.assertTrue(match_property(property_d, {"key": 42}))
+
+        property_e = self.property(key="key", value="30", operator="lt")
+        self.assertTrue(match_property(property_e, {"key": "29"}))
+
+        # depending on the type of override, we adjust type comparison
+        self.assertTrue(match_property(property_e, {"key": "100"}))
```
liyiy, line 1812: shouldn't this be false??

liyiy: should we be converting these strings to numbers for the comparison

liyiy: or are we expecting users to expect this because they're passing in a string in both instances

neilkakkar: we can't always just convert, because we can't be sure it's a number (for example: 1.21 might be a version string, which can fail a numeric comparison, since 1.190 > 1.21)

So, we use the type of the override: if it's string, do a string comparison, if it's number, do a numeric comparison.

liyiy (approved):

PR #115, opened 2024-02-05
fix(routing): Update hosts to point to right ingestion host

neilkakkar:
Tested locally for different hosts, see things work fine

### DIFF
--- a/posthog/request.py
+++ b/posthog/request.py
@@ -13,8 +13,21 @@ from posthog.version import VERSION
 
 _session = requests.sessions.Session()
 
-DEFAULT_HOST = "https://app.posthog.com"
+US_INGESTION_ENDPOINT = "https://us-api.i.posthog.com"
+EU_INGESTION_ENDPOINT = "https://eu-api.i.posthog.com"
+DEFAULT_HOST = US_INGESTION_ENDPOINT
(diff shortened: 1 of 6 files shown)


posthog/request.py
```diff
@@ -13,10 +13,24 @@
...
+EU_INGESTION_ENDPOINT = "https://eu-api.i.posthog.com"
+DEFAULT_HOST = US_INGESTION_ENDPOINT
 USER_AGENT = "posthog-python/" + VERSION
 
 
+def determine_server_host(host: Optional[str]) -> str:
+    """Determines the server host to use."""
+    host_or_default = host or DEFAULT_HOST
+    trimmed_host = remove_trailing_slash(host_or_default)
+    if trimmed_host in ("https://app.posthog.com", "https://us.posthog.com"):
```
frankh, line 26: let's add t.posthog.com too here maybe?

neilkakkar: What is `t.posthog.com` ?

frankh (approved):

posthog/request.py
frankh: no idea, but it's in the tests and presumably we have a little bit of legacy traffic

PR #117, opened 2024-02-14
Add historical_migration option to toplevel Client

bretthoerner:
[`**kwargs` is turned into `body` and just JSON serialized](https://github.com/PostHog/posthog-python/blob/ba7a77c21b5f5a15241bf3e7cf518c41601fd318/posthog/request.py#L39-L44). That's how `batch` even gets into the body, which is a sibling to `historical_migration` as we'd expect.

The idea is we'd just pass `historical_migration=True` in the migration script here: https://github.com/PostHog/posthog-migration-tools/blob/ae2429c58bff15801eeace078b259ba2dfc9b9fd/migrate.py#L172

### DIFF
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -49,6 +49,7 @@ class Client(object):
         project_api_key=None,
         disabled=False,
         disable_geoip=True,
+        historical_migration=False,
     ):
         self.queue = queue.Queue(max_queue_size)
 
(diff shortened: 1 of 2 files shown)


tiina303 (approved):

neilkakkar:
hey @bretthoerner reminder to bump version.py and add to changelog.md to release this change: https://github.com/PostHog/posthog-python?tab=readme-ov-file#releasing-versions

(Generally, you can do it in the same PR as your change)

PR #119, opened 2024-02-23
Switch us-api.i hosts to just us.i

frankh:
We dropped the -api as it's a bit confusing (is capture really api?)


posthog/version.py
```diff
@@ -1,4 +1,4 @@
-VERSION = "3.4.2"
```
neilkakkar, line 1: update CHANGELOG.md too please?

neilkakkar (approved):

benjackwhite (approved):

PR #120, opened 2024-02-29
fix(flags): Add a shorter configurable timeout for flag requests

neilkakkar:
^^

Docs update to follow


jurajmajerik (approved):

benjackwhite (approved):

jhgaylor:
Thanks folks. This will help my org out a bunch. We really want to send feature flag info with captures but the ergonomics of that when the api is slow/down are no fun. I think this will fix my issue and I'm looking forward to the release.

PR #121, opened 2024-05-02
Use flags_by key for faster access

webjunkie:


posthog/client.py
```diff
@@ -578,23 +578,21 @@ def get_feature_flag(
...
-                            group_properties=group_properties,
-                        )
-                        self.log.debug(f"Successfully computed flag locally: {key} -> {response}")
-                    except InconclusiveMatchError as e:
-                        self.log.debug(f"Failed to compute flag {key} locally: {e}")
-                        continue
-                    except Exception as e:
-                        self.log.exception(f"[FEATURE FLAGS] Error while computing variant locally: {e}")
-                        continue
+            flag = self.feature_flags_by_key.get(key)
```
haacked, line 581: The problem here is `feature_flags_by_key` is not guaranteed to be populated here.

haacked (changes requested):

posthog/client.py
haacked: I think a better solution would be to make `feature_flags` a `@property`.  That way, we can pre-populate other related properties when it's set. Something like:

```python
@property
    def feature_flags(self):
        return self._feature_flags

    @feature_flags.setter 
    def feature_flags(self, value):
        self._feature_flags = value
        self.feature_flags_by_key = {
            flag["key"]: flag for flag in (value or []) if flag.get("key") is not None
        }
```

That way, we won't need to update every unit test.

haacked:
I have a PR that fixes this: #207

PR #123, opened 2024-05-15
Fix for Case Sensitivity Issue in Python Library When Retrieving Feature Flag Payload

reihtw:
When trying to retrieve the feature flag payload, it is not possible to do so using uppercase characters because the code always converts the provided feature flag key to lowercase.

This issue occurs only in the Python library. I tried retrieving the payload using other languages, and it works normally.

### DIFF
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -670,7 +670,7 @@ class Client(object):
             decide_payloads = self.get_feature_payloads(
                 distinct_id, groups, person_properties, group_properties, disable_geoip
             )
-            response = decide_payloads.get(str(key).lower(), None)
+            response = decide_payloads.get(str(key), None)
 
         return response
 


dmarticus:
Hi @reihtw, sorry for the delay in reviewing this, I was OOO for the last week.  

I reproduced this issue locally and confirmed that your fix works.  Thanks for the contribution!  Do you mind adding a test to `test/test_feature_flags.py` that verifies this behavior?  It couldn't hurt to encode these types of bugs into our tests.  Thank you!

marandaneto:
@PostHog/team-feature-flags should we get this merged?

dmarticus:
@marandaneto I ended up doing this with https://github.com/PostHog/posthog-python/pull/191, so we can close this out

PR #124, opened 2024-05-29
Remove deprecated `datetime.utcnow()` in favour of `datetime.now(tz=tzutc())`

silvanocerza:
[As documented](https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow) the function `datetime.utcnow()` is deprecated and will be removed in future versions.

They recommend using `datetime.now(timezone.utc)` in its place. I decided to use `tzutc()` in this case as I see the project is using `python-dateutils` and this might be prefered.

This won't probably be removed in `3.13` as it just been deprecated in `3.12` so it's low priority, though still something that needs to be updated as it emits annoying deprecation warnings.


silvanocerza:
@neilkakkar should I close this?

neilkakkar:
i'll get this merged in, sorry this fell through the cracks!

neilkakkar (approved):

PR #128, opened 2024-08-21
feat(error-capture): Add basic exception autocapture

neilkakkar:
Some things still TODO, but basic structure is in place. And some other things which I'm unclear about (see todos). Also need to add tests...

Looks like this:

image

I think I need to look at the stack trace parsing, it looks a bit wonky in app currently.


posthog/exception_capture.py
```diff
@@ -0,0 +1,54 @@
...
+from typing import TYPE_CHECKING
+
+from posthog.exception_utils import single_exception_from_error_tuple
+
+if TYPE_CHECKING:
+    from posthog.client import Client
+
+
+class ExceptionCapture:
+    # TODO: Some method of avoiding internal SDK exceptions?
```
pauldambra, line 13: i guess maybe something like the JS client side rate limiting too.... we don't want a tight loop with an error to generate too many messages. not blocking since it'll take a while to shake out things like that

posthog/exception_capture.py
```diff
@@ -0,0 +1,54 @@
...
+            if "stacktrace" in exception_info and exception_info["stacktrace"].get("frames")
+            else None
+        )
+
+        properties = {
+            "$exception_type": exc_type.__name__,
+            "$exception_message": str(exc_value),
+            "$exception_stack_trace_raw": json.dumps(stack_trace),
+            # TODO: Can we somehow get distinct_id from context here? Stateless lib makes this much harder? 😅
+            # '$exception_personURL': f'{self.client.posthog_host}/project/{self.client.token}/person/{self.client.get_distinct_id()}'
```
pauldambra, line 48: almost want in-memory registered properties
can imagine code that has info useful in an exception is distant from code that throws it
so you could register super properties including distinct id and we can pull properties out to use them like her for the distinct id but arbitrary exception properties seems useful

obvs wouldn't work as well as in a stateless library but nice to have effectively "per request" properties

#ScopeCreep :p

posthog/exception_capture.py
```diff
@@ -0,0 +1,54 @@
...
+        }
+
+        # TODO: What distinct id should we attach these server-side exceptions to?
```
pauldambra, line 51: `distinct_id or. exception_type or "python-exceptions"`

🙈 

not sure there's any benefit to spreading the exception distinct ids over multiple fake distinct ids like exception type though....

posthog/exception_capture.py
neilkakkar: Even with these, I think it will be pretty hard for it to be correct (atleast in python land, seems ideal for PHP land where generally the SDK is initialised per request). Consider:

1. There's a single instance of the SDK per thread - the overarching request can pass info across processes/threads, but there's no good way for SDKs to pass info along (for example, error in celery task, fired from an API request - will have different python SDK installations)
2. An SDK singleton across shared resources, in which case we don't know when one context ends and another one begins (example: 2 requests come in, first one sends a capture event, then second one sends a capture event, then first one errors out, and the super properties are mangled)

I can still try out doing the in-memory registered props and see how it works for us, but doesn't _seem_ likely to me that it will be accurate :p 
At the same time, sentry does manage to connect enough useful information though, which seems to be about manually tagging scope in the right place at the right time (context manager wins), but I'd love for it to be automatic, if people are also capturing other event information around the same time. Guess we can explore this, and if it doesn't work, fallback to what sentry does, i.e. pass the onus onto the user.

posthog/__init__.py
```diff
@@ -19,6 +19,7 @@
 poll_interval = 30  # type: int
 disable_geoip = True  # type: bool
 feature_flags_request_timeout_seconds = 3  # type: int
+alpha_enable_exception_autocapture = False  # type: bool
```
marandaneto, line 22: what if we add a comment that is in alpha? otherwise, we'd need to rename it to `beta_*` and remove the prefix afterward (unnecessary breaking changes).

posthog/__init__.py
```diff
@@ -454,6 +455,10 @@ def _proxy(method, *args, **kwargs):
             disabled=disabled,
             disable_geoip=disable_geoip,
             feature_flags_request_timeout_seconds=feature_flags_request_timeout_seconds,
+            # TODO: Currently this monitoring begins only when the Client is initialised (which happens when you do something with the SDK)
+            # This kind of initialisation is very annoying for exception capture. We need to figure out a way around this,
+            # or deprecate this proxy option fully (it's already in the process of deprecation, no new clients should be using this method since like 5-6 months)
```
marandaneto, line 460: > which happens when you do something with the SDK

that means someone has to do `posthog.capture` and only then it works?
assuming that the auto capture exception also calls `posthog.capture`, should it work as it is?

posthog/exception_capture.py
marandaneto, line 13: the SDK has to degrade gracefully which means wrapping up the whole SDK with a try-catch and fallback to logging if something goes wrong, but it should never capture its exception otherwise it causes an infinite loop

posthog/exception_capture.py
```diff
@@ -0,0 +1,54 @@
...
+            exception_info["stacktrace"]["frames"]
+            if "stacktrace" in exception_info and exception_info["stacktrace"].get("frames")
+            else None
+        )
+
+        properties = {
+            "$exception_type": exc_type.__name__,
+            "$exception_message": str(exc_value),
+            "$exception_stack_trace_raw": json.dumps(stack_trace),
+            # TODO: Can we somehow get distinct_id from context here? Stateless lib makes this much harder? 😅
```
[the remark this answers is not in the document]
marandaneto: for exceptions happening out of a request context, maybe we can have our "service distinct id" which is generated during the SDK init, so we all know those errors were not experienced by users, or at least not directly.

posthog/exception_capture.py
marandaneto: didn't see this comment before but https://github.com/PostHog/posthog-python/pull/128/files#r1724772600
assuming distinct id is a mandatory field, otherwise I'd just not set at all

posthog/exception_capture.py
```diff
@@ -0,0 +1,54 @@
...
+            # '$exception_personURL': f'{self.client.posthog_host}/project/{self.client.token}/person/{self.client.get_distinct_id()}'
+        }
+
+        # TODO: What distinct id should we attach these server-side exceptions to?
+        # Any heuristic seems prone to errors - how can we know if exception occurred in the context of a user that captured some other event?
```
marandaneto, line 52: https://github.com/PostHog/posthog/issues/20026
by sharing the session id & more metadata via headers and setting to the scope of that request, otherwise its out of the user-request scope eg a background service

marandaneto:
one important point is that when an unhandled exception happens and the process is gonna die, is that we flush all the pending events first otherwise it's all lost (if in memory only), assuming the process can die for certain types of errors

posthog/exception_capture.py
daibhin: This feels more and more necessary - will be so good for replay too

posthog/__init__.py
neilkakkar: hmm, we can make it not-breaking, but agree, makes sense to not need the changes at all. 👍

posthog/exception_capture.py
neilkakkar: is mandatory, thanks for the other pointer!

posthog/__init__.py
neilkakkar: sadly no, because at that point its too late, we need to initialise so the hooks exist before any exception occurs (capture is called after the exception). I'll make sure all docs don't mention the proxy version and only the initialisation version which has none of these problems

neilkakkar:
I think this is now in a good enough state to ship and try out for ourselves. I want to do this before figuring out distinct id handling

posthog/exception_capture.py
```diff
@@ -0,0 +1,61 @@
...
+                exception_info["stacktrace"]["frames"]
+                if "stacktrace" in exception_info and exception_info["stacktrace"].get("frames")
+                else None
+            )
+            all_exceptions_with_trace = exceptions_from_error_tuple((exc_type, exc_value, exc_traceback))
+
+            properties = {
+                "$exception_type": all_exceptions_with_trace[0].get("type"),
+                "$exception_message": all_exceptions_with_trace[0].get("value"),
+                "$exception_stack_trace_raw": json.dumps(stack_trace),
```
marandaneto, line 50: these 3 properties should be inferred from the 1st item of the `all_exceptions_with_trace`, otherwise it's just duplicating data

marandaneto (approved):
left a last comment about code/data dupe but otherwise LGTM (not a blocker).

posthog/exception_capture.py
neilkakkar: whoops, nice catch, missed this in cleanup! - We can delete `$exception_stack_trace_raw` now imo, its superceded by `$exception_list`

neilkakkar: (will wait to hear from @daibhin / @pauldambra  before merging incase we want to change names / adjust something here / figure out if for some reason this doesn't work well in js-land)

marandaneto: JS also has chained exceptions via the `clause` prop so its technically possible.

daibhin: Thought we might use `$exception_stack_trace_raw` in the Sentry integration but it doesn't look like we do. Happy to remove it in that case

daibhin (approved):

neilkakkar:
oh, who else is using this? I thought it was only us!

PR #129, opened 2024-08-21
fix(feature flags): Guard for None values when comparing person Properties

Phanatic:
## Problem

Porting over the fix from `posthog-node` to `posthog-python` : https://github.com/PostHog/posthog-js-lite/pull/249

The current behavior of stringifying `None` values to compare with operators can lead to invalid results in local evaluation. 

This causes local evaluation to return `true`, while the `/decide` API does the right thing and returns false for `undefined` values in these properties.  

Customer reported that local evaluation always evaluates a condition to true, when it should evaluate to false. 
The conditions in question are : 

``` json
[
    {
        "key": "latestBuildVersion",
        "type": "person",
        "value": ".+",
        "operator": "regex"
    },
    {
        "key": "latestBuildVersionMajor",
        "type": "person",
        "value": "23",
        "operator": "gt"
    },
    {
        "key": "latestBuildVersionMinor",
        "type": "person",
        "value": "31",
        "operator": "gt"
    },
    {
        "key": "latestBuildVersionPatch",
        "type": "person",
        "value": "0",
        "operator": "gt"
    }
]
```

And the person properties passed into local evaluation are : 

``` python
{
    "personProperties": {
        "latestBuildVersion": None,
        "latestBuildVersionMajor": None,
        "latestBuildVersionMinor": None,
        "latestBuildVersionPatch": None
    }
}
```

Since we stringify all of the `None` values, the above call returns `true` when it should return `false` because the property values are undefined and not the string `None`. 

## Changes

Add a `NONE_VALUES_ALLOWED_OPERATORS` array of operators that allow NONE values which is just `is_not` right now . Any other operator will fail if the overrideValue is `NONE`

### DIFF
--- a/posthog/feature_flags.py
+++ b/posthog/feature_flags.py
@@ -13,7 +13,8 @@ __LONG_SCALE__ = float(0xFFFFFFFFFFFFFFF)
 
 log = logging.getLogger("posthog")
 
-NONE_VALUES_ALLOWED_OPERATORS = ['is_not']
+NONE_VALUES_ALLOWED_OPERATORS = ["is_not"]
+
 
 class InconclusiveMatchError(Exception):
     pass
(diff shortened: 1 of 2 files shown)


dmarticus (approved):
Will this need a version bump as well?

PR #131, opened 2024-08-29
feat(errors): Add django integration and in app frames

neilkakkar:
image


posthog/exception_integrations/django.py
```diff
@@ -0,0 +1,69 @@
...
+class DjangoRequestExtractor:
+
+    def __init__(self, request):
+        # type: (Any) -> None
+        self.request = request
+
+    def extract_person_data(self):
+        headers = self.headers()
+        return {
+            "distinct_id": headers.get("X-PostHog-Distinct-ID"),
```
neilkakkar, line 62: hmm, this doesn't work as well as I'd hoped, because the client side requests coming in don't have this header, only our SDK requests coming in will, which is useless 🤦.

Need something like we did with the X-posthog-session-id, but automatic, where requests from a clients app add this header. Need to research if this is possible. But the good thing is, the django integration part is covered - if the requests have it, it will assign it correctly.

posthog/exception_capture.py
```diff
@@ -1,24 +1,46 @@
...
+            # TODO: Maybe find a better way of enabling integrations
+            # This is very annoying currently if we had to add any configuration per integration
+            if integration == Integrations.Django:
+                try:
+                    from posthog.exception_integrations.django import DjangoIntegration
+
+                    enabled_integration = DjangoIntegration(self.exception_receiver)
+                    self.enabled_integrations.append(enabled_integration)
+                except Exception as e:
+                    self.log.exception(f"Failed to enable Django integration: {e}")
```
marandaneto, line 43: it'd be cool if the integration could configure itself.

```
integrations = [DjangoIntegration]
init(client, integrations)
```

The `Integration` interface has a `install|uninstall` method that is called after the SDK is init.
the `install` method in the `DjangoIntegration` class does all the config.
if the SDK close is called, we also call all `uninstall` methods from all `integrations`
an example https://github.com/PostHog/posthog-android/blob/bb0f3d30cc3d9819000b43c281296a30483841fa/posthog-android/src/main/java/com/posthog/android/internal/PostHogActivityLifecycleCallbackIntegration.kt#L73-L77

posthog/exception_integrations/django.py
neilkakkar: I don't think that's useful enough, at minimum we need to then make it easier to do this 🤔 .

Will update this to use trace-state though, thanks 👍

posthog/exception_capture.py
neilkakkar: nice, will do

marandaneto: maybe integrations can be also a property in the `client`

```
client.integrations = [DjangoIntegration]
init(client)
```

posthog/exception_integrations/django.py
marandaneto: yep, if the SDK is on both sides, everything should be automatic via monkey patching whenever possible, or easy enough via docs, that's the state of the art today, otherwise there's no way to connect the dots on both ends

posthog/exception_capture.py
neilkakkar: was a bit reluctant to spend more time on this interface, given i'm not yet sure how most integrations will look like. If they need to be initialised post client initialisation / need some config options from users to happen first, would make sense. We can make breaking changes here though as long as its in alpha, so not too worried.

posthog/exception_integrations/django.py
```diff
@@ -0,0 +1,88 @@
...
+
+        # Extract traceparent and tracestate headers
+        traceparent = headers.get("traceparent")
+        tracestate = headers.get("tracestate")
+
+        # Extract the distinct_id from tracestate
+        distinct_id = None
+        if tracestate:
+            # TODO: Align on the format of the distinct_id in tracestate
+            # We can't have comma or equals in header values here, so maybe we should base64 encode it?
```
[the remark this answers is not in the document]
marandaneto: we are just reading from `tracestate` so we don't need to encode anything, or rather, should we decode? it depends on what the client does

marandaneto (approved):

posthog/exception_integrations/django.py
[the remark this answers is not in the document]
neilkakkar: yeah I meant client should encode (and thus decode here) as otherwise distinct ids with `,` or `=` in them will go wonky

PR #134, opened 2024-09-05
feat(errors): Add manual exception capture

neilkakkar:


posthog/__init__.py
```diff
@@ -251,6 +251,50 @@ def alias(
...
 
 
+def capture_exception(
+    exception=None,  # type: Optional[BaseException]
+    distinct_id=None,  # type: Optional[str]
+    properties=None,  # type: Optional[Dict]
+    context=None,  # type: Optional[Dict]
+    timestamp=None,  # type: Optional[datetime.datetime]
+    uuid=None,  # type: Optional[str]
+    groups=None,  # type: Optional[Dict]
```
marandaneto, line 261: Not sure if `groups` would apply here?

posthog/__init__.py
```diff
@@ -251,6 +251,50 @@ def alias(
     )
 
 
+def capture_exception(
+    exception=None,  # type: Optional[BaseException]
```
marandaneto, line 255: I think this should be mandatory, or if called without, the method is a NoOp.
If the goal is to still capture a synthetic error (there's no error instance), we should have a let's say `capture_message` which has a more semantical name

posthog/__init__.py
neilkakkar: why not? if custom exceptions, the caller can / should be able to assign them to groups

posthog/__init__.py
neilkakkar: Hmm, explained below:

> - `exception` to specify the exception to capture. If not provided, the current exception is captured via `sys.exc_info()`

Makes sense?

Although indeed I'm not sure if I'm handling the case where exc_info() is empty, will add handling for that 👍 , and some tests, since this doesn't depend on the django integration

posthog/__init__.py
marandaneto: mm, it's a good point, I just have not seen anything like this, maybe that's the equivalent of what we'd call "tags" elsewhere? would you be able to filter exceptions per groups too?

posthog/__init__.py
marandaneto: yep, makes a lot of sense, It's a nice fallback.

daibhin (approved):

posthog/__init__.py
neilkakkar: Definitely possible. Since they are eventually stored as events, (and since js automatically adds this info to all events) this will allow for seeing exceptions over time per group (not just persons)

marandaneto: awesome

PR #138, opened 2024-10-03
feat: add super properties

daibhin:
### Problem

Customers of error tracking want to know what repository a particular exception came from by filtering in the UI. Rather than adding some specific property (e.g. `exception_source_identifier`) that would then need to be added to the JS SDK and any other language we support in future, I thought it would be easier to reuse the super properties concept found in our other SDKs 

I'm not tied to this approach but "PRs > Slack" and all that... Open to feedback here.

### Changes

- Went with a very basic implementation that allows you to set `super_properties` when initialising the client
  - Could have gone with the `register` method seen elsewhere but didn't want to mislead customers into thinking that properties were somehow stored elsewhere (like they are in local storage). Better to make sure that they're always included when the client is initialised

### DIFF
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -213,9 +213,6 @@ class Client(object):
         require("properties", properties, dict)
         require("event", event, string_types)
 
-        if self.super_properties:
-            properties = {**properties, **self.super_properties}
-
         msg = {
             "properties": properties,
             "timestamp": timestamp,
@@ -447,6 +444,9 @@ class Client(object):
 
         if disable_geoip:
             msg["properties"]["$geoip_disable"] = True
+
+        if self.super_properties:
+            msg["properties"] = {**msg["properties"], **self.super_properties}
 
         msg["distinct_id"] = stringify_id(msg.get("distinct_id", None))
 
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -55,6 +55,7 @@ class Client(object):
         disable_geoip=True,
         historical_migration=False,
         feature_flags_request_timeout_seconds=3,
+        super_properties=None,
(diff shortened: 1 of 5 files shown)


posthog/client.py
```diff
@@ -211,6 +213,9 @@ def capture(
         require("properties", properties, dict)
         require("event", event, string_types)
 
+        if self.super_properties:
```
neilkakkar, line 216: annoyingly in python capture is not the root - we add other "super properties" (like the lib and version) in `enqueue` because other functions like identify, set, etc. etc. directly call enqueue.

So ideal for this handling to happen there as well

posthog/test/test_client.py
```diff
@@ -84,6 +84,16 @@ def test_basic_capture_with_project_api_key(self):
         self.assertEqual(msg["properties"]["$lib"], "posthog-python")
         self.assertEqual(msg["properties"]["$lib_version"], VERSION)
 
+    def test_basic_capture_with_super_properties(self):
+        client = Client(FAKE_TEST_API_KEY, super_properties={"source": "repo-name", "environment": "test"})
+
+        _, msg = client.capture("distinct_id", "python test event")
```
neilkakkar, line 90: here you can add `identify` as well and make sure its added there

neilkakkar (commented):
One blocking comment, otherwise LGTM! Pragmatic approach to adding it on sdk init, I like it

neilkakkar:
(also remember to bump version and changelog before merging)

posthog/client.py
daibhin: TIL

Updated in the latest commit

neilkakkar (approved):
noice

PR #139, opened 2024-10-24
chore: add lang to frames

daibhin:
Similar to https://github.com/PostHog/posthog-js/pull/1495

We want to know the `lang` of each frame so that we can switch on it in Rust


neilkakkar (approved):

neilkakkar:
remember to follow release instructions in this PR (add changelog and version.py) before merging!

PR #144, opened 2024-11-25
fix(err): fix distinct_id, set personless and use a uuid

oliverb123:
Ingest will punish you for using a fixed value for distinct_id, so let's not do that.

### DIFF
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -10,7 +10,7 @@ from dateutil.tz import tzutc
 from six import string_types
 
 from posthog.consumer import Consumer
-from posthog.exception_capture import DEFAULT_DISTINCT_ID, ExceptionCapture
+from posthog.exception_capture import ExceptionCapture
 from posthog.exception_utils import exc_info_from_error, exceptions_from_error_tuple, handle_in_app
 from posthog.feature_flags import InconclusiveMatchError, match_feature_flag_properties
 from posthog.poller import Poller
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -4,7 +4,7 @@ import numbers
 import os
 import sys
 from datetime import datetime, timedelta
-from uuid import UUID
+from uuid import UUID, uuid4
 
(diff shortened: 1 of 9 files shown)


posthog/client.py
```diff
@@ -362,7 +362,7 @@ def page(
     def capture_exception(
         self,
         exception=None,
-        distinct_id=DEFAULT_DISTINCT_ID,
```
daibhin, line 365: We should remove this variable entirely

posthog/exception_capture.py
```diff
@@ -61,14 +62,15 @@ def exception_receiver(self, exc_info, extra_properties):
...
 
-            distinct_id = metadata.get("distinct_id") if metadata else DEFAULT_DISTINCT_ID
-            # Make sure we have a distinct_id if its empty in metadata
-            distinct_id = distinct_id or DEFAULT_DISTINCT_ID
+            # if there's no distinct_id, we'll generate one and set personless mode
+            # via $process_person_profile = false
+            properties = {}
+            if distinct_id is None:
+                properties["$process_person_profile"] = False
+                distinct_id = uuid.uuid4()
```
daibhin, line 72: Should we be using v7 going forward? Unrelated to this change

daibhin (approved):

posthog/exception_capture.py
oliverb123: Maybe? We don't really care what's in `distinct_id`, and don't make recommendations to customers about it I think - we probably should us v7 everywhere just for the sake of the habit, but the python standard library doesn't support it yet afaict.

oliverb123:
I ran `black .` locally and then committed exactly what it gave me, and CI says it's wrong. Should I care about lint checks?

posthog/client.py
oliverb123: Done

daibhin:
@oliverb123 I think we use ruff instead. My VSCode is just setup to auto format afaik

Issue #149, opened 2024-11-29
Logging should have a prefix


stanislavkozlovski:
Today, (in debug, since that's what I use) the library logs arbitrary stuff like:
```
consumer exited.

queueing: {'properties': {'..., '$lib': 'posthog-python', '$lib_version': '3.7.2', '$geoip_disable': True}, 'timestamp': '2024-11-29T13:25:36.143687+00:00', 'context': {}, 'distinct_id': '019362b1-37ca-7e98-bb6a-0c9c54c1b99a', 'event': '...'}

enqueued [event].

making request: {"batch": ..."}, "$lib": "posthog-python", "$lib_version": "3.7.2", "$geoip_disable": true}, "timestamp": "2024-11-29T13:25:36.143687+00:00", "context": {}, "distinct_id": "019362b1-37ca-7e98-bb6a-0c9c54c1b99a", "event": "..."}], "historical_migration": false, "sentAt": "2024-11-29T13:25:36.536523+00:00", "api_key": "..."}

data uploaded successfully
```

In any regular app, this is easy to confuse with the regular application logs. And more importantly - **it's hard to tell apart** when you want to debug posthog specifically.

It would be great if the logger had some set prefix.

stanislavkozlovski:
The way I get around this is through this method:
```
def fix_posthog_logging(posthog: Posthog):
    class PosthogClientLogPrefixFilter(logging.Filter):
        def filter(self, record):
            record.msg = f'[PostHog-Client] {record.msg}'
            return True
    for f in posthog.log.filters:
        if isinstance(f, PosthogClientLogPrefixFilter) or f.__class__.__name__ == 'PosthogClientLogPrefixFilter':
            return # it's already there!

    posthog.log.addFilter(PosthogClientLogPrefixFilter())
```

It's a bit inefficient re: the message append, and it's weird code because I have to double-check for the filter's presence to avoid adding it twice (something regarding Django initalization I guess)

PR #150, opened 2024-11-29
MINOR: Set a prefix in the logger

stanislavkozlovski:
Fixes https://github.com/PostHog/posthog-python/issues/149


rafaeelaudibert:

PR #156, opened 2025-01-03
fix: CI

dmarticus:
Cleans up CI.yml with the following changes:
- updates the `checkout` action to a more current stable version
- updates the `setup-python` action to a more current stable version
- updates the python version to one supported by our CI architecture
- updates the `.gitignore` so that folks can include a pyrightconfig.json file for their IDE (relevant for folks using `basedpyright` instead of `pylance` for their IDE)

### DIFF
--- a/example.py
+++ b/example.py
@@ -1,7 +1,7 @@
 # PostHog Python library example
 
 # Import the library
-import time
+# import time
 
 import posthog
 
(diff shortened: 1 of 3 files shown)


example.py
```diff
@@ -1,7 +1,7 @@
 # PostHog Python library example
 
 # Import the library
-import time
+# import time
```
dmarticus, line 4: we use this in the example but my editor is chirping me for keeping it in so commenting it out

havenbarnes (approved):

PR #157, opened 2025-01-07
Allow specifying the flag in the example script

haacked:
This is a little quality of life improvement. When using the example script, it creates an event for the feature flag `person-on-events-enabled`. This change adds an optional argument that lets you change the feature-flag key for the event like so:

```bash
python3 example.py --flag person-on-events-enabled-v1
```

If the flag is omitted, the script behaves the same as it always did.

### DIFF
--- a/example.py
+++ b/example.py
@@ -3,10 +3,10 @@ import argparse
 import posthog
 
 # Add argument parsing
-parser = argparse.ArgumentParser(description='PostHog Python library example')
-parser.add_argument('--flag', 
-                   default='person-on-events-enabled',
-                   help='Feature flag key to check (default: person-on-events-enabled)')
+parser = argparse.ArgumentParser(description="PostHog Python library example")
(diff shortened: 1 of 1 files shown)


example.py
```diff
@@ -18,7 +23,7 @@
 
 print(
     posthog.feature_enabled(
-        "person-on-events-enabled",
+        args.flag,  # Use the flag from command line arguments
```
dmarticus, line 26: fwiw, this only changes the first test flag in this case, the subsequent flags will still use the hard coded values. Seems fine if you're only really using the example script for the first case.

dmarticus (approved):
I don't feel strongly either way about this; I typically just go in and configure the example script to do what I want it to, but this seems useful enough.

example.py
haacked: I’ll cross those bridges if the need arises. 😂

PR #158, opened 2025-01-09
Feat: Add llm observability to python sdk

k11kirky:
Initial PR for LLM Observability

- Support for OpenAI SDK chat completions with OpenAI and AsyncOpenAI
- Example file to generate data, usage example below

example data can be created by exporting the following env vars
```
export POSTHOG_PROJECT_API_KEY=<your-project-key>
export POSTHOG_PERSONAL_API_KEY=<your-personal-key>
export OPENAI_API_KEY=<your-openai-key>
```

`python3 llm_observability_examples`


posthog/ai/providers/openai/openai.py
```diff
@@ -0,0 +1,157 @@
...
+                            for k in ["prompt_tokens", "completion_tokens", "total_tokens"]
+                        }
+                    if chunk.choices[0].delta.content:
+                        accumulated_content.append(chunk.choices[0].delta.content)
+                    yield chunk
+            finally:
+                end_time = time.time()
+                latency = end_time - start_time
+                output = "".join(accumulated_content)
+                self._capture_streaming_event(distinct_id, posthog_trace_id, posthog_properties, kwargs, usage_stats, latency, output)
```
Twixes, line 116: Can we reuse the `track_usage()` util here?

posthog/ai/providers/openai/openai.py
```diff
@@ -0,0 +1,164 @@
...
+        output: str,
+    ):
+
+        event_properties = {
+            "$ai_provider": "openai",
+            "$ai_model": kwargs.get("model"),
+            "$ai_model_parameters": get_model_params(kwargs),
+            "$ai_input": kwargs.get("messages"),
+            "$ai_output": {
+                "choices": [
```
skoob13, line 144: Do we want to send an object here instead of the array with choices directly? We've already flattened the structure with additional fields like input/output tokens, so this might be the output itself.

posthog/ai/providers/openai/openai.py
```diff
@@ -0,0 +1,164 @@
...
+                        "content": output,
+                        "role": "assistant",
+                    }
+                ]
+            },
+            "$ai_http_status": 200,
+            "$ai_input_tokens": usage_stats.get("prompt_tokens", 0),
+            "$ai_output_tokens": usage_stats.get("completion_tokens", 0),
+            "$ai_latency": latency,
+            "$ai_trace_id": posthog_trace_id,
```
skoob13, line 155: I think we have to set a trace_id to some ID by default. Otherwise, grouping won't reliably work.

posthog/ai/utils.py
```diff
@@ -0,0 +1,156 @@
...
+        "max_tokens",
+        "top_p",
+        "frequency_penalty",
+        "presence_penalty",
+        "n",
+        "stop",
+        "stream",
+    ]:
+        if param in kwargs:
+            model_params[param] = kwargs.get(param)
```
skoob13, line 23: Do we want to send null values here? I would get rid of them.

posthog/ai/utils.py
```diff
@@ -0,0 +1,156 @@
...
+    response = None
+    error = None
+    http_status = 200
+    usage: Dict[str, Any] = {}
+
+    try:
+        response = call_method(**kwargs)
+    except Exception as exc:
+        error = exc
+        http_status = getattr(exc, "status_code", 500)
```
skoob13, line 65: Why 500? If there is an exception without a status_code, that's most likely an issue with the SDK rather than the server error. Maybe something like 0, so we can display a proper UI for that. The LangChain's integration uses 0, but we can change it to any other value.

posthog/ai/providers/openai/openai.py
k11kirky: Although for most use cases "choices" only has one option, the user can ask the llm for mulitple im which case the choices array represents n options for a response. If we remove the object and just return the array it would look more like the messages array representing a history.

posthog/ai/providers/openai/openai.py
```diff
@@ -0,0 +1,164 @@
...
+        Args:
+            api_key: OpenAI API key.
+            posthog_client: If provided, events will be captured via this client instead
+                            of the global posthog.
+            **openai_config: Any additional keyword args to set on openai (e.g. organization="xxx").
+        """
+        self._openai_client = openai.OpenAI(**openai_config)
+        self._posthog_client = posthog_client
+
+    def __getattr__(self, name: str) -> Any:
```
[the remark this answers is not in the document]
k11kirky: Yes - this is really annoying, i tried subclassing, but having issues with that. Open to suggestions

posthog/ai/utils.py
k11kirky: this is what this does right? only adds them if it has a value

posthog/ai/utils.py
```diff
@@ -0,0 +1,163 @@
...
+    response = None
+    error = None
+    http_status = 200
+    usage: Dict[str, Any] = {}
+
+    try:
+        response = call_method(**kwargs)
+    except Exception as exc:
+        error = exc
+        http_status = getattr(exc, "status_code", 0)  # default to 0 becuase its liekly an SDK error
```
Twixes, line 71:
```suggestion
        http_status = getattr(exc, "status_code", 0)  # default to 0 because it's likely an SDK error
```

posthog/ai/utils.py
Twixes: Yeah, so actually this would be slightly more Pythonic given the `if` above:
```suggestion
            model_params[param] = kwargs[param]
```

posthog/ai/providers/openai/openai.py
```diff
@@ -0,0 +1,167 @@
...
+        self._openai_client = openai_client
+
+    def create(
+        self,
+        posthog_distinct_id: Optional[str] = None,
+        posthog_trace_id: Optional[str] = None,
+        posthog_properties: Optional[Dict[str, Any]] = None,
+        **kwargs: Any,
+    ):
+        distinct_id = posthog_distinct_id or uuid.uuid4()
```
Twixes, line 56: How about UUIDv7?

k11kirky: we use 4 in the rest of the repo, so trying to keep it consistant

timgl (approved):

PR #164, opened 2025-01-14
feat: privacy_mode

k11kirky:
add privacy mode for AI observability


posthog/__init__.py
```diff
@@ -26,6 +26,8 @@
 exception_autocapture_integrations = []  # type: List[Integrations]
 # Used to determine in app paths for exception autocapture. Defaults to the current working directory
 project_root = None  # type: Optional[str]
+# Used for our AI observability feature to not capture any prompt or output just usage + metadata
+privacy_mode = False  # type: bool
```
timgl, line 30: I think there's a risk of this being confusing given it's at the top level. I'd do it on the initialisation of the client and on the individual invocation

PR #165, opened 2025-01-17
Feat: Add Anthropic to Python SDK

k11kirky:
Add Anthropic to the SDK

Matches Same Usage pattern as OpenAI


skoob13 (approved):
LGTM. One Anthropic's specific is that a system prompt passed as a separate field. PR with a fix: https://github.com/PostHog/posthog-python/pull/167.

PR #166, opened 2025-01-17
fix(llm-observability): flatten langchain's additional_kwargs

skoob13:
## Problem

OpenAI wrappers send data as:
```json
{
    "role": "assistant",
    "content": "",
    "tool_calls": [
        {
            "type": "function",
            "id": "123",
            "function": {
                "name": "test",
                "args": "{\"a\": 1}",
            },
        }
    ]
}
```
But the Langchain integration adds the tool calls into the `additional_kwargs` field:
```json
{
    "role": "assistant",
    "content": "",
    "additional_kwargs": {
        "tool_calls": [
            {
                "type": "function",
                "id": "123",
                "function": {
                    "name": "test",
                    "args": "{\"a\": 1}"
                }
            }
        ]
    }
}
```
## Changes

Flatten `additional_kwargs` to mirror the OpenAI wrapper to remove the discrepancies.

### DIFF
--- a/posthog/ai/langchain/callbacks.py
+++ b/posthog/ai/langchain/callbacks.py
@@ -333,10 +333,8 @@ def _convert_message_to_dict(message: BaseMessage) -> Dict[str, Any]:
     else:
         message_dict = {"role": message.type, "content": str(message.content)}
 
-    if "name" in message.additional_kwargs:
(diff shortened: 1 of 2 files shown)


Twixes (approved):
Ah, this makes sense – will need to update PostHog/posthog#27626 too

PR #167, opened 2025-01-17
fix(llm-observability): capture system prompt for anthropic

skoob13:
System prompt is Anthropic's models passed as a separate field: https://docs.anthropic.com/en/api/messages#body-system.


k11kirky (approved):
👍

PR #168, opened 2025-01-17
fix: flatten core model params

k11kirky:
Extract core values from model params.
- temperature
- stream
- max_completion_tokens

### DIFF
--- a/posthog/ai/utils.py
+++ b/posthog/ai/utils.py
@@ -28,6 +28,34 @@ def get_model_params(kwargs: Dict[str, Any]) -> Dict[str, Any]:
             model_params[param] = kwargs[param]
     return model_params
 
+def extract_core_model_params(kwargs: Dict[str, Any], provider: str) -> Dict[str, Any]:
+    """
+    Extracts core model parameters from the kwargs dictionary.
+    """
+    output = {}
+    if provider == "anthropic":
+        if "temperature" in kwargs:
+            output["$ai_temperature"] = kwargs.get("temperature")
+        if "max_tokens" in kwargs:
+            output["$ai_max_tokens"] = kwargs.get("max_tokens")
+        if "stream" in kwargs:
+            output["$ai_stream"] = kwargs.get("stream")
+    elif provider == "openai":
+        if "temperature" in kwargs:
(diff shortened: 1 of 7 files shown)


posthog/ai/utils.py
```diff
@@ -29,6 +29,35 @@ def get_model_params(kwargs: Dict[str, Any]) -> Dict[str, Any]:
     return model_params
 
 
+def extract_core_model_params(kwargs: Dict[str, Any], provider: str) -> Dict[str, Any]:
```
Twixes, line 32: Let's do this in the plugin server, so that we only have to update one place to cover all SDKs!

posthog/ai/utils.py
```diff
@@ -29,6 +29,35 @@ def get_model_params(kwargs: Dict[str, Any]) -> Dict[str, Any]:
...
+    """
+    Extracts core model parameters from the kwargs dictionary.
+    """
+    output = {}
+    if provider == "anthropic":
```
Twixes, line 37: To track AI cost processing's case-insensitive behavior:
```suggestion
	if provider:
	    provider = provider.lower()
    if provider == "anthropic":
```

posthog/ai/utils.py
```diff
@@ -29,6 +29,35 @@ def get_model_params(kwargs: Dict[str, Any]) -> Dict[str, Any]:
...
+            output["$ai_temperature"] = kwargs.get("temperature")
+        if "max_tokens" in kwargs:
+            output["$ai_max_tokens"] = kwargs.get("max_tokens")
+        if "stream" in kwargs:
+            output["$ai_stream"] = kwargs.get("stream")
+    elif provider == "openai":
+        if "temperature" in kwargs:
+            output["$ai_temperature"] = kwargs.get("temperature")
+        if "max_completion_tokens" in kwargs:
+            output["$ai_max_tokens"] = kwargs.get("max_completion_tokens")
```
Twixes, line 48: Might also be clearest to standardize on `max_completion_tokens` (though hard to say if other providers releasing reasoning models will follow this naming)

Twixes (changes requested):
This looks good, but we'll have to duplicate the effort in other SDKs (TypeScript), so proposing inlining into the plugin server too!

PR #172, opened 2025-01-22
fix(llm-observability): parallel traces

skoob13:
## Problem

Traces implementation doesn't catch parallel traces if a single instance of LangChain callbacks is used.

## Changes

Store multiple parallel traces in the trace storage. Also, I've added capturing of trace latency.

### DIFF
--- a/posthog/ai/langchain/callbacks.py
+++ b/posthog/ai/langchain/callbacks.py
@@ -212,7 +212,7 @@ class CallbackHandler(BaseCallbackHandler):
         self._pop_parent_of_run(run_id)
 
         if parent_run_id is None:
-            self._capture_trace(run_id, outputs=outputs)
+            self._pop_trace_and_capture(run_id, outputs=outputs)
 
     def on_chain_error(
         self,
@@ -226,7 +226,7 @@ class CallbackHandler(BaseCallbackHandler):
         self._pop_parent_of_run(run_id)
 
         if parent_run_id is None:
-            self._capture_trace(run_id, outputs=None)
+            self._pop_trace_and_capture(run_id, outputs=None)
 
     def on_llm_end(
         self,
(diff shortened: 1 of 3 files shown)


posthog/ai/langchain/callbacks.py
```diff
@@ -452,10 +467,14 @@ def _get_langchain_run_name(self, serialized: Optional[Dict[str, Any]], **kwargs
 
     def _capture_trace(self, run_id: UUID, *, outputs: Optional[Dict[str, Any]]):
         trace_id = self._get_trace_id(run_id)
+        run = self._pop_run_metadata(run_id)
```
Twixes, line 470: Pretty sneaky to pop in a `_capture_...` method, how about renaming it to `_pop_trace_and_capture` for clarity? Alternatively, we can just do `self._runs[run_id]` here and pop outside of here

posthog/ai/langchain/callbacks.py
```diff
@@ -32,13 +32,22 @@
...
     model: str
+    """Model used in the run"""
     model_params: Dict[str, Any]
+    """Model parameters of the run: temperature, max_tokens, etc."""
     base_url: str
+    """Base URL of the provider's API used in the run."""
     start_time: float
+    """Start time of the run."""
     end_time: float
+    """End time of the run."""
```
Twixes, line 50: Was also thinking of traces as runs in my earlier PR, but then thought `Run` is maybe a bit too generation specific for the trace use case (e.g. traces don't have a model). So _possibly_ we could have a `GenerationMetadata` dataclass, a `TraceMetadata` one, and later `SpanMetadata` for clearer separation.
But for now `RunMetadata` for both is indeed a better model than the simplistic one I went with before ✅

Twixes (approved):
A possible code clarity suggestion, but overall looking good

posthog/ai/langchain/callbacks.py
skoob13: Yeah, it makes sense. I'll do that in the spans PR.

PR #175, opened 2025-01-27
Move accessing variants outside of loop

haacked:
`flag_variants` doesn't depend on condition so it doesn't make sense to declare it in the loop.

### DIFF
--- a/posthog/test/test_feature_flags.py
+++ b/posthog/test/test_feature_flags.py
@@ -1007,12 +1007,14 @@ class TestLocalEvaluation(unittest.TestCase):
             "beta-feature",
             "some-distinct-id",
             person_properties={
-                "latestBuildVersion": "24.32..1",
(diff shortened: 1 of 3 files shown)


haacked:
I'm on vacation so I don't have time to test this properly, but it looks like a pretty safe change. I noticed it while working on `posthog-dotnet`. Feel free to take this over if you want, or it can wait till I get back.

posthog/test/test_feature_flags.py
```diff
@@ -1007,13 +1007,15 @@ def test_feature_flags_local_evaluation_None_values(self, patch_get, patch_decid
             "beta-feature",
             "some-distinct-id",
             person_properties={
-                "latestBuildVersion": "24.32..1",
```
dmarticus, line 1010: oof

dmarticus (approved):
this looks great to me!  I can make sure it gets deployed.

PR #176, opened 2025-01-27
feat(llm-observability): LangChain spans

skoob13:
## Problem

We want to capture the input and output of intermediary chains during agent workflows.

## Changes

- Add the spans support for the LangChain callback handler.
- Replace the RunMetadata dict with dataclasses.
- Fix tests.

New fields:
- `$ai_parent_id` to capture a parent ID and build a tree.
- `$ai_generation_id` to capture a generation ID and unify API.
- `$ai_span_id` to capture an ID for a span.

### DIFF
--- a/posthog/ai/langchain/callbacks.py
+++ b/posthog/ai/langchain/callbacks.py
@@ -163,7 +163,7 @@ class CallbackHandler(BaseCallbackHandler):
         **kwargs: Any,
     ):
         self._log_debug_event("on_chain_error", run_id, parent_run_id, error=error)
-        self._pop_run_and_capture_trace_or_span(run_id, parent_run_id, str(error))
+        self._pop_run_and_capture_trace_or_span(run_id, parent_run_id, error)
 
     def on_chat_model_start(
         self,
@@ -262,7 +262,7 @@ class CallbackHandler(BaseCallbackHandler):
         **kwargs: Any,
     ) -> Any:
         self._log_debug_event("on_tool_error", run_id, parent_run_id, error=error)
-        self._pop_run_and_capture_trace_or_span(run_id, parent_run_id, str(error))
+        self._pop_run_and_capture_trace_or_span(run_id, parent_run_id, error)
 
     def on_retriever_start(
         self,
@@ -299,7 +299,7 @@ class CallbackHandler(BaseCallbackHandler):
     ) -> Any:
         """Run when Retriever errors."""
         self._log_debug_event("on_retriever_error", run_id, parent_run_id, error=error)
-        self._pop_run_and_capture_trace_or_span(run_id, parent_run_id, str(error))
+        self._pop_run_and_capture_trace_or_span(run_id, parent_run_id, error)
 
     def on_agent_action(
         self,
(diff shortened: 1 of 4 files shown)


posthog/ai/langchain/callbacks.py
```diff
@@ -403,89 +371,161 @@ def _set_llm_metadata(
...
-            serialized (Optional[Dict[str, Any]]): A dictionary containing the runnable's serialized data.
-            **kwargs (Any): Additional keyword arguments, potentially including the 'name' override.
+    def _pop_run_and_capture_trace_or_span(self, run_id: UUID, parent_run_id: Optional[UUID], outputs: Any):
+        trace_id = self._get_trace_id(run_id)
+        self._pop_parent_of_run(run_id)
+        run = self._pop_run_metadata(run_id)
+        if not run:
+            return
+        if isinstance(run, GenerationMetadata):
+            log.warning(f"Run {run_id} is not a generation but attempted to be captured as a trace or span.")
```
Twixes, line 422:
```suggestion
            log.warning(f"Run {run_id} is a generation, but attempted to be captured as a trace or span.")
```

posthog/ai/langchain/callbacks.py
```diff
@@ -403,89 +371,161 @@ def _set_llm_metadata(
...
+    def _pop_run_and_capture_generation(
+        self, run_id: UUID, parent_run_id: Optional[UUID], response: Union[LLMResult, BaseException]
+    ):
         trace_id = self._get_trace_id(run_id)
+        self._pop_parent_of_run(run_id)
         run = self._pop_run_metadata(run_id)
         if not run:
             return
+        if not isinstance(run, GenerationMetadata):
+            log.warning(f"Run {run_id} is not a generation but attempted to be captured as a generation.")
```
Twixes, line 472:
```suggestion
            log.warning(f"Run {run_id} is not a generation, but attempted to be captured as a generation.")
```

posthog/ai/langchain/callbacks.py
```diff
@@ -403,89 +371,161 @@ def _set_llm_metadata(
...
-        try:
-            return serialized["name"]
-        except (KeyError, TypeError):
-            pass
+        if parent_run_id is not None:
+            event_properties["$ai_span_id"] = run_id
+            event_properties["$ai_parent_id"] = parent_run_id
+            event_properties["$ai_span_name"] = run.name
+        else:
+            event_properties["$ai_trace_name"] = run.name
```
Twixes, line 448: This is one thing that gets me, should we be using `$ai_span_id` and `$ai_span_name` for both span and trace IDs/names after all? It feels _pretty_ intuitive and definitely much simpler than somewhat duplicated properties

skoob13: I thought about unifying them, too. Intuitively, I want to inherit `$ai_generation` and `$ai_trace` from `$ai_span`, so it makes sense to use `$ai_span_id` and `$ai_span_name` on all LLM events. We still want to keep `$ai_trace_id` and `$ai_parent_id`, though.

So I propose renaming `$ai_trace_name` (existing) to `$ai_span_name`, keeping the naming unified across all three events.

skoob13: Doing it

posthog/ai/langchain/callbacks.py
```diff
@@ -31,26 +32,38 @@
...
+    base_url: Optional[str] = None
     """Base URL of the provider's API used in the run."""
-    start_time: float
-    """Start time of the run."""
-    end_time: float
-    """End time of the run."""
 
 
-RunStorage = Dict[UUID, RunMetadata]
+RunMetadataUnion = Union[SpanMetadata, GenerationMetadata]
```
Twixes, line 65: Nit: A little redundant to suffix with `Union`
```suggestion
RunMetadata = Union[SpanMetadata, GenerationMetadata]
```

Twixes (approved):
LGTM

posthog/ai/langchain/callbacks.py
skoob13: Oh yeah, I've missed that. There was a naming conflict before haha. I'll fix it!

PR #177, opened 2025-01-27
fix: serialize pydantic models

skoob13:
## Problem

Pydantic models are skipped by default. This behavior is annoying and unclear for customers.

## Changes

Serialize Pydantic objects to dicts:
- For v1, use the `dict` method.
- For v2, use the `model_dump` method.

If something goes wrong during serialization, omit the key.


Twixes (approved):
Looking good

PR #179, opened 2025-02-05
fix: langchain tool parent add

k11kirky:
I believe we should have:

self._set_parent_of_run(run_id, parent_run_id)

on each of the callbackStart methods

### DIFF
--- a/posthog/version.py
+++ b/posthog/version.py
@@ -1,4 +1,4 @@
-VERSION = "3.11.0"
+VERSION = "3.11.1"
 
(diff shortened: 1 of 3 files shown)


skoob13 (approved):
Yep, you're totally right. I missed those two methods. Thanks!

posthog/version.py
```diff
@@ -1,4 +1,4 @@
-VERSION = "3.11.0"
```
skoob13, line 1: The changelog wasn't updated though

PR #180, opened 2025-02-08
feat(flags): Add method for fetching decrypted remote config flag payload

havenbarnes:
Context: https://github.com/PostHog/posthog/pull/27414/files
Screenshot 2025-02-10 at 2 01 50 PM
Screenshot 2025-02-10 at 2 01 56 PM


haacked:
Just so I understand, decryption always happens on the server, correct? And it looks like there's an API endpoint to get a single decrypted setting, but not a way to get all of them?

posthog/request.py
```diff
@@ -80,6 +80,9 @@ def decide(api_key: str, host: Optional[str] = None, gzip: bool = False, timeout
     res = post(api_key, host, "/decide/?v=3", gzip, timeout, **kwargs)
     return _process_response(res, success_message="Feature flags decided successfully")
 
+def remote_config(api_key: str, host: Optional[str] = None, flag_id: str = "", timeout: int = 15) -> Any:
+    res = get(api_key, host, f"/api/projects/@current/feature_flags/{flag_id}/remote_config/", timeout)
+    return _process_response(res, success_message="Feature flags decided successfully")
```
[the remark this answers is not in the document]
dmarticus: i agree with greptile here

dmarticus (approved):

posthog/version.py
```diff
@@ -1,4 +1,4 @@
-VERSION = "3.11.1"
+VERSION = "3.12.1"
```
marandaneto, line 1:
```suggestion
VERSION = "3.12.0"
```

the patch has to be always 0 when there's a new minor

posthog/request.py
```diff
@@ -81,6 +81,11 @@ def decide(api_key: str, host: Optional[str] = None, gzip: bool = False, timeout
     return _process_response(res, success_message="Feature flags decided successfully")
 
 
+def remote_config(api_key: str, host: Optional[str] = None, flag_id: int = 0, timeout: int = 15) -> Any:
+    """Get remote config flag value from remote_config API endpoint"""
+    return get(api_key, f"/api/projects/@current/feature_flags/{flag_id}/remote_config/", host, timeout)
```
marandaneto, line 86: all other methods call `_process_response` to do better error handling, should we?

posthog/__init__.py
```diff
@@ -435,6 +435,26 @@ def get_feature_flag_payload(
     )
 
 
+def get_decrypted_feature_flag_payload(
+    flag_id,  # type: int
```
haacked, line 439: Just curious, why is this `flag_id` rather than `flag_key`? How do consumers of the client SDK typically get their hands on the flag id?

posthog/request.py
haacked, line 86: `api_key` here is the personal api token, correct?

posthog/version.py
havenbarnes: Oops, duh thank you

posthog/request.py
havenbarnes: Correct, I'll label it explicitly

posthog/__init__.py
havenbarnes: Yeah, debated this.. was just trying to follow the existing REST pattern we have here https://posthog.com/docs/api/feature-flags

But might be worth making an exception or supporting either / or? I'll start thread in Slack

haacked: There's 2 issues here. What should the REST endpoint accept and what should the client require?

I think all the current methods on the client accept `feature_key` because that's how the SDK users identify feature flags. Also, there's not a consistent way for them to even retrieve the `feature_id` across the board AFAICT.

Obviously, that means if the REST api accepts feature id (to be in-line with the other REST methods), we have a problem because we have to convert the `feature_key` to a `feature_id`, which would be another request.

So my vote would be to have the REST api accept the `feature_key` as well as the `project_api_key` (as I'm assuming the combination of both is required to uniquely identify a feature key, correct?). It may not match the existing REST pattern, but it feels practical.

Otherwise, it'll take two requests to get the decrypted payload (whether we hide that from the user or we make the user do it themselves). One request to get the feauture_id and another to get the decrypted payload.

haacked: BTW, I bring this up because right now, this API would be uncallable from `posthog-dotnet` unless I start exposing the flag id somehow.

havenbarnes: Fair enough! I agree, since our REST hierarchy always includes the project / environment ID, this will be both reliably unique and more convenient. Will update https://github.com/PostHog/posthog/pull/27414 accordingly

posthog/request.py
havenbarnes: Done

posthog/__init__.py
havenbarnes: Done

posthog/client.py
```diff
@@ -849,6 +859,21 @@ def get_feature_flag_payload(
...
+        if self.personal_api_key is None:
+            self.log.warning(
+                "[FEATURE FLAGS] You have to specify a personal_api_key to fetch decrypted feature flag payloads."
+            )
+            return None
+
+        try:
+            return self.get_remote_config_payload(key)
+        except Exception as e:
+            self.log.exception(f"[FEATURE FLAGS] Unable to get decrypted feature flag payload: {e}")
```
haacked, line 875: Feel free to disregard, but I wonder if we should be capturing all exceptions here. For example, if a personal api key is supplied, but incorrect, I'd almost want my app to crash with a message to that effect. It would. make the issue immediately noticeable with a clear path to fix it. Otherwise, this call fails silently and I have to check the logs to understand why.

haacked: Just for context, I'm on the fence on this. I'm trying to decide how to approach this with `posthog-dotnet`. I can see the other scenario where you have a working app both locally and remote. You add this one API call and it works locally, but breaks in production because you forgot to deploy the `personal_api_key` to production. 

In that scenario, perhaps you want the app to keep running and just log the error because you'll see your new call isn't working and you can then check the logs to figure out why.

haacked: At minimum, I think we may want to specifically handle the exception when we get a 401 status and show the detail. It'll say something like "Personal API key found in request Authorization header is invalid." which is much more helpful when trying to figure out why it's not working.

haacked: Disregard my last message. I see that we are showing the exception message as part of the logged message. I need to make sure to do that in my own code. 😆

havenbarnes: Hmm, yeah that is tough. I agree that we should probably be propagating up certain types of errors. I also feel like as a developer I might only expect the SDK to provide detailed / helpful error messages to stdout, but would not expect to have to try/catch every SDK call just in case. 

I did it this way as it seemed to be the established pattern, and I think we should probably stick with it until having a conversation with the larger SDK group about it!

haacked: Yup! Sounds good to me.

haacked: fwiw, in C# I tend to do something like this:

```csharp
try {}
catch (Exception e) when (e is not ArgumentException and not NullReferenceException) {}
```

Mainly because `ArgumentException` and `NullReferenceException` are 99.99% of the time are due to a deterministic coding mistake during development and not something anyone would encounter in production unless I deployed a coding mistake. Those I want immediately apparent during development. I don't expect consumers to try/catch those because I make every effort to ensure they don't happen.

posthog/request.py
havenbarnes: We should; i thought it wasn't compatible but I read it wrong. Will update

havenbarnes: Actually it's not needed, `get` already calls `_process_response` 👍

posthog/client.py
```diff
@@ -849,6 +859,21 @@ def get_feature_flag_payload(
 
         return payload
 
+    def get_decrypted_feature_flag_payload(self, key: str):
```
haacked, line 862: Should this be named `get_decrypted_remote_config_payload` in the same way we have `get_remote_config_payload`. 

Also, I noticed this method does exactly the same thing as `get_remote_config_payload`. I'm thinking maybe we should just keep `get_remote_config_payload` as the decryption happens transparently, correct?

haacked: My point being: `get_remote_config_payload` works for all remote config settings, whether the payload is encrypted or not. The encryption/decryption is an implementation detail for the consumers of this library.

havenbarnes: Sg, will drop "decrypted" here and in the other SDKs too

PR #181, opened 2025-02-10
Stop capturing all feature flags on `$feature_flag_called` event.

haacked:
The `client.get_feature_flag` method will by default call `client.capture` with the event named `$feature_flag_called`. 

Some time ago (in e60d52c199ef75ac4e7b0aa962a2c892c94c9efc), the `capture` method was changed to enrich the event with all locally cached feature flags even if `send_feature_flags` is `False`.

What this leads to is `client.get_feature_flag` -> `client.capture` -> `client.get_all_flags`. But the call to `client.get_all_flags` isn't passed the original `person_properties` that were passed to `client.get_feature_flag`. This leads to inconsistency in the enriched data. This can also cause confusion because it's unclear which feature flag is the one being called because the event adds "all" the locally cached feature flags.

This PR takes the simplest path by not enriching the captured event with all feature flags for the `$feature_flag_called` event.

Note that this didn't show up in the tests because most of the tests of `get_feature_flag` apply `@mock.patch.object(Client, "capture")` to the test, which hides this unexpected behavior.

### DIFF
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -245,7 +245,7 @@ class Client(object):
             except Exception as e:
                 self.log.exception(f"[FEATURE FLAGS] Unable to get feature variants: {e}")
 
-        elif self.feature_flags:
+        elif self.feature_flags and event != "$feature_flag_called":
             # Local evaluation is enabled, flags are loaded, so try and get all flags we can without going to the server
             feature_variants = self.get_all_flags(
                 distinct_id, groups=(groups or {}), disable_geoip=disable_geoip, only_evaluate_locally=True
--- a/posthog/test/test_feature_flags.py
+++ b/posthog/test/test_feature_flags.py
@@ -2339,6 +2339,58 @@ class TestCaptureCalls(unittest.TestCase):
             disable_geoip=None,
         )
 
+    @mock.patch("posthog.client.decide")
+    def test_capture_is_called_but_does_not_add_all_flags(self, patch_decide):
+        patch_decide.return_value = {"featureFlags": {"decide-flag": "decide-value"}}
+        client = Client(FAKE_TEST_API_KEY, personal_api_key=FAKE_TEST_API_KEY)
+        client.feature_flags = [
+            {
+                "id": 1,
+                "name": "Beta Feature",
+                "key": "complex-flag",
+                "active": True,
+                "filters": {
+                    "groups": [
+                        {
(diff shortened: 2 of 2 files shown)


haacked:
This change should probably be applied to all the other client libraries.

dmarticus (commented):
I think this is a reasonable change, and the implementation + test makes sense.  Two follow-up questions from my perspective:

1. does this behavior exist for all SDKs (at the very least, the most popular ones that aren't python (web, posthog-js-lite, and Android)?
2. what's the documentation story behind this?  My guess is we don't have this documented; if that's the case, I think it's worth adding docs around this.  If we do have it documented, we'd need to update the docs.  Overall, it feels like docs around how we emit posthog events seems like a smart thing to do

haacked:
1. Here's what I've found:
  a. This [`getFeatureFlag` method](https://github.com/PostHog/posthog-js-lite/blob/main/posthog-node/src/feature-flags.ts#L99-L105) in https://github.com/PostHog/posthog-js-lite/ doesn't have the `sendFeatureFlags` argument.
  b.  This [`getFeatureFlag` method](https://github.com/PostHog/posthog-js-lite/blob/main/posthog-node/src/posthog-node.ts#L204-L214) appears to have the same issue. I'd have to test it to be sure. I can start on that one.
  c. The [`capture` method](https://github.com/PostHog/posthog-android/blob/main/posthog/src/main/java/com/posthog/PostHog.kt#L391-L397) in [posthog-android](https://github.com/PostHog/posthog-android/) doesn't include feature flags.
2. There doesn't seem to be any documentation about this that I could find. The [Capturing events page](https://posthog.com/docs/product-analytics/capture-events) doesn't mention the behavior. The SDK documentation for the posthog-js-lite says this:

> * @param sendFeatureFlags OPTIONAL | Used with experiments. Determines whether to send feature flag values with the event.

It may make sense to amend that to be more correct. Something like:

> * @param sendFeatureFlags OPTIONAL | Used with experiments. Determines whether to request and send up-to-date feature flag values with the event. If `false`, but locally evaluated feature flags are already cached, it will send the locally cached feature flags with the event.

I can certainly follow-up with the documentation and the other libraries. But that probably shouldn't hold up this change.

dmarticus:
Awesome, your investigation makes sense, and agreed that it's not blocking for this, just wanted to get an idea of the scope.

dmarticus (approved):

PR #184, opened 2025-02-10
Use casefold to compare strings case insensitively

haacked:
Calling `lower()` on two strings in order to compare them is incorrect in some cases (297 as of Unicode 13.0.0). For example, the lower case of `Straße` has two correct values: `straße` and `strasse`.

This PR adds a failing unit test that demonstrates the problem. Then it replaces string comparisons using `lower()` with `casefold()` to ensure the test passes.

This aligns with the official Python [docs on casefold](https://docs.python.org/3/library/stdtypes.html#str.casefold) which notes (emphasis mine): 

> Return a casefolded copy of the string. __Casefolded strings may be used for caseless matching.__

Contrast this with what the Unicode standard [Section 3.13](https://www.unicode.org/versions/Unicode13.0.0/ch03.pdf) has to say about changing the case vs case folding:

> Case folding is related to case conversion. However, the main purpose of case folding is to contribute to caseless matching of strings, whereas the main purpose of case conversion is to put strings into a particular cased form.

If this change is accepted, I can follow-up with some of the other libraries to see if they run afoul of this issue.

### DIFF
--- a/posthog/feature_flags.py
+++ b/posthog/feature_flags.py
@@ -129,8 +129,8 @@ def match_property(property, property_values) -> bool:
 
         def compute_exact_match(value, override_value):
             if isinstance(value, list):
-                return str(override_value).lower() in [str(val).lower() for val in value]
-            return str(value).lower() == str(override_value).lower()
+                return str(override_value).casefold() in [str(val).casefold() for val in value]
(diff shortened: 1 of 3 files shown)


marandaneto (approved):
Nice!

PR #186, opened 2025-02-11
Remove the usage of `is_simple_flag`

haacked:
It's no longer used anywhere. This is just cleanup code.


havenbarnes (approved):
This makes sense to me - thoughts on removing the column with a django migration?

haacked:
> This makes sense to me - thoughts on removing the column with a django migration?

I'll look into that. I wanted to remove it from the client tests first.

haacked:
> This makes sense to me - thoughts on removing the column with a django migration?

The good news is it's not a column on the `posthog_featureflag`. AFAICT, it's always been a computed property.

PR #188, opened 2025-02-11
fix: Move code under mypy type

### LINKED ISSUE #187: [3.12.0] Invalid syntax when running mypy after updating
Hi,

I'm currently getting the following error when running `mypy` after bumping posthog to `3.12.0`

```
posthog/__init__.py:54: error: invalid syntax  [syntax]
Found 1 error in 1 file (errors prevented further checking)
```

I'm assuming the issue is this change that move the type hint to an invalid location:
https://github.com/PostHog/posthog-python/commit/415508087f62845afe114f5a955fbe55988f69be#diff-c73bc092fb325121409c3eb11517c52ce313ee385528f50f109f4522a4ee1756R54

Image

Thanks in advance

rafaeelaudibert:
This is incorrect, we should've added these slightly lower in the method definition to avoid mypy from breaking

Closes #187

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,12 @@
+## 3.12.1 - 2025-02-11
+
+1. Fix mypy support for 3.12.0
+2. Deprecate `is_simple_flag`
+
 ## 3.12.0 - 2025-02-11
 
 1. Add support for OpenAI beta parse API.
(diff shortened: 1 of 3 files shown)


haacked (approved):
Thanks for fixing this!

PR #182, opened 2025-02-10
Deprecate the context argument

haacked:
It's not used anywhere, a vestigial remnant of a time long ago.

This PR supercedes https://github.com/PostHog/posthog-python/pull/54.

PR #189, opened 2025-02-11
Add mypy to CI

haacked:
A recent issue #187 noted that a recent PR (https://github.com/PostHog/posthog-python/pull/182) introduced new errors when running `mypy .`. Shout out to @rafaeelaudibert for jumping on a fix (https://github.com/PostHog/posthog-python/pull/188).

In the issue, @rafaeelaudibert  asked (emphasis mine):

> Agh, great catch, let me fix this. __I don't understand why CI didn't catch this__

Well I found the reason. CI doesn't run `mypy`. 😆 

And for good reason, it would produce almost 700 errors. However, we can try to be good citizens moving forward by not allowing new mypy errors and fix existing ones over time.

So this PR does the following:

1. Adds additional dev dependencies related to running `mypy`.
2. Adds a baseline filter file created with `mypy-baseline`
3. Adds `mypy . | mypy-baseline filter` to the CI. This should only warn on new errors.
4. Adds a `mypy.ini` so we can configure the `mypy` errors we care about.

When someone fixes old mypy errors, they'll need to run `mypy . | mypy-baseline sync` to generate a new baseline.

__Open Question for Reviewers__

- [ ] Do we care about all of the mypy warnings? Right now I have them all set to `True`.
- [ ] I ignored all errors in tests. Seems good for now, but maybe we want _some_ checking in the future?


haacked:
> I don't have any suggestions on what the setup should look like. We should try to make it look the closest possible to our posthog/posthog project, but I'm fine to just use a separate subset of checks

Good idea! I updated the mypy.ini here to look like `https://github.com/PostHog/posthog/blob/master/mypy.ini`.

PR #195, opened 2025-02-20
feat(feature-flags): support quota limiting for feature flags

dmarticus:
with [this PR](https://github.com/PostHog/posthog/pull/28564), we now start to respond different with the `/decide` and `/local_evaluation` APIs if users have gone over their quota limit.  This change is an example for how our SDKs can deal with that.  I'll make the same type of changes to all of the SDKs, but wanted to start with this one as an example.

### DIFF
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -607,6 +607,12 @@ class Client(object):
                 self.feature_flags_by_key = {}
                 self.group_type_mapping = {}
                 self.cohorts = {}
+                
+                if self.debug:
+                    raise APIError(
+                        status=402,
+                        message="PostHog feature flags quota limited",
+                    )
             else:
                 self.log.error(f"[FEATURE FLAGS] Error loading feature flags: {e}")
         except Exception as e:
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -714,6 +714,9 @@ class Client(object):
         if self.disabled:
             return None
 
+        # Normalize key to lowercase
+        key = str(key).lower()
+
         person_properties, group_properties = self._add_local_person_and_group_properties(
             distinct_id, groups, person_properties, group_properties
         )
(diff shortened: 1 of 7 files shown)


posthog/client.py
```diff
@@ -600,6 +600,13 @@ def _load_feature_flags(self):
                         "To use feature flags, please set a personal_api_key "
                         "More information: https://posthog.com/docs/api/overview",
                     )
+            elif e.status == 402:
+                self.log.warning("[FEATURE FLAGS] PostHog feature flags quota limited")
+                # Reset all feature flag data when quota limited
+                self.feature_flags = []
+                self.feature_flags_by_key = {}
+                self.group_type_mapping = {}
+                self.cohorts = {}
```
havenbarnes, line 609: nit: should we stay consistent with the 401 case above and raise an APIError if debug mode is on?

posthog/request.py
```diff
@@ -68,7 +68,17 @@ def _process_response(
...
     if res.status_code == 200:
         log.debug(success_message)
-        return res.json() if return_json else res
+        response = res.json() if return_json else res
+        # Handle quota limited decide responses by raising a specific error
+        if (
+            isinstance(response, dict)
+            and "quotaLimited" in response
+            and isinstance(response["quotaLimited"], list)
+            and "feature_flags" in response["quotaLimited"]
```
havenbarnes, line 77: nit: I'm assuming this is to cover the `/decide` case, correct? Can / should we avoid the special condition for `feature_flags` and instead log any quota limited services, maybe raising an error for any of them (or, keep a list of services that should throw an error if quota limited)

havenbarnes (approved):

posthog/client.py
dmarticus: yeah good call

posthog/request.py
dmarticus: oh yeah, good call. I think for now I explicitly want to limit feature flag events for now (since I don't know how other services are handling quota limits in the SDK).

dmarticus: I'll make a comment to that end though.

PR #204, opened 2025-03-11
fix: distro requirement for analytics package

daibhin:
@pauldambra I think you forgot to add `distro` for the analytics package

### DIFF
--- a/setup_analytics.py
+++ b/setup_analytics.py
@@ -8,13 +8,19 @@ except ImportError:
 
 # Don't import module here, since deps may not be installed
 sys.path.insert(0, os.path.join(os.path.dirname(__file__), "posthoganalytics"))
-from version import VERSION
 
 long_description = """
 PostHog is developer-friendly, self-hosted product analytics. posthog-python is the python package.
 """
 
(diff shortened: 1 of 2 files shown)


setup_analytics.py
```diff
@@ -14,7 +14,14 @@
...
 """
 
-install_requires = ["requests>=2.7,<3.0", "six>=1.5", "monotonic>=1.5", "backoff>=1.10.0", "python-dateutil>2.1"]
+install_requires = [
+    "requests>=2.7,<3.0",
+    "six>=1.5",
+    "monotonic>=1.5",
+    "backoff>=1.10.0",
+    "python-dateutil>2.1",
+    "distro>=1.5.0",  # Required for Linux OS detection in Python 3.9+
```
pauldambra, line 23: https://github.com/PostHog/posthog-python/pull/198/files#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R23

what's the difference between where I added it and this file 🙈

daibhin: We publish the `posthog` package as `posthoganalytics` so that it doesn't clash with the module name in our main Django backend. Most customers will use `posthog` which is where you changed it, we use `posthoganalytics` which this setup file relates to

pauldambra: thank you, obvious once explained :)

PR #205, opened 2025-03-13
feat: add support for responses api

k11kirky:
Supporting the new OpenAI release with support for the responses sdk


posthog/ai/openai/openai_async.py
```diff
@@ -31,6 +31,169 @@ def __init__(self, posthog_client: PostHogClient, **kwargs):
...
+                        ):
+                            usage_stats["cache_read_input_tokens"] = chunk.usage.input_tokens_details.cached_tokens
+
+                    yield chunk
+
+            finally:
+                end_time = time.time()
+                latency = end_time - start_time
+                output = final_content
+                self._capture_streaming_event(
```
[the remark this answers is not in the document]
havenbarnes: If i'm not mistaken greptile has a good catch here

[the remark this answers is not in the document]
k11kirky: thanks! just fixed it

posthog/ai/utils.py
```diff
@@ -39,20 +39,46 @@ def get_usage(response, provider: str) -> Dict[str, Any]:
         }
     elif provider == "openai":
         cached_tokens = 0
+        input_tokens = 0
+        output_tokens = 0
+        reasoning_tokens = 0
```
havenbarnes, line 44: Any desire to capture `total_tokens` here too? Seems standard across their API responses

k11kirky: we do this sum ourselves on the plugin-server

havenbarnes: Probably trivial to calculate on our end though

posthog/ai/utils.py
```diff
@@ -101,23 +175,61 @@ def format_tool_calls(response, provider: str):
...
-    if kwargs.get("system") is None:
-        return messages
-    return [{"role": "system", "content": kwargs.get("system")}] + messages
+    if provider == "anthropic":
+        messages = kwargs.get("messages") or []
+        if kwargs.get("system") is None:
+            return messages
+        return [{"role": "system", "content": kwargs.get("system")}] + messages
+
+    # For OpenAI, handle both Chat Completions and Responses API
```
havenbarnes, line 201: nit: This is pretty neat. But, for a newcomer like me it may be helpful to label which bits of logic are for chat completion vs. responses where possible

havenbarnes (approved):
Looks good!

PR #207, opened 2025-03-24
feat(flags): Add more details such as version, id, and reason to `$feature_flag_called` events

haacked:
Add support for `/decide?v=4` (see https://github.com/PostHog/posthog/pull/29751). Similar to how it's done in https://github.com/PostHog/posthog-js-lite/pull/427.

## Updates to the `$feature_flag_called` event

When capturing the `$feature_flag_called` event, additional information are now captured:

The end result is `$feature_flag_called` events will have additional properties:

| Property Name                |  Descirption
| ------------------------ | -------------
| `$feature_flag_version` | The version of the feature flag. This is visible in the diffs of the feature flag activity log.
| `$feature_flag_reason` | The reason the feature flag matched or didn't match.
| `$feature_flag_id`          | The database id of the feature flag (Just in case a flag was deleted and then recreated with the same key, we can determine the difference.

## Backwards compatibility:

The changes are all backwards compatible with `/decide?v=3`.

## Testing

Unit tests and some manual testing

### DIFF
--- a/posthog/test/ai/openai/test_openai.py
+++ b/posthog/test/ai/openai/test_openai.py
@@ -83,6 +83,7 @@ def mock_openai_response_with_responses_api():
         usage=ResponseUsage(
             input_tokens=10,
             output_tokens=10,
+            input_tokens_details={"prompt_tokens": 10, "cached_tokens": 0},
             output_tokens_details={"reasoning_tokens": 15},
             total_tokens=20,
         ),
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+## 3.22.0 – 2025-03-26
+
+1. Add more information to `$feature_flag_called` events.
+2. Support for the `/decide?v=3` endpoint which contains more information about feature flags.
+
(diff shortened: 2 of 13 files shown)


posthog/test/ai/openai/test_openai.py
```diff
@@ -83,6 +83,7 @@ def mock_openai_response_with_responses_api():
         usage=ResponseUsage(
             input_tokens=10,
             output_tokens=10,
+            input_tokens_details={"prompt_tokens": 10, "cached_tokens": 0},
```
dmarticus, line 86: was this a missing type or something?  Feels not in scope with what we're doing lol

dmarticus (approved):
this is just good clean fun

posthog/test/ai/openai/test_openai.py
haacked: Yeah, mypy kept complaining so I fixed it.

haacked:
I’ll merge after bumping minor version and adding changelog notes

PR #215, opened 2025-04-10
feat: log captured exceptions

daibhin:
Adds a `log_captured_exceptions` config option so that all `$exception` events can also be logged

This is useful internally to capture exceptions if our own product is unavailable.

We have a version of this already in https://github.com/PostHog/posthog/blob/c2c65827b31b13c8e2e84d4ddc33df34b5924487/posthog/exceptions.py#L58-L66 but that only applies to DFR views.

By doing this at the SDK level we can log _any_ exception that is being captured. One difference is that the logger in the SDK is not structured but I think that's ok

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 3.24.0 – 2025-04-10
+
+1. Add config option to `log_captured_exceptions`
+
 ## 3.23.0 – 2025-03-26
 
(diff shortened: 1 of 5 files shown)


daibhin:
Currently blocked by failing tests. Fix in https://github.com/PostHog/posthog-python/pull/216

timgl (approved):

PR #216, opened 2025-04-10
fix failing ai test

daibhin:
Don't know what has changed because the last merged PR passed all tests without issues

- Off by one error on AI calls
- Linting rules causing flake8 issues

Tagging @PostHog/team-max-ai as @k11kirky is off


skoob13 (approved):
Thank you! I suspect it happened because the version of LangGraph/LangChain is not fixed. I'll check if they had a breaking change.

PR #220, opened 2025-04-15
ubuntu 20.04 is EOL

dmarticus:
updating our release job for `posthog-python` since GH actions have officially discontinued 20.04.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -6,7 +6,7 @@ on:
 jobs:
     release:
         name: Publish release
-        runs-on: ubuntu-20.04
+        runs-on: ubuntu-latest
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -6,7 +6,7 @@ on:
 jobs:
     release:
         name: Publish release
-        runs-on: ubuntu-20.04
+        runs-on: ubuntu-latest
```
dmarticus, line 9: any reason not to just jump to `latest` here?  Our release process is super straightforward.

lricoy: That's the `latest LTS,` so it should be fine, as Ubuntu is not as frequently updated as others.

lricoy (approved):

PR #223, opened 2025-04-18
chore: Make `condition_index` optional. Also added some scripts for local dev.

haacked:
See https://github.com/PostHog/posthog/issues/30569 for why I'm making this change to `condition_index`.

See PostHog/meta#309 for the rationale behind these scripts.


rafaeelaudibert (approved):
low-context stamp

PR #227, opened 2025-04-23
feat: Add new FeatureFlagResult class and tests

haacked:
Added new method `get_feature_flag_result` which returns a `FeatureFlagResult` object. This object breaks down the result of a feature flag into its enabled state, variant, and payload. The benefit of this method is it allows you to retrieve the result of a feature flag and its payload in a single API call. You can call `get_value` on the result to get the value of the feature flag, which is the same value returned by `get_feature_flag` (aka the string `variant` if the flag is a multivariate flag or the `boolean` value if the flag is a boolean flag).

Example:

```python
result = posthog.get_feature_flag_result("my-flag", "distinct_id")
print(result.enabled)     # True or False
print(result.variant)     # 'the-variant-value' or None
print(result.payload)     # {'foo': 'bar'}
print(result.get_value()) # 'the-variant-value' or True or False
```

🚨  Breaking change ⚠️ :

`get_feature_flag_payload` now deserializes payloads from JSON strings to `Any`. Previously, it returned the payload as a JSON encoded string.

Before:

```python
payload = get_feature_flag_payload('key', 'distinct_id') # "{\"some\": \"payload\"}"
```

After:

```python
payload = get_feature_flag_payload('key', 'distinct_id') # {"some": "payload"}
```


posthog/types.py
```diff
@@ -102,6 +103,68 @@ class FlagsAndPayloads(TypedDict, total=True):
...
+@dataclass(frozen=True)
+class FeatureFlagResult:
+    """
+    The result of calling a feature flag which includes the flag result, variant, and payload.
+    """
+
+    key: str
+    enabled: bool
+    variant: Optional[str]
+    payload: Optional[Any]
```
dmarticus, line 115: seems good; is it worth including the reason here, too?  obviously we capture it but maybe (especially in the case of failure) it'd be nice to expose that to the caller.

dmarticus (approved):
great as always, one question about what we decided to return in the `FeatureFlagResponse` but nothing blocking.

posthog/types.py
haacked: Yeah, do you think I should just include the reason description, or the whole reason object?

dmarticus: the description string feels sufficient.

PR #235, opened 2025-05-22
fix: feature flag request use geoip_disable

orian:
it incorrectly used disable_geoip


orian:
Changing it everywhere breaks the API and would require some more hacking around

Issue #234, opened 2025-05-08
Incompatible with OpenAI 1.78.0


drupol:
Hello,

There is an incompatibility with OpenAI Python package 1.78.0:

```
============================= test session starts ==============================
platform linux -- Python 3.12.10, pytest-8.3.5, pluggy-1.5.0
rootdir: /build/source
configfile: setup.cfg
plugins: anyio-4.9.0
collected 226 items / 1 error / 30 deselected / 2 skipped / 196 selected       

==================================== ERRORS ====================================
____________ ERROR collecting posthog/test/ai/openai/test_openai.py ____________
posthog/test/ai/openai/test_openai.py:19: in <module>
    from posthog.ai.openai import OpenAI
posthog/ai/openai/__init__.py:1: in <module>
    from .openai import OpenAI
posthog/ai/openai/openai.py:38: in <module>
    class WrappedResponses(openai.resources.responses.Responses):
E   AttributeError: module 'openai.resources' has no attribute 'responses'
=============================== warnings summary ===============================
../../nix/store/094gpcj7d1z8wdvymhxvi2rns757vb2p-python3.12-pytest-8.3.5/lib/python3.12/site-packages/_pytest/config/__init__.py:1441
  /nix/store/094gpcj7d1z8wdvymhxvi2rns757vb2p-python3.12-pytest-8.3.5/lib/python3.12/site-packages/_pytest/config/__init__.py:1441: PytestConfigWarning: Unknown config option: asyncio_mode
  
    self._warn_or_fail_if_strict(f"Unknown config option: {key}\n")

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
=========================== short test summary info ============================
ERROR posthog/test/ai/openai/test_openai.py - AttributeError: module 'openai.resources' has no attribute 'responses'
!!!!!!!!!!!!!!!!!!!! Interrupted: 1 error during collection !!!!!!!!!!!!!!!!!!!!
============ 2 skipped, 30 deselected, 1 warning, 1 error in 0.96s =============
error: builder for '/nix/store/6l9j1q9010kxx4wwqy5wdkfhy4fyygik-python3.12-posthog-4.0.1.drv' failed with exit code 2;
       last 50 log lines:
       > Running phase: pythonOutputDistPhase
       > Executing pythonOutputDistPhase
       > Finished executing pythonOutputDistPhase
       > Running phase: fixupPhase
[...]

PR #236, opened 2025-05-22
feat: composition over inheritance

k11kirky:
Support both new and old openai sdks (since resource change) by changing to a composition approach over inheritance. This means we do not need to import the resources and should be more defensive to dependency changes

fixes: https://github.com/PostHog/posthog-python/issues/234

### DIFF
--- a/posthog/ai/openai/openai.py
+++ b/posthog/ai/openai/openai.py
@@ -4,7 +4,6 @@ from typing import Any, Dict, List, Optional
 
 try:
     import openai
-    import openai.resources
 except ImportError:
     raise ModuleNotFoundError("Please install the OpenAI SDK to use this feature: 'pip install openai'")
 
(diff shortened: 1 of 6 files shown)


posthog/ai/openai/openai.py
```diff
@@ -29,14 +28,38 @@ def __init__(self, posthog_client: PostHogClient, **kwargs):
...
+        if self._original_chat is not None:
+            self.chat = WrappedChat(self, self._original_chat)
+
+        if self._original_embeddings is not None:
+            self.embeddings = WrappedEmbeddings(self, self._original_embeddings)
+
+        if self._original_beta is not None:
+            self.beta = WrappedBeta(self, self._original_beta)
+
+        # Only add responses if available (newer OpenAI versions)
```
rafaeelaudibert, line 48: Useless comment, the above comment explains it

posthog/ai/openai/openai.py
```diff
@@ -29,14 +28,38 @@ def __init__(self, posthog_client: PostHogClient, **kwargs):
...
+
+
+class WrappedResponses:
+    """Wrapper for OpenAI responses that tracks usage in PostHog."""
+
+    def __init__(self, client: OpenAI, original_responses):
+        self._client = client
+        self._original = original_responses
+
+    def __getattr__(self, name):
```
rafaeelaudibert, line 60: Will this be called only in case there's no method, or is this called all the time? Do we have unit tests for this wrapper?

posthog/ai/openai/openai_async.py
```diff
@@ -394,8 +452,6 @@ async def create(
             posthog_distinct_id: Optional ID to associate with the usage event.
             posthog_trace_id: Optional trace UUID for linking events.
             posthog_properties: Optional dictionary of extra properties to include in the event.
-            posthog_privacy_mode: Whether to store input and output in PostHog.
```
rafaeelaudibert, line 397: Are we removing args or just docs were wrong? If the former, we'd need a major release. I'll assume it's the latter

rafaeelaudibert (approved):
Not expert in this code, but makes sense. We'd benefit from tests showing this works as expected (the fallback behaviour)

posthog/ai/openai/openai.py
k11kirky: we do have unit tests, just not integration tests that will run with different versions. i have tested that manually.

that method is only called if we do not have an override. it allows us to fall back to the original client if we do not support

posthog/ai/openai/openai.py
k11kirky: fair. Thats the ai adding random comments that i didnt reject like i usually do

posthog/ai/openai/openai_async.py
k11kirky: this was incorrect docs. fixing

k11kirky:
Agree with your comments @rafaeelaudibert I am the only one that works on this so far and have a local test file that i moved out of this repo (without pyttest). will move back in and join with pytest soon

PR #239, opened 2025-06-05
feat(err): add context manager and tag functions

oliverb123:
We've been hacking around this internally with `tag_queries` for a while, but we should just actually support it properly in the SDK.


example.py
```diff
@@ -104,4 +104,27 @@
...
+# Recommended: you can use the `@posthog.scoped` decorator to enter a new context.
+@posthog.scoped
+def process_order(order_id):
+    posthog.tag("order_id", order_id)
+    # Exception will be captured and tagged automatically
+    raise Exception("Order processing failed")
+
+
+# You can also use the directly context manager, but you should prefer to use the decorator where possible,
+# because it's tricky to remember to always capture the exception within the context scope.
```
daibhin, line 117: Is there any reason not to autocapture exceptions here too? Or at least have `posthog.new_context(enable_exception_autocapture=True)` as an argument

posthog/client.py
```diff
@@ -662,6 +663,16 @@ def capture_exception(
         try:
             properties = properties or {}
 
+            # Check if this exception has already been captured
+            if exception is not None and hasattr(exception, "__posthog_exception_captured"):
+                self.log.debug("Exception already captured, skipping")
+                return
```
daibhin, line 669: What would happen in the case where I have multiple nested contexts? It obviously captures the exception within the most deeply nested context - does that have the tags from the parent contexts? Based on [this test](https://github.com/PostHog/posthog-python/pull/239/files#diff-48c5cc7539da673a409bdba411d3ade7a6ada43eef4b8d67561b533488b5f8f7R48-R62) I guess not but I would have expected it to

example.py
oliverb123: Mostly that it wasn't obvious to me how to do that within a context manager? I don't think context managers can try/catch the inner code the way function decorators can, and I'm not sure they get access to whether the block is being exited due to an exception or some other control flow

posthog/client.py
oliverb123: Yeah, contexts are mutually exclusive. It's not a huge change to make them "fold" the tags from all parent contexts  downwards, but I guess I wonder why you'd expect that? It makes it a lot trickier for a user to control what tags are on a given exception (they have to think about all the scopes potentially above them, rather than just the one they just created).

posthog/scopes.py
```diff
@@ -0,0 +1,112 @@
...
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        with new_context():
+            try:
+                return func(*args, **kwargs)
+            except Exception as e:
+                # Capture the exception with current context tags
+                # The capture_exception function will handle deduplication
+                posthog.capture_exception(e, properties=get_tags())
+                raise  # Re-raise the exception after capturing it
```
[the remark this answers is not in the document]
daibhin: I guess it depends on how broadly we consider the term "scoped". For me it blends into a tracing product (which is great) where you would have "spans". I would expect every event (exception or otherwise) captured within a span to contain all tags from that context and its parents

example.py
oliverb123: Done, was being silly

posthog/client.py
daibhin: I think my mental model is that all the the data from parent contexts is accessible. That's probably a mix of React contexts / Kea bound logics and other tracing providers I've used in the past. It's handy if you want to add properties like user_email where it's accessible at the root level (or maybe something like `query_runner_type` to be very domain specific) and not have to pass that through to children further down. I know it could technically clobber values higher up but it feels like a feature in my mind to be able to override parent context values with more specific alternatives

oliverb123: With tracing products/otel, my general understanding (haven't built one... yet...) each span has tags that are only the tags of that span, and then the overall trace is the sum of the tags in all spans.

The thing is we're not building a tracing product - if we did, we'd still use mutually exclusive scopes for each spans tags, but we'd flatten the tags into the trace (depending on how it was build) - which is exactly what we do here.

posthog/scopes.py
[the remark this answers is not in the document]
oliverb123: Sorry, don't see how this is related to:
> We should probably be checking if a customer has exception autocapture enabled before capturing exceptions as part of the context

posthog/client.py
oliverb123: Did this, made it inherit by default

example.py
```diff
@@ -104,4 +104,39 @@
 print(posthog.get_remote_config_payload("encrypted_payload_flag_key"))
 
 
+# You can add tags to a context, and these are automatically added to exceptions captured within that context.
+# You can enter a new context using a with statement. Any exceptions thrown in the context will be captured,
+# and tagged with the context tags. By default, it inherits tags from the parent context.
+with posthog.new_context():
+    posthog.tag("user_id", 123)
```
pauldambra, line 111: i think with the name `new_context` and `tag` not referencing errors or exceptions i'd expect them to apply to any capture call with the scope

should they? that'd be a super cool function - like JS's register super properties (does python have super properties) this is a nice way to add them

alternatively

should they be called `new_error_tracking_context` and `error_tag`?

posthog/test/test_scopes.py
```diff
@@ -0,0 +1,138 @@
...
+
+        try:
+            with new_context():
+                tag("inner_context", "inner_value")
+                raise test_exception
+        except RuntimeError:
+            pass  # Expected exception
+
+        # Verify capture_exception was called
+        mock_capture.assert_called_once_with(test_exception)
```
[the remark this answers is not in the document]
oliverb123: No, capture_exception calls `get_tags` internally and adds them to the property set, and I'm comfortable not covering that with a test.

posthog/test/test_scopes.py
```diff
@@ -0,0 +1,138 @@
...
+        @scoped()
+        def successful_function(x, y):
+            tag("x", x)
+            tag("y", y)
+            return x + y
+
+        result = successful_function(1, 2)
+
+        # Function should execute normally
+        self.assertEqual(result, 3)
```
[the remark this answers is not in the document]
oliverb123: Done

example.py
oliverb123: The python SDK has super props afaik, yeah

I think contexts/scopes should be used elsewhere, in probably all of our server side SDKs, but I'd rather ship incrementally, and if other product teams want, they can adopt and extend.

`new_error_tracking_context ` is too verbose, and I don't really think name-spacing off error tracking here is actually useful, and it isn't a pattern we follow elsewhere - we don't call them `product_analytics_super_properties`, for example.

pauldambra: > we don't call them product_analytics_super_properties, for example.

because they apply to everything captured not just product analytics

whereas `posthog.new_context` isn't a posthog context, it's a posthog error tracking context 

🤷 

this would surprise me as an SDK consumer

i'm guessing that capture exception is built on top of capture

so doesn't seem a stretch to add this so it applies to any capture

can make that an argument to the context but the current name makes it feel like it should apply to everything and then be filtered down from that

> too verbose...

call it `new_error_context` or `error_context` 
i'm not fixed on any one name but `new_context` sounds hugely like it applies everywhere

oliverb123: Moved tag enriching to capture.

posthog/test/test_scopes.py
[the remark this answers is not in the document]
pauldambra: > I'm comfortable not covering that with a test.

bookmarking for future "i told you so"

🤣

posthog/client.py
```diff
@@ -466,6 +467,11 @@ def capture(
         require("properties", properties, dict)
         require("event", event, string_types)
 
+        # Grab current context tags, if any exist
+        context_tags = get_tags()
+        if context_tags:
+            properties.update(context_tags)
```
pauldambra, line 473: this placement is cool
since it looks to me like i can set a super property
and then use context to override it sometimes

that is pleasing

pauldambra (approved):
i'm invested in getting my other PR in so 🚢 🚢 🚢 🚢 🚢 

(thanks for moving context to capture, i think that will super powerful)

daibhin (approved):
Epic stuff!! Think a lot of my questions & points you worked through with Paul yesterday.

Left one comment about whether tags should live on the client. I'm sure whatever we choose someone will want the alternative. Also happy to go with the approach as it is now for simplicity sake and can revisit if necessary

posthog/scopes.py
```diff
@@ -0,0 +1,120 @@
+import contextvars
```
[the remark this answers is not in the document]
oliverb123: We should be using contextvars I think, but since we don't use async python in our django app (as far as I know?) I think it doesn't make a huge difference.

posthog/__init__.py
```diff
@@ -4,11 +4,19 @@
...
 from posthog.version import VERSION
 
 __version__ = VERSION
 
+"""Context management."""
+new_context = new_context
+tag = tag
+get_tags = get_tags
+clear_tags = clear_tags
+tracked = scoped
```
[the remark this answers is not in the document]
oliverb123: I think we'd have to pass a reference to the client in at the callsite to entering a new context to do this. It's an extension I'd consider later, maybe, but generally I think the model of "when you call capture, all the tags you've set get added" is simpler to keep in your head than "when you call capture, all the tags you've set get added, depending on the client you're using and the ones you passed to construct the context with" - particularly in the face of e.g. the way context variables get inherited (if I'm in context A associated with client 1 and I construct context B and associate it with client 2, does context B start with context A's tags?).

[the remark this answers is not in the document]
oliverb123: https://github.com/PostHog/posthog-python/pull/257

PR #240, opened 2025-06-05
fix: opinionated setup and clean fn fix

pauldambra:
copyable setup block in readme so i don't have to figure it out again in future

we also had some python 3 unsafe code that we're luckily only calling in a safe way right now that mypy was incidentally warning us about. should do a patch release with this change

### DIFF
--- a/posthog/test/test_utils.py
+++ b/posthog/test/test_utils.py
@@ -7,6 +7,7 @@ from uuid import UUID
 
 import six
 from dateutil.tz import tzutc
+from parameterized import parameterized
 from pydantic import BaseModel
 from pydantic.v1 import BaseModel as BaseModelV1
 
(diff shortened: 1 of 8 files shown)


posthog/test/test_utils.py
pauldambra, line 1: might want to go commit  by commit
i tidied this up as well as fixing some assertions

posthog/test/test_utils.py
```diff
@@ -101,19 +115,26 @@ class ModelV1(BaseModelV1):
...
+            def model_dump(self, required_param: str) -> dict:
+                return {}
 
-        # Skips a class with a defined non-Pydantic `model_dump` method.
-        self.assertEqual(utils.clean({"test": Dummy()}), {})
+        # previously python 2 code would cause an error while cleaning,
+        # and this entire object would be None, and we would log an error
+        # let's allow ourselves to clean `Dummy` as None,
+        # without blatting the `test` key
+        assert utils.clean({"test": Dummy()}) == {"test": None}
```
pauldambra, line 137: this test was hiding that the behaviour was determined by throwin an exception in an exception handler
and then swallowing that exception

posthog/utils.py
```diff
@@ -99,14 +100,35 @@ def _clean_dataclass(dataclass_):
...
-        item = ":".join(exception)
-        item.decode("utf-8", "strict")
+        if isinstance(cmplx, bytes):
+            item = cmplx.decode("utf-8", "strict")
+        elif isinstance(cmplx, str):
+            item = cmplx
+    except Exception as exception:
+        item = ":".join(map(str, exception.args))
         log.warning("Error decoding: %s", item)
         return None
```
pauldambra, line 130: not super convinced this logging behaviour is correct, but considering it was always throwing before now, it can't be too bad

PR #245, opened 2025-06-06
feat: add setup method

daibhin:
We cannot guarantee that the default PostHog client is always created. This means things like exception autocapture handlers will not get set until some other function creates the client e.g. capture. Let's expose a function to setup the client for use in environments like Celery where we could potentially need to capture an exception event if no events are being captured


pauldambra (approved):
LGTM

daibhin:
Kept hitting rebase issues with the changelog so folded this into https://github.com/PostHog/posthog-python/pull/252

PR #249, opened 2025-06-09
feat: add before_send callback

pauldambra:
we've got this in posthog-js and posthog-ruby
let's add it here

ultimate power to edit/drop/fix events as you see fit

### DIFF
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -144,6 +144,7 @@ class Client(object):
         exception_autocapture_integrations=None,
         project_root=None,
         privacy_mode=False,
+        before_send=None,
     ):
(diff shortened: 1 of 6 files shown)


posthog/client.py
```diff
@@ -144,6 +144,7 @@ def __init__(
         exception_autocapture_integrations=None,
         project_root=None,
         privacy_mode=False,
+        before_send=None,
```
pauldambra, line 147: interestingly adding a type to this parameter kicked mypy into being strict about the other parameters and we got ~14 errors

i'm going to punt on that for now since this PR doesn't introduce them

robbie-c (approved):
Aren't you on vacation today?

PR #253, opened 2025-06-09
fix changelog to match pypi

pauldambra:
now that the release process is fixed we can update the changelog so that it matches pypi

https://pypi.org/project/posthog/#history


daibhin (approved):
Will let this get merged ahead of my PR: https://github.com/PostHog/posthog-python/pull/252

PR #254, opened 2025-06-09
fix: makefile import

daibhin:
Getting an error when trying to use the new context scopes feature because we call `import posthog`. This fails because the Makefile doesn't replace the call with `import posthoganalytics`


pauldambra (approved):
👍

PR #255, opened 2025-06-10
remove 'import posthog'

daibhin:
Follow on from https://github.com/PostHog/posthog-python/pull/254

That PR had the intended effect of replacing `import posthog` with `import posthoganalytics`. I failed to realise that everywhere `posthog.{method_name}` was called also needed replacing. Going to try this simpler approach of never importing the entire package

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 4.6.2 - 2025-06-09
+
+- fix: replace `import posthog` with direct method imports
+
 ## 4.6.1 - 2025-06-09
 
 - fix: replace `import posthog` in `posthoganalytics` package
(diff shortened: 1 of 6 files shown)


posthog/scopes.py
```diff
@@ -39,7 +39,7 @@ def new_context(fresh=False):
              raise ValueError("Something went wrong")
 
     """
-    import posthog
```
pauldambra, line 42: would be cool if we can have ruff forbid this in future

CHANGELOG.md
pauldambra, line 1: oh.. @daibhin i updated setup instructions in the readme
you need to run a command to setup the precommit hooks and then ruff will autorun on commit for you

PR #257, opened 2025-06-11
fix: change scoped export, add capturing param

oliverb123:
Figured while I was here we may as well let people use the tagging infrastructure without forcing them to use error tracking - it seems like a nice pattern for using posthog generally.

I do think we should default opt-in to error tracking tho - if we get complaints, we can expose a global config to disable it at the client level, but capturing exceptions in the most deeply nested context means getting the most tags on them, which is the ideal scenario.

### DIFF
--- a/posthog/scopes.py
+++ b/posthog/scopes.py
@@ -12,31 +12,34 @@ def _get_current_context() -> Dict[str, Any]:
 
 
 @contextmanager
-def new_context(fresh=False):
+def new_context(fresh=False, capturing=True):
     """
-     Create a new context scope that will be active for the duration of the with block.
-     Any tags set within this scope will be isolated to this context. Any exceptions raised
+    Create a new context scope that will be active for the duration of the with block.
+    Any tags set within this scope will be isolated to this context. Any exceptions raised
     or events captured within the context will be tagged with the context tags.
 
-     Args:
-         fresh: Whether to start with a fresh context (default: False).
-                If False, inherits tags from parent context.
-                If True, starts with no tags.
+    Args:
+        fresh: Whether to start with a fresh context (default: False).
+               If False, inherits tags from parent context.
+               If True, starts with no tags.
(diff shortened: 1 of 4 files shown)


posthog/scopes.py
```diff
@@ -49,7 +52,8 @@ def new_context(fresh=False):
     try:
         yield
     except Exception as e:
-        capture_exception(e)
+        if capturing:
```
daibhin, line 55: We could default to `enable_exception_autocapture` if this is not set

posthog/scopes.py
```diff
@@ -12,31 +12,34 @@ def _get_current_context() -> Dict[str, Any]:
 
 
 @contextmanager
-def new_context(fresh=False):
+def new_context(fresh=False, capturing=True):
```
daibhin, line 15: It wasn't immediately obvious to me that `capturing` had anything to do with exceptions, I thought it might disable event capture inside a context. Maybe something like `handle_exceptions`?

oliverb123: I think `handle` implies it doesn't re-throw, and was leaning towards shorter names purely as a personal bias. Will switch to `capture_exceptions`.

posthog/scopes.py
oliverb123: Huh, for some reason I thought that was client specific?

daibhin: It can be globally set using the proxy if the customer has configured their setup like
```
import posthog

posthog.enable_exception_autocapture = True
```

If that is None we should default (probably to True)

oliverb123: I think, given if we deferred to the global setting it'd always be off, it's good enough for now to default to on and let users disable. We can revisit it later, if we decide it's warranted

More broadly, I dislike complex behaviours and global configurations taking precedence over local ones - there may be good reasons a user wants to say "absolutely do not autocapture exceptions at this scope level", and we should respect it, imo.

PR #260, opened 2025-06-13
fix(err): add synthetic exception, use it to prevent attributeerrors

oliverb123:


oliverb123:
Closed in favour of https://github.com/PostHog/posthog-python/pull/261

PR #262, opened 2025-06-16
fix: delete sentry integration

oliverb123:
We don't support this mode of using posthog error tracking, so this code is ~entirely unmaintained. I'm fairly sure if anyone /did/ try to use it, it'd send invalid data to posthog, which our pipeline would drop. Tearing it out as part of general SDK improvement work, first step in improving our django middleware/integration offering.


posthog/version.py
```diff
@@ -1,4 +1,4 @@
-VERSION = "4.10.0"
+VERSION = "5.0.0"
```
daibhin, line 1: I know this is technically a breaking change but seems like a pretty niche feature to necessitate a major upgrade for 🤔 Don't suppose there's anything we could bundle in as part of this change like we did for Node recently?

hpouillot (approved):

posthog/version.py
oliverb123: Version numbers are free, they let you make them as high as you want, even the major ones 😉.

More seriously, the next breaking change I want to make to this SDK is to make it optional to pass a distinct ID, and that's too large a change to bundle in here IMO. I'm not aware of anyone else working on anything breaking, and I think "biting the bullet" of major version bump for a relatively niche feature here wins us some credibility about future breaking changes.

PR #263, opened 2025-06-17
feat(err): add django middleware

oliverb123:
The distinct_id bit is currently kind-of unused - I'm going to add additional machinery and handling for it in future PRs

### DIFF
--- a/posthog/exception_utils.py
+++ b/posthog/exception_utils.py
@@ -787,7 +787,7 @@ def exception_is_already_captured(error):
         return error[2] is not None and hasattr(
             error[2], "__posthog_exception_captured"
         )
-    else:
+    else:  # type: ignore[unreachable]
         return False
 
 
--- a/posthog/exception_utils.py
+++ b/posthog/exception_utils.py
@@ -783,9 +783,11 @@ def exception_is_already_captured(error):
     # type: (Union[BaseException, ExcInfo]) -> bool
     if isinstance(error, BaseException):
         return hasattr(error, "__posthog_exception_captured")
-    elif isinstance(error, tuple) and len(error) == 3:
-        return error[2] is not None and hasattr(
-            error[2], "__posthog_exception_captured"
+    # Autocaptured exceptions are passed as a tuple from our system hooks,
+    # the second item is the exception value (the first is the exception type)
+    elif isinstance(error, tuple) and len(error) > 1:
+        return error[1] is not None and hasattr(
+            error[1], "__posthog_exception_captured"
         )
     else:
         return False  # type: ignore[unreachable]
@@ -795,9 +797,11 @@ def mark_exception_as_captured(error):
     # type: (Union[BaseException, ExcInfo]) -> None
     if isinstance(error, BaseException):
         setattr(error, "__posthog_exception_captured", True)
-    elif isinstance(error, tuple) and len(error) == 3:
-        if error[2] is not None:
-            setattr(error[2], "__posthog_exception_captured", True)
+    # Autocaptured exceptions are passed as a tuple from our system hooks,
+    # the second item is the exception value (the first is the exception type)
+    elif isinstance(error, tuple) and len(error) > 1:
+        if error[1] is not None:
+            setattr(error[1], "__posthog_exception_captured", True)
 
 
 def exc_info_from_error(error):
--- a/posthog/exception_utils.py
+++ b/posthog/exception_utils.py
@@ -787,8 +787,8 @@ def exception_is_already_captured(error):
         return error[2] is not None and hasattr(
             error[2], "__posthog_exception_captured"
         )
-    else:  # type: ignore[unreachable]
-        return False
+    else:
+        return False  # type: ignore[unreachable]
 
 
 def mark_exception_as_captured(error):
--- a/posthog/exception_utils.py
+++ b/posthog/exception_utils.py
@@ -779,6 +779,27 @@ def set_in_app_in_frames(frames, in_app_exclude, in_app_include, project_root=No
     return frames
 
 
+def exception_is_already_captured(error):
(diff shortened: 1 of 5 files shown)


hpouillot (approved):

posthog/exception_utils.py
```diff
@@ -779,6 +779,27 @@ def set_in_app_in_frames(frames, in_app_exclude, in_app_include, project_root=No
     return frames
 
 
+def exception_is_already_captured(error):
+    # type: (Union[BaseException, ExcInfo]) -> bool
+    if isinstance(error, BaseException):
+        return hasattr(error, "__posthog_exception_captured")
+    elif isinstance(error, tuple) and len(error) == 3:
```
daibhin, line 786: Not immediately obvious to me why this would be a tuple. Worth adding a comment?

posthog/integrations/django.py
```diff
@@ -0,0 +1,116 @@
...
+    from django.http import HttpRequest, HttpResponse  # noqa: F401
+    from typing import Callable, Dict, Any, Optional  # noqa: F401
+
+
+class PosthogContextMiddleware:
+    """Middleware to automatically track Django requests.
+
+    This middleware wraps all calls with a posthog scope. It attempts to extract the following from the request headers:
+    - Session ID as $session_id, (extracted from `X-POSTHOG-SESSION-ID`)
+    - Distinct ID as $distinct_id, (extracted from `X-POSTHOG-DISTINCT-ID`)
```
daibhin, line 14: Is this set in our SDK? I didn't see it referenced anywhere but maybe it's coming in a separate PR. I know we append `X-POSTHOG-SESSION-ID` when tracing headers are enabled

posthog/integrations/django.py
```diff
@@ -0,0 +1,116 @@
...
+    - Request URL as $current_url
+    - Request Method as request_method
+
+    The scope will also auto-capture exceptions and send them to PostHog, unless you disable it by setting
+    `POSTHOG_MW_CAPTURE_EXCEPTIONS` to `False` in your Django settings.
+
+    The middleware behaviour is customisable through 3 additional functions:
+    - `POSTHOG_MW_EXTRA_TAGS`, which is a Callable[[HttpRequest], Dict[str, Any]] expected to return a dictionary of additional tags to be added to the scope.
```
daibhin, line 22: Should we be calling them `properties`? Tags are not a concept we use / mention elsewhere
```suggestion
    - `POSTHOG_MW_EXTRA_PROPERTIES`, which is a Callable[[HttpRequest], Dict[str, Any]] expected to return a dictionary of additional tags to be added to the scope.
```

posthog/integrations/django.py
```diff
@@ -0,0 +1,116 @@
...
+    """Middleware to automatically track Django requests.
+
+    This middleware wraps all calls with a posthog scope. It attempts to extract the following from the request headers:
+    - Session ID as $session_id, (extracted from `X-POSTHOG-SESSION-ID`)
+    - Distinct ID as $distinct_id, (extracted from `X-POSTHOG-DISTINCT-ID`)
+    - Request URL as $current_url
+    - Request Method as request_method
+
+    The scope will also auto-capture exceptions and send them to PostHog, unless you disable it by setting
+    `POSTHOG_MW_CAPTURE_EXCEPTIONS` to `False` in your Django settings.
```
daibhin, line 19: How does this work alongside the exceptions Django integration? Is it possible that we could double capture exceptions? Maybe we should be folding them both into one

posthog/integrations/django.py
```diff
@@ -0,0 +1,116 @@
...
+            tags["$distinct_id"] = distinct_id
+
+        # Extract current URL
+        absolute_url = request.build_absolute_uri()
+        if absolute_url:
+            tags["$current_url"] = absolute_url
+
+        # Extract request method
+        if request.method:
+            tags["request_method"] = request.method
```
daibhin, line 93: Should we append the `$` given this will go in our taxonomy? We call it `$request_path` in the exception integration
```suggestion
            tags["$request_method"] = request.method
```

posthog/integrations/django.py
```diff
@@ -0,0 +1,116 @@
...
+
+
+class PosthogContextMiddleware:
+    """Middleware to automatically track Django requests.
+
+    This middleware wraps all calls with a posthog scope. It attempts to extract the following from the request headers:
+    - Session ID as $session_id, (extracted from `X-POSTHOG-SESSION-ID`)
+    - Distinct ID as $distinct_id, (extracted from `X-POSTHOG-DISTINCT-ID`)
+    - Request URL as $current_url
+    - Request Method as request_method
```
daibhin, line 16: Worth adding request_action too?

posthog/integrations/django.py
oliverb123: I used tags because they're scope/context tags (which we do call `tags()`)... I'll have a think about naming a bit more (I also mix "scope" and "context" in a way that's unhelpful), and either change it here or, more likely, make some breaking change in the next major version bump (which will come in a few days).

posthog/integrations/django.py
oliverb123: what's a request_action?

posthog/integrations/django.py
oliverb123: Updated the comment to talk in terms of tags and contexts, included a note about how context tags end up as properties on events.

posthog/integrations/django.py
daibhin: Fair. Deffo think we should consolidate

posthog/integrations/django.py
daibhin: This is all it should be I think: https://github.com/PostHog/posthog-js/pull/2026

PR #269, opened 2025-06-20
chore: ensure session_id context works with page method

lricoy:
I am investigating some tickets (e.g, https://posthoghelp.zendesk.com/agent/tickets/32213 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/34980)) that send events on both frontend and backend, and wish to keep the sessionId in sync. 

The changes here build upon https://github.com/PostHog/posthog-python/pull/264 (thanks @oliverb123, great work!) to ensure we can also keep the session_id in context for the `page` utility function, as well as some redundant tests to verify that we're passing it on to other  methods.

### DIFF
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -651,6 +651,9 @@ class Client(object):
         require("distinct_id", distinct_id, ID_TYPES)
         require("properties", properties, dict)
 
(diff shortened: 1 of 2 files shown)


rafaeelaudibert (approved):
That makes a lot of sense to me, LGTM :)

PR #273, opened 2025-06-24
feat: prep for 6.0.0, bunch of breaking changes

oliverb123:
This change:
- Makes `distinct_id` an optional argument to most event-capturing functions
- Changes the order of arguments to most event-capturing functions
- Uses kwargs to make it impossible to upgrade to 6.0.0 without modifying how you call these functions
- Changes these functions to return the UUID of the captured event, rather than the event contents
- Deletes `identify`, `page` and `screen`  - `identify` is `set` with a misleading name, `page` and `screen` are just `capture`
- Deletes unmaintained exception-capture integrations (users should prefer the general purpose django middleware, and we should prefer to build general purpose integrations)

Generally, the goal of this PR is to move the python SDK towards a place where we can use it as the "template" for all server-side SDKs - focusing on semantic scope based state management, simplified interfaces, and reducing how much we expose the internals of our libraries, reducing the cost of future refactoring and improvements to our users. It's the culmination of the previous PR's I've made re: nested contexts, context-based identity and session management, etc.


CHANGELOG.md
daibhin, line 1: Might be worth adding a migration guide somewhere telling people the main things needed to migrate from v4 / v5 to v6

CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+# 6.0.0
+
+This release contains a number of major breaking changes:
+- feat: make distinct_id an optional parameter in posthog.capture and related functions
+- feat: make capture and related functions return `Optional[str]`, which is the UUID of the sent event, if it was sent
+- fix: remove `identify` (prefer `posthog.set()`), and `page` and `screen` (prefer `posthog.capture()`)
```
daibhin, line 6: What's the thinking behind removing these methods?

- I get `identify` is pretty pointless because it doesn't actually persist across captures in the same way it does for the JS Web SDK. Can't you set the user id with contexts now? I'm guess that's the better way than setting the distcint_id on every capture
- Worth keeping the `page` method so that people don't have to familiarize themselves with the `$pageview` property? Screen I'm less concerned with because it doesn't really make sense on the web and is only used in the mobile SDKs afaik

posthog/exception_integrations/django.py
```diff
@@ -1,117 +0,0 @@
...
-        # Extract traceparent and tracestate headers
-        traceparent = headers.get("Traceparent")
-        tracestate = headers.get("Tracestate")
-
-        # Extract the distinct_id from tracestate
-        distinct_id = None
-        if tracestate:
-            # TODO: Align on the format of the distinct_id in tracestate
-            # We can't have comma or equals in header values here, so maybe we should base64 encode it?
-            match = re.search(r"posthog-distinct-id=([^,]+)", tracestate)
```
daibhin, line 78: Any idea if this was being set anywhere and whether we should include it in the middleware?

posthog/exception_integrations/django.py
```diff
@@ -1,117 +0,0 @@
...
-            if match:
-                distinct_id = match.group(1)
-
-        return {
-            **self.user(),
-            "distinct_id": distinct_id,
-            "ip": headers.get("X-Forwarded-For"),
-            "user_agent": headers.get("User-Agent"),
-            "traceparent": traceparent,
-            "$request_path": self.request.path,
```
daibhin, line 88: Same with ip, user agent and request path. Should we be adding them to the Django integration or are they there already?

posthog/exception_integrations/django.py
```diff
@@ -1,117 +0,0 @@
...
-                user_data.setdefault("$user_id", user_id)
-        except Exception:
-            pass
-
-        try:
-            email = str(user.email)
-            if email:
-                user_data.setdefault("email", email)
-        except Exception:
-            pass
```
daibhin, line 111: These are pretty reasonable heuristics for anyone using Django... should we include them in the other integration?

oliverb123: As above, this can be added to the django middleware if we want

daibhin: I went ahead and added the necessary stuff from the request extractor in this commit [`fca65f7` (#273)](https://github.com/PostHog/posthog-python/pull/273/commits/fca65f754056a59c10c1877e17416e2e04053b0b). Couldn't get the tests to run locally to write another one though :/

daibhin: Seeing this error when running `pytest posthog/test/integrations/test_middleware.py` if you've come across that Olly
```
from posthog.scopes import new_context, get_context_session_id, get_context_distinct_id
ModuleNotFoundError: No module named 'posthog'
```

oliverb123: I don't see that error, but you were forgetting to pass the request to the `self.extract_request_user` call on line 81. Fixed it now. Also used `getattr` on the `is_authenticated` property, because I don't trust that it always exists.

posthog/test/test_before_send.py
```diff
@@ -40,16 +40,28 @@ def my_before_send(event):
...
-        self.assertEqual(len(processed_events), 1)
-        self.assertEqual(processed_events[0]["event"], "test_event")
+        with mock.patch("posthog.client.batch_post") as mock_post:
+            client = Client(
+                FAKE_TEST_API_KEY,
+                on_error=self.set_fail,
+                before_send=my_before_send,
+                sync_mode=True,
+            )
+            msg_uuid = client.capture("user1", "test_event", {"original": "value"})
```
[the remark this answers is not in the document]
oliverb123: Change it to what? The client object itself already supported optional distinct IDs (it already being a keyword arg), so only the proxy functions needed to change.

posthog/__init__.py
```diff
@@ -54,27 +63,31 @@
 default_client = None  # type: Optional[Client]
 
 
+# NOTE - this and following functions take and unpack kwargs because we needed to make
+# it impossible to write `posthog.capture(distinct-id, event-name)` - basically, to enforce
+# the breaking change made between 5.3.0 and 6.0.0. This decision can be unrolled in later
+# versions, without a breaking change, to get back the type information in function signatures
 def capture(
```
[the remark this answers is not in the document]
oliverb123: I'd rather not, it feels like a lot of boilerplate tbh. My preference is to release this version, and then in subsequent versions go back to ordinary `def func(key=value)` syntax.

posthog/test/test_before_send.py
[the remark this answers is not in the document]
oliverb123: Switched to using `Unpack[TypedDict]` based kwargs everywhere it seemed to make sense

PR #275, opened 2025-06-27
feat(flags): add a native flag evaluation that tracks the status of feature flags and uses them as fallback values whenever the `/flags` API isn't available

dmarticus:
what it says on the tin.

Somewhere between a prototype and an actual thing that I can ship, maybe a bit closer to the latter 😉

### DIFF
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -50,6 +50,7 @@ from posthog.types import (
     to_values,
 )
 from posthog.utils import (
+    FlagCache,
     SizeLimitedDict,
     clean,
     guess_timezone,
@@ -126,6 +127,9 @@ class Client(object):
         project_root=None,
         privacy_mode=False,
         before_send=None,
+        enable_flag_cache=True,
+        flag_cache_size=10000,
+        flag_cache_ttl=300,
     ):
         self.queue = queue.Queue(max_queue_size)
 
@@ -151,6 +155,10 @@ class Client(object):
         )
         self.poller = None
         self.distinct_ids_feature_flags_reported = SizeLimitedDict(MAX_DICT_SIZE, set)
+        self.flag_cache = (
+            FlagCache(flag_cache_size, flag_cache_ttl) if enable_flag_cache else None
+        )
+        self.flag_definition_version = 0
         self.disabled = disabled
         self.disable_geoip = disable_geoip
         self.historical_migration = historical_migration
@@ -707,6 +715,9 @@ class Client(object):
 
     def _load_feature_flags(self):
         try:
+            # Store old flags to detect changes
+            old_flags_by_key = self.feature_flags_by_key or {}
+
             response = get(
                 self.personal_api_key,
                 f"/api/feature_flag/local_evaluation/?token={self.api_key}&send_cohorts",
@@ -717,6 +728,12 @@ class Client(object):
             self.feature_flags = response["flags"] or []
             self.group_type_mapping = response["group_type_mapping"] or {}
             self.cohorts = response["cohorts"] or {}
+
+            # Check if flag definitions changed and update version
+            if self.flag_cache and old_flags_by_key != self.feature_flags_by_key:
+                old_version = self.flag_definition_version
+                self.flag_definition_version += 1
+                self.flag_cache.invalidate_version(old_version)
 
         except APIError as e:
             if e.status == 401:
@@ -738,6 +755,10 @@ class Client(object):
                 self.feature_flags = []
                 self.group_type_mapping = {}
                 self.cohorts = {}
+
+                # Clear flag cache when quota limited
+                if self.flag_cache:
+                    self.flag_cache.clear()
 
                 if self.debug:
                     raise APIError(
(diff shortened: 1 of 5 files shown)


dmarticus:
going to revisit this PR and incorporate some suggestion from this thread: https://posthog.slack.com/archives/C07Q2U4BH4L/p1751403259100799

Will leave it open for now but will likely close it soon.

posthog/client.py
```diff
@@ -95,7 +97,33 @@ def add_context_tags(properties):
...
+        With feature flag fallback cache (memory):
+        >>> client = Client("your-api-key", enable_flag_fallback_cache=True)
+
+        With Redis fallback cache for high-scale applications:
+        >>> client = Client(
+        ...     "your-api-key",
+        ...     enable_flag_fallback_cache=True,
+        ...     flag_fallback_cache_backend="redis",
+        ...     flag_fallback_cache_redis_url="redis://localhost:6379/0"
+        ... )
```
rafaeelaudibert, line 115: I'd propose a slightly different API, similar to how we do configs on `posthog-js`

```python
# In-memory
client = Client("api-key", flag_fallback_cache=True)

# Redis URL
client = Client("api-key", flag_fallback_cache=FlagFallbackCache(backend="redis", url="redis://localhost:6379/0")

# Existing redis
client = Client("api-key", flag_fallback_cache=FlagFallbackCache(backend="redis", client=redis_client)
```

This avoids us flooding the main `args` list and offloads all of the config-handling to a separate class. That class can also handle instantiating the cache, for example. I believe the code would look much more modular that way.

posthog/utils.py
```diff
@@ -1,6 +1,9 @@
+import json
 import logging
 import numbers
+import pickle
```
oliverb123, line 4: I think we need to use json for this - keep in mind more than just python SDKs could be talking to the same cache _at the same time_ here

posthog/client.py
```diff
@@ -126,6 +154,12 @@ def __init__(
         project_root=None,
         privacy_mode=False,
         before_send=None,
+        enable_flag_fallback_cache=False,
+        flag_fallback_cache_size=10000,
```
oliverb123, line 158: I can't tell what the unit is here, and don't think a user will be able to easily either? See note below, seems backend-specific so I'd put it in a url param.

posthog/client.py
```diff
@@ -126,6 +154,12 @@ def __init__(
...
+        flag_fallback_cache_ttl=300,
+        flag_fallback_cache_backend="memory",
```
oliverb123, line 160: As above - you'll need to accept URLs for this anyway, and URLs are fundamentally a serialisation format to describe an access protocol and a location. I'd put these as query parameters into the URL, where they're relevant, like:
- `memory://local/?ttl=300&size=1000`
- `redis://username:password@host:port/?some_param=some_value`.

This has the advantage of being universal across languages, so we can expose it in a consistent manner (avoiding interface skew a bit), and letting you take vastly different arguments depending on the backend without an endlessly growing list of parameters or wrapper/enum types exposed

posthog/client.py
oliverb123: I think we want to specifically prevent people passing clients here, though - for reasons [outlined here](https://posthog.slack.com/archives/C07Q2U4BH4L/p1751412934674349?thread_ts=1751403259.100799&cid=C07Q2U4BH4L) - letting them pass a custom client is the same as defining an interface, except now we'd need to define one for every cache service rather than an abstraction.

I also can't find places we do this custom wrapper stuff in posthog-js?

posthog/client.py
```diff
@@ -95,7 +97,30 @@ def add_context_tags(properties):
...
+        ...     "your-api-key",
+        ...     flag_fallback_cache_url="redis://localhost:6379/0/?ttl=300"
+        ... )
+
+        With Redis authentication:
+        >>> client = Client(
+        ...     "your-api-key",
+        ...     flag_fallback_cache_url="redis://username:password@localhost:6379/0/?ttl=300"
+        ... )
+    """
```
dmarticus, line 123: is this overkill to include in the docstring?  makes it seem like this is the only thing worth passing to the Client, but I didn't know where else to put the docs.

posthog/client.py
```diff
@@ -1278,6 +1345,99 @@ def _get_all_flags_and_payloads_locally(
...
+                except ImportError:
+                    self.log.warning(
+                        "[FEATURE FLAGS] Redis not available, flag caching disabled"
+                    )
+                    return None
+                except Exception as e:
+                    self.log.warning(
+                        f"[FEATURE FLAGS] Redis connection failed: {e}, flag caching disabled"
+                    )
+                    return None
```
dmarticus, line 1429: originally i was falling back to `"memory"` if they couldn't connect to redis, but that was a bad idea because I don't want to silently create an in-memory cache if they can't initialize a redis connection.  Better to just warn and move on.

oliverb123 (approved):
Nice, this feels like a good interface to me

PR #279, opened 2025-07-07
feat: add minimum fastapi exception capture repro case

oliverb123:


rafaeelaudibert:

PR #280, opened 2025-07-08
Init reference doc generation

gewenyu99:
👋 Helloooo, it's time for some basic reference doc generation for Python.

For now, we document:

- client class and all methods
- module level methods
- types where possible

## Generation script

The specs will be compiled when you run `uv run python docs/generate_json_schemas.py`

## Automatic generation

This is out of scope for now. Instead, we'll be manually adding this and pinning versions to the Posthog.com repo for a while before we worry about automating anything.

I want to wait until :

- we generate a few of these and feed reasonably confident 
- decide if these specs are going to be versioned or we always use latest version
- replicate this per our Q3 goal for a few more SDKs (we might decide to change the schema/etc.)

Basically I think automating this early is going to be painful for now


posthog/__init__.py
```diff
@@ -417,27 +584,85 @@ def get_all_flags_and_payloads(
...
+    Category:
+        Client management
+    """
     _proxy("flush")
 
 
 def join():
-    """Block program until the client clears the queue"""
+    """
+    Block program until the client clears the queue.
```
rafaeelaudibert, line 639: We should probably add something here saying that you should most likely be using `shutdown` rather than using `join` directly, and this is a function you'll use in exceptional situations only

posthog/client.py
```diff
@@ -1003,10 +1300,7 @@ def get_feature_flag_result(
         disable_geoip=None,
     ) -> Optional[FeatureFlagResult]:
         """
-        Get a FeatureFlagResult object which contains the flag result and payload for a key by evaluating locally or remotely
-        depending on whether local evaluation is enabled and the flag can be locally evaluated.
-
-        This also captures the $feature_flag_called event unless send_feature_flag_events is False.
+        Get a FeatureFlagResult object
```
rafaeelaudibert, line 1303: This one isnt following your proposed structure

uv.lock
rafaeelaudibert, line 1: Why was this changed? This doesnt look good. Github isnt rendering this for me, but this *should* have been kept the same

docs/doc_constant.py
```diff
@@ -0,0 +1,80 @@
...
+OUTPUT_CONFIG = {
+    "filename": "posthog-python-references.json",
+    "indent": 2
+}
+
+# Documentation structure defaults
+DOC_DEFAULTS = {
+    "showDocs": True,
+    "releaseTag": "public",
+    "return_type_void": "void",
```
rafaeelaudibert, line 78: Should this be `None` like the bot is suggesting?

docs/doc_constant.py
rafaeelaudibert, line 1: We have an RFC that states we should use the `bin` folder for scripts. Can you add that there instead? I'd be happy with a `bin/docs` executable, and then maybe a `bin/docs` folder with your helper scripts. How does that sound?

I'm looking for the RFC and can link it to you if I find it

haacked: https://github.com/PostHog/meta/pull/309

gewenyu99: Yes ofc, let me go through suggestions. thanks as always for sitting through reviews of code like this, my eyes started watering near the end of this.

posthog/client.py
gewenyu99: Do you mean that this is different from other methods here or different form the JS doc comment structure.

If the latter,
This is a good point to discuss, this is closer to [google style python doc strings](https://google.github.io/styleguide/pyguide.html#doc-function-raises). I think a part of this was to keep this closer to:
1. what we already use (latest doc strings by olly follows closer to google stype doc strings)
2. what is native to this language.

I think it would look pretty wild to have ts/jsdoc in python code

gewenyu99: But I think you mean the former :P I think I missed a few methods 😆

uv.lock
gewenyu99: 🙃 I did something dumb with my venv, sorry reverting

posthog/__init__.py
gewenyu99: YES I missed join, only had it for flush. Thanks for the reminder

PR #283, opened 2025-07-10
feat(flags):  make the sendFeatureFlags parameter more declarative and ergonomic

dmarticus:
### Problem
The current `send_feature_flags` boolean parameter provides limited control over feature flag evaluation behavior. Users couldn't specify evaluation preferences or custom properties per capture event.

### Solution
Modified `send_feature_flags` to accept either:
- `bool` (existing behavior, fully backward compatible)
- `SendFeatureFlagsOptions` object with granular control:
  ```python
  {
      "only_evaluate_locally": bool,     # Force local vs remote evaluation
      "person_properties": dict,         # Custom properties for this event
      "group_properties": dict           # Custom group properties
  }
  ```

### Benefits
- **Declarative control**: Explicitly specify local vs remote evaluation per event
- **Custom properties**: Supply event-specific properties without affecting global state
- **Backward compatible**: Existing boolean usage unchanged
- **Ergonomic**: More flexible API that matches user expectations

### Usage Examples
```python
# Traditional usage (unchanged)
client.capture("event", send_feature_flags=True)

# New declarative usage
client.capture("event", send_feature_flags={
    "only_evaluate_locally": True,
    "person_properties": {"plan": "premium"}
})
```

This brings the Python SDK in line with the enhanced Node.js SDK functionality from https://github.com/PostHog/posthog-js-lite/pull/566

### DIFF
--- a/posthog/args.py
+++ b/posthog/args.py
@@ -5,6 +5,8 @@ from datetime import datetime
 import numbers
 from uuid import UUID
 
+from posthog.types import SendFeatureFlagsOptions
+
 ID_TYPES = Union[numbers.Number, str, UUID, int]
 
 
@@ -22,7 +24,8 @@ class OptionalCaptureArgs(TypedDict):
             error ID if you capture an exception).
         groups: Group identifiers to associate with this event (format: {group_type: group_key})
         send_feature_flags: Whether to include currently active feature flags in the event properties.
-            Defaults to False
+            Can be a boolean (True/False) or a SendFeatureFlagsOptions object for advanced configuration.
+            Defaults to False.
         disable_geoip: Whether to disable GeoIP lookup for this event. Defaults to False.
     """
 
@@ -32,8 +35,8 @@ class OptionalCaptureArgs(TypedDict):
     uuid: NotRequired[Optional[str]]
     groups: NotRequired[Optional[Dict[str, str]]]
     send_feature_flags: NotRequired[
-        Optional[bool]
-    ]  # Optional so we can tell if the user is intentionally overriding a client setting or not
+        Optional[Union[bool, SendFeatureFlagsOptions]]
+    ]  # Updated to support both boolean and options object
     disable_geoip: NotRequired[
         Optional[bool]
     ]  # As above, optional so we can tell if the user is intentionally overriding a client setting or not
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -397,11 +397,53 @@ class Client(object):
 
         extra_properties: dict[str, Any] = {}
         feature_variants: Optional[dict[str, Union[bool, str]]] = {}
-        if send_feature_flags:
+
+        # Parse send_feature_flags parameter
+        should_send_flags = False
+        only_evaluate_locally = None
+        flag_person_properties = None
+        flag_group_properties = None
+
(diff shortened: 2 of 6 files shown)


posthog/client.py
```diff
@@ -397,11 +397,53 @@ def capture(
...
+                    feature_variants = self.get_all_flags(
+                        distinct_id,
+                        groups=(groups or {}),
+                        person_properties=flag_person_properties,
+                        group_properties=flag_group_properties,
+                        disable_geoip=disable_geoip,
+                        only_evaluate_locally=True,
+                    )
+                elif only_evaluate_locally is False:
+                    # Force remote evaluation via /decide
```
haacked, line 430:
```suggestion
                    # Force remote evaluation via /flags
```

posthog/client.py
```diff
@@ -397,11 +397,53 @@ def capture(
...
+                    )
+                else:
+                    # Default behavior - use remote evaluation
+                    feature_variants = self.get_feature_variants(
+                        distinct_id,
+                        groups,
+                        person_properties=flag_person_properties,
+                        group_properties=flag_group_properties,
+                        disable_geoip=disable_geoip,
+                    )
```
haacked, line 446: These two branches of logic do the same thing, so we can combine them, unless I'm missing something.

```suggestion
                else:
                    # only_evaluate_locally is False or None so use remote evaluation (Default behavior)
                    feature_variants = self.get_feature_variants(
                        distinct_id,
                        groups,
                        person_properties=flag_person_properties,
                        group_properties=flag_group_properties,
                        disable_geoip=disable_geoip,
                    )
```

posthog/types.py
```diff
@@ -9,6 +9,24 @@
 BeforeSendCallback = Callable[[dict[str, Any]], Optional[dict[str, Any]]]
 
 
+class SendFeatureFlagsOptions(TypedDict, total=False):
+    """Options for sending feature flags with capture events.
+
+    Args:
+        only_evaluate_locally: Whether to only use local evaluation for feature flags.
+            If True, only flags that can be evaluated locally will be included.
+            If False, remote evaluation via /decide API will be used when needed.
```
haacked, line 18:
```suggestion
            If False, remote evaluation via /flags API will be used when needed.
```

haacked (approved):
Nice! Some stylistic suggestions and comment changes, but nothing blocking.

posthog/args.py
```diff
@@ -32,8 +35,8 @@ class OptionalCaptureArgs(TypedDict):
     uuid: NotRequired[Optional[str]]
     groups: NotRequired[Optional[Dict[str, str]]]
     send_feature_flags: NotRequired[
-        Optional[bool]
-    ]  # Optional so we can tell if the user is intentionally overriding a client setting or not
+        Optional[Union[bool, SendFeatureFlagsOptions]]
+    ]  # Updated to support both boolean and options object
     disable_geoip: NotRequired[
         Optional[bool]
     ]  # As above, optional so we can tell if the user is intentionally overriding a client setting or not
```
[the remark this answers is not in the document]
dmarticus: no the "as above" is in the config here: https://github.com/PostHog/posthog-python/blob/master/posthog/args.py#L26

Issue #226, opened 2025-04-22
bug: payloads are not properly deserialized


haacked:
The `posthog-python` library doesn't handle payloads correctly. It makes the wrong assumption that payloads are JSON fields of the `/decide` response. But in actuality, payloads are always JSON encoded strings.

You can see this faulty assumption embedded in a bunch of tests:

```python
"featureFlagPayloads": {"beta-feature": 100, "beta-feature2": 300},
```

When it should be:

```python
"featureFlagPayloads": {"beta-feature": "100", "beta-feature2": "300"},
```

In all our other client SDKs, we call the equivalent of `json.loads` when returning a payload. Unfortunately, changing `get_feature_flag_payload` to do this would be a breaking change, but possibly worth it?

I plan to add a new method `get_feature_flag_result` which an object that includes `enabled`, `variant`, `payload`. I can call `json.loads` there for sure since it's a new method. But the open question, do we go ahead and break `get_feature_flag_payload`?

haacked:
Ok, this issue was already fixed.

PR #284, opened 2025-07-10
feat: Expose `get_feature_flag_result` method in public API

haacked:
This exposes the existing `get_feature_flag_result` method that returns a `FeatureFlagResult` object containing `enabled`, `variant`, and `payload` properties. The payload is automatically deserialized from JSON.

Related to #226 where I suggested I would add this method. Here I go.

### DIFF
--- a/example.py
+++ b/example.py
@@ -59,6 +59,14 @@ print(posthog.feature_enabled("beta-feature", "distinct_id"))
 # get payload
 print(posthog.get_feature_flag_payload("beta-feature", "distinct_id"))
 print(posthog.get_all_flags_and_payloads("distinct_id"))
+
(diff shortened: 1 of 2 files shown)


dmarticus (commented):
this seems reasonable; but didn't we already ship `get_feature_flag_result` in `posthog-python` 4.0.0?  How is this significantly different?

haacked:
> @haacked is this still relevant? Don't we already expose this method?

It is still revelant. This PR exports `get_feature_flag_result publicly` by adding the function to `__init__.py` so users can call `posthog.get_feature_flag_result()` directly.

dmarticus (approved):

Issue #286, opened 2025-07-16
Django middleware doesn't capture exceptions


jamesbrewerdev:
# Bug Report

## Summary

`posthog.integrations.django.PosthogContextMiddleware` doesn't capture exceptions raised by Django views or middlewares. When a Django view or middleware raises an exception, Django converts the exception into a response instead of propagating the exception throw the middleware stack. See the comment in [convert_exception_to_response](https://github.com/django/django/blob/40d6eb2c0104f2119d5260b77402737cebf3f85c/django/core/handlers/exception.py#L24).

Django still makes the exception available to each middleware, but it does so using the `Middleware.process_exception` function, not `Middleware.__call__`. See [_get_response](https://github.com/django/django/blob/40d6eb2c0104f2119d5260b77402737cebf3f85c/django/core/handlers/base.py#L174) and [_get_response_async](https://github.com/django/django/blob/40d6eb2c0104f2119d5260b77402737cebf3f85c/django/core/handlers/base.py#L228).

This isn't caught by the tests in [test_middleware.py](https://github.com/PostHog/posthog-python/blob/6a271026d154a477c10dcf32c214abd92ae03b74/posthog/test/integrations/test_middleware.py) because the tests use `MockRequest` which bypasses all of Django code that handles exceptions.

## Solution

Add a `process_exception` function to `posthos.integrations.django.PosthogContextMiddleware`:

```python
class PosthogContextMiddleware:
    # ...

    def process_exception(self, request, exception):
        if self.request_filter and not self.request_filter(request):
            return

        with contexts.new_context(self.capture_exceptions, client=self.client):
            for k, v in self.extract_tags(request).items():
                contexts.tag(k, v)

            if self.capture_exceptions:
                if self.client:
                    self.client.capture_exception(exception)
                else:
                    contexts.capture_exception(exception)
```

I copied the implementation from the `__call__` function and changed it to handle the exception directly instead of relying on the error handling code in `contexts.new_context`. It's not clear to me what the downstream effects of re-raising the exception are, so I figure it's better not to.

## Environment

**Django** 5.2.4
**PostHog Python SDK** 6.1.0
**Python** 3.13.5

## Steps to Reproduce

```bash
$ mkdir posthog-django
$ cd posthog-django
$ uv init --name posthog-django --python 3.13.5 .
$ rm main.py
$ uv add django posthog
$ uv run django-admin startproject django_posthog .
```

[...]

PR #287, opened 2025-07-16
fix: capture django processed exceptions

daibhin:
Originally reported by @jamesbrewerdev in https://github.com/PostHog/posthog-python/issues/286

Django has its own exception handler for views and middleware that catches and processes exceptions. We need to hook into that and capture the exception manually. There might be some followup work needed here to process stack traces correctly but given this is pretty critical I'll follow up with that if necessary

Verified by manually editing the package (hence the old version number) that the view exception is correctly captured (including the context tags for URL etc)
Screenshot 2025-07-16 at 20 46 40

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 6.1.1 - 2025-07-16
+
+- fix: correctly capture exceptions processed by Django from views or middleware
+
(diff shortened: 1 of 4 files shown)


jamesbrewerdev:
I really appreciate the incredibly fast turnaround on this 😲 ❤️

PR #288, opened 2025-07-18
feat(flags): Implement local evaluation for flag dependencies

haacked:
Feature flags will [soon be able to have conditions based on how other flags are evaluated](https://github.com/PostHog/posthog/pull/35284). This PR adds support for that in local evaluation.

I'm going to create it as a DRAFT because there's some open questions about the flag dependency format in the referenced PR that need to be resolved before we back this in.


posthog/client.py
```diff
@@ -970,13 +994,14 @@ def join(self):
...
             ```
         """
-        for consumer in self.consumers:
-            consumer.pause()
-            try:
-                consumer.join()
-            except RuntimeError:
-                # consumer thread has not started
-                pass
+        if self.consumers:
```
haacked, line 997: Not part of this PR, but makes mypy happy.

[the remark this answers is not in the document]
haacked: The comment is not legit. The check `if self.consumers:` is actually:

  1. Necessary - to prevent a TypeError when self.consumers = None (sync mode)
  2. Harmless - when self.consumers = [] (empty list), both approaches do nothing
  3. Correct - when self.consumers has items, both approaches iterate over them

posthog/dependency_graph.py
```diff
@@ -0,0 +1,535 @@
+import logging
```
andyzzhao, line 1: My main concern is adding dependency resolution logic to the SDKs will make them more complex and harder to maintain. 

Is it possible to move the dependency resolution logic to the server? That way, the local evaluation endpoint returns flags that are already resolved and can be evaluated

haacked: Oh interesting! If we return the flags in evaluation order, then the SDK logic gets way simpler. Great idea!

We can also transform the filters when sending them to the client so IDs are already resolved to the flag keys.

PR #291, opened 2025-07-21
feat(llmo): Make it optional to pass `posthog` client

Twixes:
It's been inconvenient to have to explicitly pass a `posthog_client` to every AI provider wrapper. We should do what happens when you call `posthog` methods like `posthog.capture()` – run `setup()` to init the `default_client` and use `default_client`.

This is what this PR introduces.


k11kirky (approved):
Looks good. Added to gemini as well, Realized i need to fix the docs as well and support this for node.

PR #299, opened 2025-08-01
fix(llmo): tool calls are broken for most providers

Radu-Raicea:
### Problem 1

The `$ai_tools` represents the available tools given to the LLM, not a tool call in the response. Some providers did not send that property at all, others set it to the tool calls instead of the available tools.

This properly sets `$ai_tools` for all providers in the Python SDK.

### Problem 2

`$ai_output_choices` doesn't include the tool calls for all the providers. Sometimes, we can also receive multiple output messages from the LLM, like for Anthropic, where there can be both a text output and a tool call.

This sends a an array for `content` inside `$ai_output_choices`, with all the text, tool call, and image messages.

Next step is to reproduce both changes in the Node SDK.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # 6.3.4 - 2025-08-04
 
-- fix: Set `$ai_tools` properly for all providers
+- fix: Set `$ai_tools` for all providers and `$ai_output_choices` for all non-streaming provider flows properly
 
 # 6.3.3 - 2025-08-01
 


Radu-Raicea:
@greptileai

k11kirky (approved):
Makes sense - The only thing different from what we discussed was having the tool_calls as a property on the message with role assistant rather than within the content property. 

I do not have strong feelings on this, so happy to move ahead as long as we are consistant.

k11kirky:
This is anthropics response:

```
{
  "role": "assistant",
  "content": [
    {
      "type": "text",
      "text": "<thinking>To answer this question, I will: 1. Use the get_weather tool to get the current weather in San Francisco. 2. Use the get_time tool to get the current time in the America/Los_Angeles timezone, which covers San Francisco, CA.</thinking>"
    },
    {
      "type": "tool_use",
      "id": "toolu_01A09q90qw90lq917835lq9",
      "name": "get_weather",
      "input": {"location": "San Francisco, CA"}
    }
  ]
}
```

PR #309, opened 2025-08-18
consumer.py: mark task_done for discarded items

### LINKED ISSUE #308: Flush never completes if too large of item is submitted
```python
import os
from posthog import Posthog

POSTHOG_API_KEY = os.environ["POSTHOG_API_KEY"]

posthog = Posthog(POSTHOG_API_KEY, host="https://us.i.posthog.com")
posthog.capture_exception(
    Exception("Test Exception"),
    properties={
        "user_id": "123",
        "context": "test"
    }
)
posthog.flush()
print("Flushed first exception")

posthog.capture_exception(
    Exception("1234567890" * 90000),
    properties={
        "user_id": "123",
        "context": "test"
    }
)
posthog.flush()                     # <----- hangs here
print("Flushed second exception")
```

The consumer code needs to mark `self.queue.task_done()` when discarding items that are too large.

masenf:
Fixes #308

### DIFF
--- a/posthog/consumer.py
+++ b/posthog/consumer.py
@@ -110,6 +110,7 @@ class Consumer(Thread):
                     self.log.error(
                         "Item exceeds 900kib limit, dropping. (%s)", str(item)
                     )
+                    queue.task_done()
                     continue
                 items.append(item)
                 total_size += item_size


masenf:
we're no longer using the python lib and sending updates directly via the http api

PR #318, opened 2025-08-29
feat(llma): redact base64 images

### LINKED ISSUE #305: [LLMO] Redact inline base64 images from input
We need to replicate these changes in the Python SDK for LLMO: https://github.com/PostHog/posthog-js/pull/2217

The original issue: https://github.com/PostHog/posthog-js/issues/2134

Allowing inline base64 encoded images to be sent as part of the PostHog event is not safe due to the maximum size of the event.

A better approach is to first upload the image via some sort of Files API, and then use its URL. For those that still want to use inline base64 encoded images, they need to be redacted from the PostHog event.

carlos-marchal-ph:
Removes base64 images from LLM traces to prevent them from exceeding the maximum event size. Replicates [the logic](https://github.com/PostHog/posthog-js/pull/2217) recently introduced into the JS SDK. Resolves #305.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 6.7.1 - 2025-09-01
+
+- fix: Add base64 inline image sanitization
+
 # 6.7.0 - 2025-08-26
 
 - feat: Add support for feature flag dependencies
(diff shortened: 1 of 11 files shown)


Radu-Raicea:
I think you're missing some sanitization:
- https://github.com/PostHog/posthog-python/blob/8da6920dd5260b8fd1747369395ada2204b8c07d/posthog/ai/openai/openai.py#L433
- https://github.com/PostHog/posthog-python/blob/8da6920dd5260b8fd1747369395ada2204b8c07d/posthog/ai/openai/openai_async.py#L437

PR #320, opened 2025-09-02
feat/add support for litellm

carlos-marchal-ph:
Implements support for LiteLLM for chat completions and embeddings.


carlos-marchal-ph:
We went for submitting a PR to them instead of building a wrapper in the end.

PR #321, opened 2025-09-02
fix(flags): we don't fallback to `/decide` any more

dmarticus:
it's `/flags` now actually


haacked (approved):
I'm glad you `decide`d to `flag` this for a PR.

PR #326, opened 2025-09-05
fix/system prompt sometimes missing

carlos-marchal-ph:
Fixes missing system prompt logging for the following cases:
- Gemini both streaming and non streaming
- OpenAI responses API when sending the prompt as `instructions`

### DIFF
--- a/posthog/ai/gemini/gemini_converter.py
+++ b/posthog/ai/gemini/gemini_converter.py
@@ -261,7 +261,9 @@ def extract_gemini_tools(kwargs: Dict[str, Any]) -> Optional[Any]:
     return None
 
 
-def format_gemini_input_with_system(contents: Any, config: Any = None) -> List[FormattedMessage]:
+def format_gemini_input_with_system(
+    contents: Any, config: Any = None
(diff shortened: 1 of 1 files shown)


posthog/ai/utils.py
```diff
@@ -172,7 +172,31 @@ def merge_system_prompt(kwargs: Dict[str, Any], provider: str):
         from posthog.ai.gemini.gemini_converter import format_gemini_input
 
         contents = kwargs.get("contents", [])
-        return format_gemini_input(contents)
+        formatted_messages = format_gemini_input(contents)
+        
+        # Check if system instruction is provided in config parameter
```
Radu-Raicea, line 177: If this is provider specific formatting, it should live in that provider's converter file.

Radu-Raicea (approved):
Other than that little comment, it looks good. We only had those two providers that didn't work? Did you try the streaming versions too?

carlos-marchal-ph:
Yes, I tried all of them before and after the changes, these two were the only ones broken for both streaming and non-streaming.

PR #327, opened 2025-09-10
fix: Prevent core Client methods from raising exceptions

dustinbyrne:
The goal is to ensure that our client doesn't cause a panic in an end-user application. This change updates
capture/set/set_once/group_identify/alias to swallow and log any exceptions that occur. Note that this won't prevent errors from propagating via the `on_error` callback if an error occurs while processing the queue.

Other methods like feature flags are documented to raise exceptions:
https://posthog.com/docs/libraries/python#error-handling

Resolves [#36982](https://github.com/PostHog/posthog/issues/36982)

### DIFF
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -3,7 +3,7 @@ import logging
 import os
 import sys
 from datetime import datetime, timedelta
-from typing import Any, Dict, Optional, Union
+from typing import Any, Callable, Dict, Optional, Union
 from typing_extensions import Unpack
 from uuid import uuid4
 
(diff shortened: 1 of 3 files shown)


posthog/client.py
```diff
@@ -99,6 +99,31 @@ def add_context_tags(properties):
...
+
+    def decorator(func):
+        from functools import wraps
+
+        @wraps(func)
+        def wrapper(self, *args, **kwargs):
+            try:
+                return func(self, *args, **kwargs)
+            except Exception as e:
+                self.log.exception(f"Error in {func.__name__}: {e}")
```
haacked, line 119: Should we check `self.debug` and re-raising in debug mode?

haacked (approved):
I really like the use of a decorator here! Nice job!

Because this is a change of behavior, it makes sense to update the docstring to note this behavior for each of the annotated methods with something like:

```
"""
...
Returns:
    Optional[str]: Event UUID if successful, None if error occurs.

Note: This method will not raise exceptions. Errors are logged.
"""
```

But that's not a blocker.

:shipit:

PR #332, opened 2025-09-23
Add workflow to create and save versioned references 

gewenyu99:
We want automatically updated, versioned, and persisted references for all SDKs.

See [related RFC](https://github.com/PostHog/meta/pull/368/files#diff-c5df39f9badac88d0947f3fc589c0061b7ef39cef8b84506feb327cfc3897d36R28)

This PR does these things:
- Updates script to produce a latest and versioned copy of the references
- The workflow will run on push to master
- The references and only changes in the references folder will be committed to master
- We will pull these in on build on the website later.

See [this run on my fork](https://github.com/gewenyu99/posthog-python/actions/runs/17955696272/job/51066265213) for the workflow

Feel free to bring up suggestions/objections to this approach, but let's keep the discussion in the  [related RFC](https://github.com/PostHog/meta/pull/368/files#diff-c5df39f9badac88d0947f3fc589c0061b7ef39cef8b84506feb327cfc3897d36R28) for anything that's not repo specific <3 

Looking forward to y'all's wisdom and help!

### DIFF
--- a/.github/workflows/generate-references.yaml
+++ b/.github/workflows/generate-references.yaml
@@ -24,7 +24,7 @@ jobs:
           python-version: 3.11.11
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v6
+        uses: astral-sh/setup-uv@0c5e2b8115b80b4c7c5ddf6ffdd634974642d182 # v5.4.1
         with:
             enable-cache: true
             pyproject-file: 'pyproject.toml'


.github/workflows/generate-references.yaml
```diff
@@ -0,0 +1,53 @@
...
+          fetch-depth: 0
+          token: ${{ secrets.POSTHOG_BOT_GITHUB_TOKEN }}
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: 3.11.11
+
+      - name: Install uv
+        uses: astral-sh/setup-uv@0c5e2b8115b80b4c7c5ddf6ffdd634974642d182 # v5.4.1
```
gewenyu99, line 28:
```suggestion
        uses: astral-sh/setup-uv@v6
```

.github/workflows/generate-references.yaml
rafaeelaudibert, line 1: We don't have this as a requirement in our SDKs, but it'd be better if you set the actions to a specific commit hash rather than to a version like you're doing here. GitHub's ones are fine, but the `astral` and `stefanzweifel` ones could be easily compromised/abused. You can look at our main repo to see how that works.

Just a nice to have

.github/workflows/generate-references.yaml
```diff
@@ -0,0 +1,53 @@
+name: "Generate References"
+
+on:
+  push:
+    branches:
+      - master
+  workflow_dispatch:
```
rafaeelaudibert, line 7: Not all pushes to master release a new version. Can we add something here to restrict this from running at all times?

Or, maybe something to detect we've already generated docs for that version?

My fear is the following:
- Release V1.0.0
- This workflow generates spec for v1.0.0 and adds in the commit
- Merges a new feature on master but does NOT release it as a new feature
- This workflow generates spec for v1.0.0 and adds in the commit ⚠️⚠️⚠️⚠️⚠️ WRONG

We'd have overwritten the docs for v1.0.0 incorrectly

rafaeelaudibert (changes requested):
There's a tiny bug here since the workflow is running more often than needed

.github/workflows/generate-references.yaml
rafaeelaudibert: Hmm, I don't think so. We could update something called `edge` or `master`, but `latest` should probably always be the "latest that was actually released".

.github/workflows/generate-references.yaml
```diff
@@ -0,0 +1,52 @@
+name: "Generate References"
+
+on:
+  release:
+    types: [published]
```
rafaeelaudibert, line 5: I've double checked and we do publish releases here in Github, so lgtm!

rafaeelaudibert (approved):

PR #335, opened 2025-10-03
fix: preserve tool_calls in OpenAI input messages

andrewm4894:
- Add helper function to format tool calls from both API responses and input messages
- Handle multimodal content and tool calls in structured format
- Maintain backwards compatibility for simple string content
- Add test coverage for messages with tool calls

### DIFF
--- a/posthog/ai/openai/openai_converter.py
+++ b/posthog/ai/openai/openai_converter.py
@@ -18,6 +18,60 @@ from posthog.ai.types import (
 )
 
 
+def _format_tool_calls_to_content(tool_calls: List[Any]) -> List[FormattedFunctionCall]:
+    """
+    Helper function to format tool calls into standardized content items.
+
+    Handles both object-based tool calls (from API responses) and dict-based
+    tool calls (from input messages).
+
+    Args:
+        tool_calls: List of tool call objects or dicts from OpenAI API
+
+    Returns:
+        List of formatted function call content items
+    """
+    formatted: List[FormattedFunctionCall] = []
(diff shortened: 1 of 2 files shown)


posthog/ai/openai/openai_converter.py
```diff
@@ -160,10 +206,45 @@ def format_openai_input(
...
+            )
+
+            # Simple string content (with optional tool_call_id for tool messages)
+            if isinstance(content_value, str) and not has_tool_calls:
+                formatted_msg: FormattedMessage = {
+                    "role": role,
+                    "content": content_value,
+                }
+                # Preserve tool_call_id for tool role messages
+                if tool_call_id:
```
Radu-Raicea, line 228: Can we have a `tool_call_id` if we don't have any `tool_calls`?

posthog/ai/openai/openai_converter.py
```diff
@@ -55,17 +109,9 @@ def format_openai_response(response: Any) -> List[FormattedMessage]:
...
-                                "type": "function",
-                                "id": tool_call.id,
-                                "function": {
-                                    "name": tool_call.function.name,
-                                    "arguments": tool_call.function.arguments,
-                                },
-                            }
-                        )
+                    content.extend(
+                        _format_tool_calls_to_content(choice.message.tool_calls)
```
Radu-Raicea, line 113: This fixes it for the chat completions response API format. What about the responses API format?

andrewm4894:
gonna close this and re-open later if needed

PR #336, opened 2025-10-08
fix: fixed uncontrolled consumers

joul87:
### Problem

The PostHog Python client had an issue with consumer management where the consumers list was being initialized inside the loop that creates consumer threads, causing uncontrolled consumer creation. This could lead to:

- More consumers being created than intended
- Potential resource leaks
- Inconsistent behavior when multiple threads are specified

### Solution

- Moved consumer list initialization: Moved `self.consumers = []` outside the thread creation loop to ensure the list is properly initialized before adding consumers
- Added comprehensive test coverage: Created a test case `test_number_of_consumers()` to verify that the correct number of consumers are created based on the specified thread count

### DIFF
--- a/posthog/test/test_client.py
+++ b/posthog/test/test_client.py
@@ -2472,6 +2472,6 @@ class TestClient(unittest.TestCase):
 
         client = Client(FAKE_TEST_API_KEY, send=False, thread=expected_threads)
         try:
-            assert len(client.consumers) == expected_threads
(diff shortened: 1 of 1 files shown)


rafaeelaudibert:

dustinbyrne:
Hi @joul87, this issue looks to have been fixed already: https://github.com/PostHog/posthog-python/commit/65785b892ecc132c9f9a81b1bbe7add472e6db54

PR #340, opened 2025-10-15
Run SDK generation after release job

gewenyu99:
Helloooo!

The trigger for `published` doesn't work because of how we release. Instead we can just run following `workflows: ["Release"]`.

Similar approach in https://github.com/PostHog/posthog-js/actions/workflows/generate-references.yml

### DIFF
--- a/.github/workflows/generate-references.yaml
+++ b/.github/workflows/generate-references.yaml
@@ -1,10 +1,6 @@
 name: "Generate References"
 
 on:
-  workflow_run:
-    workflows: ["Release"]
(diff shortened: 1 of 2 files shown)


rafaeelaudibert (commented):
@gewenyu99 Why run following release rather than calling it via the gh CLI? We should probably standardize here

gewenyu99:
Fair, felt simpler but I mean there's no practical difference. Consistency is good. Thanks

rafaeelaudibert (approved):

PR #341, opened 2025-10-15
fix: Check for references directory and generate 6.7.7 specs

gewenyu99:
Embarrassing mistake 😆 Forgot to check if the references folder exists.

Sorry folks :P


lricoy (approved):
Should those really be committed?

Approving to unblock. You will know what you need to do

gewenyu99:
Yeah honestly, I don't have to commit these and can run the action to generate these 😶‍🌫️

PR #347, opened 2025-10-24
feat(ai): Add `$ai_framework` property for framework integrations

andrewm4894:
## Summary

Adds a simple `$ai_framework` property to AI events, **only when an actual framework layer is used** (e.g., LangChain). This helps identify which abstraction layer is being used on top of the underlying AI provider.

## Changes

- Add `$ai_framework` property to AI events for framework integrations (LangChain)
- Direct provider calls (OpenAI, Anthropic, Gemini) do **not** include this property
- Property contains a simple string value (e.g., `"langchain"`)

## Example Events

**Framework call (LangChain):**
```json
{
  "$ai_framework": "langchain",
  "$ai_provider": "openai",
  "$ai_model": "gpt-4",
  ...
}
```

**Direct provider call (OpenAI):**
```json
{
  "$ai_provider": "openai",
  "$ai_model": "gpt-4",
  ...
}
```

## Rationale

This approach eliminates redundancy - we only add `$ai_framework` when there's an actual framework layer. For direct provider calls, the framework would just duplicate the provider name, which is wasteful.

## Testing

- All existing tests updated and passing (119 passed, 8 skipped)
- Added assertions for `$ai_framework` in LangChain tests
- Verified property is not present in direct provider tests

### DIFF
--- a/posthog/ai/langchain/callbacks.py
+++ b/posthog/ai/langchain/callbacks.py
@@ -36,7 +36,7 @@ from langchain_core.outputs import ChatGeneration, LLMResult
 from pydantic import BaseModel
 
 from posthog import setup
-from posthog.ai.utils import get_model_params, with_privacy_mode, get_ai_lib_metadata
+from posthog.ai.utils import get_model_params, with_privacy_mode
 from posthog.ai.sanitization import sanitize_langchain
 from posthog.client import Client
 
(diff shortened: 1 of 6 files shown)


Radu-Raicea (approved):
Great idea! 🎉

PR #359, opened 2025-10-28
feat(llma): send number of web searches

Radu-Raicea:
This now track web search tool calls across various providers, and sends the web search count as an event property (`$ai_web_search_count`) to be used in calculating the LLM costs, where applicable.


andrewm4894 (approved):
lgtm apart form merge conflict but approving so no delay once that sorted

PR #361, opened 2025-10-30
fix: avoid return from finally block to fix Python 3.14 SyntaxWarning

jodal:
This fixes a SyntaxWarning on Python 3.14.

```
❯ uvx --no-cache --python 3.14.0 --with posthog==6.7.11 python -c "import posthog"
Installed 11 packages in 5ms
.../lib/python3.14/site-packages/posthog/consumer.py:92: SyntaxWarning: 'return' in a 'finally' block
  return success
````

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# 7.5.1 - 2026-01-07
+
+fix: avoid return from finally block to fix Python 3.14 SyntaxWarning (#361)[https://github.com/PostHog/posthog-python/pull/361] - thanks @jodal
+
 # 7.5.0 - 2026-01-06
 
 feat: Capture Langchain, OpenAI and Anthropic errors as exceptions (if exception autocapture is enabled)
--- a/posthog/version.py
+++ b/posthog/version.py
@@ -1,4 +1,4 @@
-VERSION = "7.5.0"
+VERSION = "7.5.1"
 
 if __name__ == "__main__":
     print(VERSION, end="")  # noqa: T201

### CODE REVIEW

File: posthog/consumer.py [hunk 1]
```diff
@@ -89,7 +89,8 @@ def upload(self):
             # mark items as acknowledged from queue
             for item in batch:
                 self.queue.task_done()
-            return success
+
+        return success
```
pauldambra: i think this changes the behaviour of the library

my understanding is that before this we would have swallowed any exceptions and returned false

but now we would throw the exceptions

am i right @webjunkie as my go-to pythonista?

posthog/consumer.py
pauldambra: am i being silly @jodal 

```
➜  ~ uvx --no-cache --python 3.14.0 --with posthog==6.7.11 python -c "import posthog"
Installed 11 packages in 6ms
➜  ~ uvx --no-cache --python 3.14.0 --with posthog==7.4.3 python -c "import posthog"
Installed 11 packages in 6ms
➜  ~ 
```

i don't get any errors
but also...

we introduced python 3.14 support in release 7.4.3 so it wouldn't surprise me that 6.7.11 doesn't support it 😅

pauldambra (changes requested):
let's at least wait until https://github.com/PostHog/posthog-python/pull/403 is merged

i want to be sure we don't break one thing while fixing another

when #403 is merged can you update this PR

posthog/consumer.py
jodal:
```
❯ uvx --no-cache --python 3.14.0 --with posthog==6.7.11 python -c "import posthog"
Installed 11 packages in 3ms
/tmp/.tmpZTdTTe/archive-v0/eBzD3AO7UNjUNwl1CRB4B/lib/python3.14/site-packages/posthog/consumer.py:92: SyntaxWarning: 'return' in a 'finally' block
  return success
❯ uvx --no-cache --python 3.14.0 --with posthog==7.4.3 python -c "import posthog"
Installed 11 packages in 3ms
/tmp/.tmpnyFRJo/archive-v0/aXixIXK21gcCmMr5IQ6Yk/lib/python3.14/site-packages/posthog/consumer.py:92: SyntaxWarning: 'return' in a 'finally' block
  return success
```

pauldambra: i'm confused though... why don't i get a syntax warning when running the same commands?

pauldambra: is it this? ➜  ~ uvx --version
uv-tool-uvx 0.6.3 (a0b9f22a2 2025-02-24)

jodal:
I've rebased the PR and made it log and swallow any errors from the `on_error` handler to keep the old behavior where the `return` in the `finally` clause implicitly swallowed any errors in the `except` block.

pauldambra:
@Piccirello i can't merge here because

```
Merging is blocked
Code scanning is waiting for results from CodeQL for the commits db70702 or ae32dfc.
```

i'm guessing that's because it's external?

you-are-our-only-hope.meme

Piccirello:
CodeQL requires you to setup CodeQL Advanced in a repo in order to run it against external contributor PRs (super annoying). We've started setting this up in repos as it pops up, like in posthog, posthog-android, posthog-ios. https://github.com/PostHog/posthog-python/pull/405 should fix this.

pauldambra:
one more ping, i can't figure out how to update this to master myself, sorry!
if you can bring this up-to-date, codeql should run and we're good to rock and roll

PR #373, opened 2025-11-18
feat: add python 3.14 support

carlos-marchal-ph:
Adds explicit support and tests for the latest version of Python (3.14) released in October.


pauldambra (approved):
I _think_ maybe this should be a breaking change to v8 since we're dropping pydantic v1 support

but then if someone is using 3.14 they already can't be using pydantic v1, right?

in which case we're good

(thinking out loud to make sure i'm not being stupid :)))

PR #382, opened 2025-12-04
feat: mask values

ablaszkiewicz:
This adds masking based on values (not only keys)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,10 @@
+# 7.3.0 - 2025-12-05
+
+feat: improve code variables capture masking
+
 # 7.2.0 - 2025-12-01
 
 feat: add $feature_flag_evaluated_at properties to $feature_flag_called events
-
 
 # 7.1.0 - 2025-11-26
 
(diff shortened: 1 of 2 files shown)


posthog/exception_utils.py
```diff
@@ -941,7 +942,31 @@ def _pattern_matches(name, patterns):
     return False
 
 
-def _serialize_variable_value(value, limiter, max_length=1024):
+def _mask_sensitive_data(value, compiled_mask):
```
hpouillot: Can we add a big try catch around the `try_attach_code_variables_to_frames` ? Just to avoid missing an exception if something goes wrong in there

posthog/exception_utils.py
```diff
@@ -54,6 +54,7 @@
     r"(?i).*privatekey.*",
     r"(?i).*private_key.*",
     r"(?i).*token.*",
+    r"(?i).*aws_access_key_id.*",
```
hpouillot:
```suggestion
    r"(?i).*aws_access_key_id.*",
    r"(?i).*_pass",
    r"(?i)sk_.*",
    r"(?i).*jwt.*",
```

PR #384, opened 2025-12-07
Fix OpenAI input image sanitization

vikas5914:
# Summary

1. Extend `sanitize_openai_image` to redact base64 data when OpenAI’s Responses API sends input_image items where image_url is a string (as documented [here](https://platform.openai.com/docs/guides/images-vision?api-mode=responses#giving-a-model-images-as-input)).

2. Restore the regression test for that payload format so we cover both dict-style image_url objects and string URLs.
The previous check only handled { "type": "image_url", "image_url": { "url": ... } }. But OpenAI also allows:

```
{
    "type": "input_image",
    "image_url": f"data:image/jpeg;base64,{base64_image}",
}
```

Example from the official docs:

```python
from openai import OpenAI

client = OpenAI()

response = client.responses.create(
    model="gpt-4.1-mini",
    input=[{
        "role": "user",
        "content": [
            {"type": "input_text", "text": "what's in this image?"},
            {
                "type": "input_image",
               "image_url": f"data:image/jpeg;base64,{base64_image}",
            },
        ],
    }],
)

print(response.output_text)
```

Without this fix, base64 strings in that shape weren’t redacted.


vikas5914:
@ablaszkiewicz Sorry for tagging you but any chance to review this?

ablaszkiewicz:
lgtm ✅ Python pipeline is currently broken. We are fixing it

vikas5914:
@ablaszkiewicz hey , any update on this

ablaszkiewicz (approved):

PR #387, opened 2025-12-10
feat: Add FlagDefinitionCacheProvider interface

dustinbyrne:
Mirrors https://github.com/PostHog/posthog-js/pull/2594 / https://posthog.com/tutorials/node-external-cache

This addition allows users to implement their own custom external caching solutions for local evaluation / flag definitions. Implementations would allow for distributed caching of flags definitions, such that one logical service can poll from a single coordinated leader, while other individual instances rely on cached flag definitions while evaluating flags locally.

Example code provided featuring an implementation for Redis, complete with coordination to poll from a single instance.

### DIFF
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -1213,9 +1213,9 @@ class Client(object):
         self, data: FlagDefinitionCacheData, old_flags_by_key: Optional[dict] = None
     ) -> None:
         """Update internal flag state from cache data and invalidate evaluation cache if changed."""
-        self.feature_flags = data.get("flags") or []
-        self.group_type_mapping = data.get("group_type_mapping") or {}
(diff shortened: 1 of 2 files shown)


examples/redis_flag_cache.py
```diff
@@ -0,0 +1,144 @@
...
+
+    # Lua script: release lock only if we own it
+    _LUA_STOP_LEAD = """
+        if redis.call('GET', KEYS[1]) == ARGV[1] then
+            return redis.call('DEL', KEYS[1])
+        end
+        return 0
+    """
+
+    def __init__(self, redis: Redis, service_key: str):
```
dmarticus: Should be `Redis[str]` since it requires decode_responses=True.

PR #395, opened 2025-12-18
fix(llma): extract model from response for OpenAI stored prompts

andrewm4894:
## Problem

When using OpenAI stored prompts, the model is defined in the OpenAI dashboard rather than passed in the API request. The PostHog AI wrapper only extracts the model from request parameters (`kwargs.get("model")`), causing generations to show up without a model and preventing cost calculations.

Fixes https://github.com/PostHog/posthog/issues/42861

## Changes

- Updated `utils.py` to fallback to `response.model` when model is not in kwargs
- Updated streaming handlers in `openai.py` and `openai_async.py` to extract model from response chunks

### DIFF
--- a/posthog/test/ai/openai/test_openai.py
+++ b/posthog/test/ai/openai/test_openai.py
@@ -1944,6 +1944,52 @@ def test_non_streaming_responses_api_extracts_model_from_response(mock_client):
         assert props["$ai_model"] == "gpt-4o-mini-stored"
 
 
+def test_non_streaming_returns_none_when_no_model(mock_client):
+    """Test that non-streaming returns None (not 'unknown') when model is not available anywhere."""
+    # Create a response without model attribute using real OpenAI types
+    mock_response = ChatCompletion(
+        id="test",
+        model="",  # Will be removed below
+        object="chat.completion",
+        created=int(time.time()),
+        choices=[
+            Choice(
+                finish_reason="stop",
+                index=0,
+                message=ChatCompletionMessage(
+                    content="Test response",
+                    role="assistant",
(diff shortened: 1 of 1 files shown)


posthog/ai/openai/openai_async.py
```diff
@@ -422,13 +443,17 @@ async def _capture_streaming_event(
         output: Any,
         tool_calls: Optional[List[Dict[str, Any]]] = None,
         available_tool_calls: Optional[List[Dict[str, Any]]] = None,
+        model_from_response: Optional[str] = None,
     ):
         if posthog_trace_id is None:
             posthog_trace_id = str(uuid.uuid4())
 
+        # Use model from kwargs, fallback to model from response
+        model = kwargs.get("model") or model_from_response
```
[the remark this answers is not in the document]
andrewm4894: this is fine as is

posthog/ai/utils.py
```diff
@@ -285,7 +285,9 @@ def call_llm_and_track_usage(
 
         event_properties = {
             "$ai_provider": provider,
-            "$ai_model": kwargs.get("model"),
+            "$ai_model": kwargs.get("model")
+            or getattr(response, "model", None)
+            or "unknown",
```
Radu-Raicea: Not sure how I feel about putting "unknown" if we don't know the model. This would trigger a cost model match that won't be correct. I'd rather not put anything in the field.

andrewm4894: seems like its in streaming its unknown only so think i should actually just match what was there before - will rejig a little to be more minimal

PR #398, opened 2026-01-01
chore: Migrate releases to `sampo`

rafaeelaudibert:
We're slowly migrating everything to this! We have it working for Elixir and PHP now, let's do for Python! Rust incoming too.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "posthog"
-version = "7.4.2"
+version = "7.9.1"
 description = "Integrate PostHog into any python application."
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
rafaeelaudibert: Updated this to make sure `sampo` can properly add the entry to the right place

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "posthog"
-dynamic = ["version"]
+version = "7.4.2"
```
rafaeelaudibert: This was hard to manage by hand previously so we made it dynamic, but now that we have `sampo` we should list it here

PR #400, opened 2026-01-02
feat(audio): add tracking for audio transcriptions in OpenAI client

JasonLovesDoggo:
This adds support for tracking transcriptions from OpenAI. It does this via a new event `$ai_transcription` which follows the pattern of embeddings. I figure that audio -> text Is different enough from text-to-text to deserve its own event.

Confirmed it worked in my own testing. Feel free to impersonate and view https://us.posthog.com/project/254263/events/2e8ded5c-acd2-45b4-b10f-7a85a438ffaa/2026-01-02T15%3A02%3A00.007000-05%3A00


JasonLovesDoggo:
cc @andrewm4894

carlos-marchal-ph:
Hi Jason! Carlos from the PostHog LLM analytics team here. Appreciate the PR, that's certainly a sensible way to go about it. Unfortunately I'm leaning towards shelving this for the moment. There's some internal reasons for it:

- We are currently implementing multimodal support. The current architecture leans towards treating all input/output to LLMs as potentially multimodal, which also supports mixed content by default.
- We are still discussing how we'd like to handle potentially large payloads in AI events. Audio might not be the worst offender here, but in line with the above we want a solution that can ingest large blobs (hi-res images, potentially even videos at some point). This might impact how we ingest them from the SDK side, so we don't want to commit to anything right now.
- In line with the above, we are trying to centralise as much of the event post-processing as possible to our backend. This makes it easier to control for us, and also avoids having to duplicate features across the different SDKs.

In the meantime, you should be able to get this working without forking by writing a small helper that calls `client.audio.transcriptions.create()` on the standard OpenAI client and then uses `posthog.capture()` to send a custom event with the properties you care about (model, latency, etc.). We'll be sure to credit you if we end up using part/all of your code moving forward. And again, thanks a lot for contributing!

PR #418, opened 2026-01-31
fix: extract_gemini_web_search_count

zachary822:
`extract_gemini_web_search_count` errors if `response.candidates` is None


rafaeelaudibert:

zachary822:
@marandaneto was this bug ever fixed? I don't mind this pr getting closed, but there is a bug in that function.

marandaneto:
> @marandaneto was this bug ever fixed? I don't mind this pr getting closed, but there is a bug in that function.

https://github.com/PostHog/posthog-python/pull/599
sorry the pr was auto closed since we changed the default branch from master to main

PR #427, opened 2026-02-06
feat(prompts): add capture_errors option for error tracking

andrewm4894:
This is just an exploratory one. As a user i kinda might want posthog errors if we fail to handle prompt as expected. I dont _think_ anything too crazy going on here but worth discussing.

## Summary

- Adds `capture_errors` parameter to `Prompts.__init__()` (default `False`, fully backward compatible)
- When `True` and a PostHog client is provided, prompt fetch failures are reported via `client.capture_exception()` before the existing fallback logic runs
- If `capture_exception()` itself fails, it is silently swallowed and never affects prompt resolution

## Motivation

When using prompts with fallbacks, fetch failures are silently handled -- the caller gets a fallback string and the error is only logged at WARNING level. With `capture_errors=True`, these failures surface in PostHog error tracking, making it easy to spot API issues, auth problems, or missing prompts without configuring Python logging.

## Usage

```python
from posthog import Posthog
from posthog.ai.prompts import Prompts

posthog = Posthog('phc_xxx', host='https://us.posthog.com', personal_api_key='phx_xxx')
prompts = Prompts(posthog, capture_errors=True)

# Fetch failures now show up in PostHog error tracking
template = prompts.get('my-prompt', fallback='default prompt')
```

## Changes

- `posthog/ai/prompts.py` -- new `capture_errors` param, `_maybe_capture_error()` helper
- `posthog/test/ai/test_prompts.py` -- 7 new tests covering all branches

## Test plan

- [x] capture_exception called on fetch failure with fallback
- [x] capture_exception called on fetch failure with stale cache
- [x] capture_exception called when error is re-raised (no fallback/cache)
- [x] NOT called when capture_errors=False (default)
- [x] NOT called when no client provided (direct personal_api_key mode)
- [x] NOT called on successful fetch
- [x] capture_exception failure does not affect fallback resolution
- [x] All 30 existing tests pass unchanged

### DIFF
--- a/posthog/ai/prompts.py
+++ b/posthog/ai/prompts.py
@@ -228,9 +228,7 @@ class Prompts:
         try:
             self._client.capture_exception(error)
         except Exception:
-            log.debug(
-                "[PostHog Prompts] Failed to capture exception to error tracking"
-            )
+            log.debug("[PostHog Prompts] Failed to capture exception to error tracking")
 
     def _fetch_prompt_from_api(self, name: str) -> str:
         """
--- a/posthog/test/ai/test_prompts.py
+++ b/posthog/test/ai/test_prompts.py
@@ -610,9 +610,7 @@ class TestPromptsCaptureErrors(TestPrompts):
         posthog.capture_exception.assert_not_called()
 
     @patch("posthog.ai.prompts._get_session")
-    def test_capture_exception_failure_does_not_affect_fallback(
-        self, mock_get_session
-    ):
+    def test_capture_exception_failure_does_not_affect_fallback(self, mock_get_session):
         """If capture_exception itself throws, the fallback should still be returned."""
         mock_get = mock_get_session.return_value.get
         mock_get.side_effect = Exception("Network error")


andrewm4894:
> @andrewm4894 is this still relevant?

Yeah I think it kinda is worth supporting. Was going to sense check the sampo stuff on it and update to get it ready at some point

posthog/version.py
```diff
@@ -1 +1,4 @@
-VERSION = "7.11.2"
+VERSION = "7.8.4"
```
marandaneto, line 1: you dont need to do this

marandaneto: https://github.com/PostHog/posthog-python/blob/main/RELEASING.md

posthog/ai/prompts.py
```diff
@@ -113,204 +75,113 @@ def __init__(
...
 
         Args:
             posthog: PostHog client instance (optional if personal_api_key provided)
-            personal_api_key: Direct personal API key (optional if posthog provided)
-            project_api_key: Direct project API key (optional if posthog provided)
+            personal_api_key: Direct API key (optional if posthog provided)
             host: PostHog host (defaults to app endpoint)
             default_cache_ttl_seconds: Default cache TTL (defaults to 300)
+            capture_errors: If True and a PostHog client is provided, prompt fetch
+                failures are reported to PostHog error tracking via capture_exception().
```
marandaneto, line 91: so we are reporting our own errors?
what's the final user gonna do with this when those errors are in their error tracking tab?

andrewm4894: Good question — I think the framing is actually from the user's perspective, not ours. The errors surfaced here aren't "PostHog internals" — they're 404s (prompt typo / deleted), 403s (key perms), auth failures, and network errors. All actionable by the app developer. If my app's prompts were silently falling back for two weeks and PostHog didn't surface it, I'd be pretty upset. It's opt-in (`capture_errors=False` by default), runs *before* the fallback chain so visibility doesn't cost reliability, and swallows capture-side failures so it can't break the caller. WDYT?

marandaneto: mmm I see, then yeah for a few errors that do make sense eg 404, 403 etc but not everything
eg network errors, timeouts, etc, this would be just noisy, affecting the noise x ratio error tracking quality cus those errors specifically arent actionable
can we at least filter only to capture the errors that are actually actionable? (this will depend on the API contract - http errors)

andrewm4894: Hmm if PostHog is down and my latest prompts aren't getting to my app you are saying to not error track that in my PostHog as a user? Do errors have to be actionable? Is that some sort of thing that users are aware or when using error tracking or something?

I'm not super up on error tracking fully tbh but feel a bit uncomfortable making such decisions here on behalf of future users (unless this kinda just is the norm or way error tracking works on posthog? Eg "actionable error tracking" where we do sort of try curate on your behalf, that's a bit different then).

Opt in too here as a gate helps maybe. 

The noise problem then is something you manage in error tracking itself via the groupings and rules and AI no?

Or maybe there is some way to make the different error types configurable and use like a sane default more like what you say perhaps. 

Although if my prod app was not getting latest prompts and error tracking was not surfacing that and it turned out PostHog prompts was just down and we missed it for a bit say, and then I found this thread where we decided to just not surface it. I'd be very mad.

Or am I being a bit stubborn here? Like I think of an error as an error, but of course it's always more nuanced where you draw the line so maybe that's the point here perhaps of where that is.

marandaneto: I don't disagree with most of your comments, but if posthog is down, or your internet is down, there's not much you can do besides having a fallback on the client
also, capturing errors most likely wont go thru anyway since most likely that endpoint is also down, or you are shooting the server with more requests and causing even more problems
Error tracking should focus on actionable errors I can address on my end. 
I wouldn't want to pay for error events that the only solution is "tell PostHog to be more reliable or tell me to have a more stable internet", but I am very opinionated about this, I know
The way I see this is more about "SDK diagnostics", we failed to request the prompts, and that's an SDK issue and not a customer issue.
I'm not familiar with this product, so I'll approve it and leave it with you. I just wanted to share my 2 cents

marandaneto:
@andrewm4894 you have to address [this](https://github.com/PostHog/posthog-python/pull/427#discussion_r3092551461) comment and fix the merge conflicts before merging

posthog/ai/prompts.py
andrewm4894: Yeah I know what you mean, I'm kinda thinking of the potential code in here to cherry pick errors as an implicit suppression rule in PostHog error tracking. 

Going to research a bit more on how it all works and what other parts of the SDK already do for stuff like this just to make sure I'm at least in line with them and things like that. 

Eg does PostHog error tracking itself then maybe need some sort of new template for these new potential prompt errors etc. 

Will check all that before merging to be sure.

marandaneto: awesome, thx

andrewm4894:
Superseded by #520 — rebased onto current main (couldn't force-push this branch due to signed-commits ruleset flagging historical pre-sampo release commits). Addressed review feedback there: dropped the version bump (per @marandaneto), added sampo changeset, added hasattr guard on capture_exception (per Greptile). Review continues on #520.

PR #447, opened 2026-02-27
feat: add semver targeting support to local flag evaluation

dmarticus:
## Summary

- Add 9 semver comparison operators (`semver_eq`, `semver_neq`, `semver_gt`, `semver_gte`, `semver_lt`, `semver_lte`, `semver_tilde`, `semver_caret`, `semver_wildcard`) for local feature flag evaluation
- Implements the client-side evaluation to match PostHog/posthog#44596
- Parsing handles v-prefix, whitespace, pre-release suffixes, leading zeros, and 4-part versions

## Why not use a semver library?

Libraries like `packaging` and `semver` exist but aren't runtime dependencies of this SDK. The current runtime deps are intentionally minimal (`requests`, `six`, `python-dateutil`, `backoff`, `distro`, `typing-extensions`), and adding a dependency for ~20 lines of string splitting isn't worth the install footprint cost for SDK consumers. The Rust feature-flags service uses its ecosystem's `semver` crate, but Rust deps don't have the same user-facing cost as Python deps.

## Test plan

- [x] All 9 operators tested with positive/negative cases
- [x] Edge cases: v-prefix, whitespace, leading zeros, pre-release suffixes, partial versions, 4-part versions
- [x] Caret operator `^0.x.y` behavior per semver spec
- [x] Invalid input and missing property handling
- [x] All 115 tests pass (12 new + 103 existing)

### DIFF
--- a/.sampo/changesets/noble-duke-louhi.md
+++ b/.sampo/changesets/noble-duke-louhi.md
@@ -1,0 +1,5 @@
+---
+pypi/posthog: patch
+---
+
+add semver targeting support to local evaluation


posthog/feature_flags.py
```diff
@@ -505,6 +505,79 @@ def compare(lhs, rhs, operator):
...
+        "semver_eq",
+        "semver_neq",
+        "semver_gt",
+        "semver_gte",
+        "semver_lt",
+        "semver_lte",
+        "semver_tilde",
+        "semver_caret",
+        "semver_wildcard",
+    ):
```
dmarticus: I know this is ugly, but the inline tuple approach matches the rest of the repo. I have refactored this in https://github.com/PostHog/posthog-python/pull/448 fwiw.

gustavohstrassburger (approved):

PR #452, opened 2026-03-04
Add warning log for local flag evaluation cold start

dmarticus:
## Summary

- When `feature_enabled()` is called with `only_evaluate_locally=True` before feature flag definitions have been fetched, the SDK silently returns `None`. This adds a warning log so users can diagnose the issue immediately instead of debugging SDK internals.

## Context

Discovered when debugging a "cold start" issue in a Django shell on a production pod. The first call to `feature_enabled(..., only_evaluate_locally=True)` returned `None` because the initial HTTP fetch of flag definitions hadn't completed (or failed silently). The second call worked fine because the poller had since fetched the flags.

The existing error path in `_fetch_feature_flags_from_api` already logs when the fetch itself fails, but there was no log at the point where `None` is actually returned to the caller.

## Test plan

- [x] All 671 existing tests pass
- [x] Warning only fires when `self.feature_flags` is falsy (not loaded yet)

### DIFF
--- a/.sampo/changesets/steady-context-resolver.md
+++ b/.sampo/changesets/steady-context-resolver.md
@@ -1,0 +1,5 @@
+---
+pypi/posthog: patch
+---
+
+fix(llma): use distinct_id from outer context if not provided, fix $process_person_profile for context-based identity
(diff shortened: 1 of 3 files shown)


sakce (approved):
stamp away

PR #453, opened 2026-03-05
fix: update localhost port in examples from 8000 to 8010

parinporecha:
I was trying out the `example.py` using local PostHog instance when I ran into these errors -

`ERROR:posthog:[FEATURE FLAGS] Unable to get feature variants: HTTPConnectionPool(host='localhost', port=8000): Max retries exceeded with url: /flags/?v=2 (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x79abc05d9190>: Failed to establish a new connection: [Errno 111] Connection refused'))`

`ERROR:posthog:error uploading: HTTPConnectionPool(host='localhost', port=8000): Max retries exceeded with url: /batch/ (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x79abbf066840>: Failed to establish a new connection: [Errno 111] Connection refused'))`

`/flags` and `/batch` endpoints are not in the Django backend on `:8000`. For the local instance we need to point to the proxy port so it can reroute to the right service.


dustinbyrne (approved):
Thank you @parinporecha!

PR #462, opened 2026-03-11
chore(flags): expose flag_definition_cache_provider

matheus-vb:
## Problem

The PostHog monolith wants to use a `HyperCacheFlagProvider` (see [posthog#50737](https://github.com/PostHog/posthog/pull/50737)) to read flag definitions directly from Redis instead of polling the API every 90 seconds per worker. However, the SDK's module-level API has no way to accept an external cache provider — the `flag_definition_cache_provider` parameter only exists on the `Client` constructor, not as a module-level variable.

### Module-level `flag_definition_cache_provider`

Expose `flag_definition_cache_provider` as a module-level variable in `posthog/__init__.py` (matching the pattern used by all other SDK settings like `personal_api_key`, `poll_interval`, etc.) and pass it through `setup()` to the `Client` constructor. This allows the monolith to set `posthoganalytics.flag_definition_cache_provider = HyperCacheFlagProvider(...)` before any flags are evaluated.

## Changes

**`posthog/__init__.py`**
- Added `flag_definition_cache_provider` module-level variable (defaults to `None`)
- Passed it through `setup()` to the `Client` constructor

- Existing tests pass (`make test`)

## Downstream dependency

This PR must be merged and released before [posthog#50737](https://github.com/PostHog/posthog/pull/50737) can land, since the monolith PR sets `posthoganalytics.flag_definition_cache_provider` at startup.

### DIFF
--- a/.sampo/changesets/wily-king-nyyrikki.md
+++ b/.sampo/changesets/wily-king-nyyrikki.md
@@ -1,0 +1,5 @@
+---
+pypi/posthog: patch
+---
+
+chore(flags): expose flag_definition_cache_provider


haacked:
The PR description mentions additional changes (stale cache fallback for failed flags in client.py, types.py, and a new test), but I don't see those in this PR.

Is the PR description incorrect or did you forget to push?

matheus-vb:
Oops those changes were supposed to be on another PR, will clean up the description

haacked (approved):
:shipit:

PR #464, opened 2026-03-13
feat: add Celery integration and improve PostHog client fork safety

parinporecha:
## Summary
This PR:
- Adds a new `PosthogCeleryIntegration` to automatically capture Celery task lifecycle events and exceptions.
- Propagates PostHog context (`distinct_id`, `session_id`, tags) from the task producer to the worker so Celery tasks can be correlated with the originating user/session.
- Makes `Client` safer across process forks by reinitializing fork-unsafe client state in child processes.

```python
from posthog.integrations.celery import PosthogCeleryIntegration
integration = PosthogCeleryIntegration(
    capture_exceptions=True,
    capture_task_lifecycle_events=True,
    propagate_context=True
)
integration.instrument()
```

## Context
I saw users asking for advice on how to use PostHog with Celery for error tracking in [community questions](https://posthog.com/docs/error-tracking/installation/python) and realized that there's currently no first-class way to instrument Celery workloads with PostHog.

That leaves a few gaps:
- background task execution is hard to observe without manual instrumentation.
- worker-side events are difficult to correlate back to the originating user or request.

This PR addresses those gaps by adding a Celery integration that helps users observe task execution end-to-end out of the box.

The integration takes inspiration from OpenTelemetry's Celery [instrumentor](https://github.com/open-telemetry/opentelemetry-python-contrib/blob/main/instrumentation/opentelemetry-instrumentation-celery/src/opentelemetry/instrumentation/celery/__init__.py) and PostHog context propagation is achieved through task headers mirroring Sentry and DataDog's Celery integrations.

While testing this, I found a separate SDK issue: when a `Client` configured in async mode is inherited across a process fork, the child process inherits a client whose consumer threads no longer exist. In practice, that means worker-side events don't get delivered. This would also be a problem when using the SDK in some Django deployments.

So this PR also adds fork handling to `Client` by reinitializing its queue, consumers, and other state in the child process via `os.register_at_fork`.

### New: Celery Integration (`posthog/integrations/celery.py`)
-   **Lifecycle Events**: Hooks into Celery signals (`task_prerun`, `task_success`, `task_failure`, etc.) to capture events like `celery task started`, `celery task success` etc. Check the docstring in the integration module code for complete list of supported events.
    -   Lifecycle events include Celery-specific properties such as task ID, task name, queue, retry count, duration, Celery version etc. Check the docstring for complete set of event properties.
-   **Context Propagation**:
    -   `_on_before_task_publish`: Injects current PostHog context (distinct_id, session_id, tags) into task headers.
    -   `_on_task_prerun`: Extracts headers in the worker and restores the PostHog context for the duration of the task. This context is exited upon task completion.
    -    Any custom events captured inside a task inherit the same propagated PostHog context and Celery task tags.
-   **Exception Capture**: Automatically captures exceptions from failed tasks.

### Refactored: Client Fork Safety `posthog/client.py`
-   Added `_reinit_after_fork` method to reset the internal queue and spin up new consumers in a child process.
-   Uses `os.register_at_fork` (on supported platforms) to automatically call this method, so that the SDK does not drop captured events when used in child processes.

### Examples `examples/celery_integration.py`
-   Added a complete example showing how to configure the integration on both the producer and worker sides and all features in practice.

## Tests
-   Added `posthog/test/integrations/test_celery_integration.py` covering:
    -   Signal handlers for all task states.
    -   Context propagation (header injection and extraction).
    -   Task filtering logic.
    -   Exception capture.
-   Added `posthog/test/test_client_fork.py` covering:
    -   Unit tests for client's fork safety
    -   End to end tests forking Client and verifying behaviour
- Manually tested the example against Celery 5.2.1 (2021), 5.3.1, 5.4.0, and 5.6.2 (2026).
 
## Screenshots (created through example script)

- Celery task lifecycle events and captured Exception -
  image

- Celery task success event emitted from worker carrying correct distinct ID, session ID set in parent and context tags -
  image

- Captured exception -
  image

### DIFF
--- a/examples/celery_integration.py
+++ b/examples/celery_integration.py
@@ -1,0 +1,191 @@
+"""
+Celery integration example for the PostHog Python SDK.
+
+Demonstrates how to use ``PosthogCeleryIntegration`` with:
+- producer-side instrumentation (publishing events and context propagation)
+- worker-side instrumentation via ``worker_process_init`` (prefork-safe)
+- context propagation (distinct ID, session ID, tags) from producer to worker
+- task lifecycle events (published, started, success, failure, retry)
+- exception capture from failed tasks
+- ``task_filter`` customization hook
+
+Setup:
+    1. Update POSTHOG_PROJECT_API_KEY and POSTHOG_HOST here with your credentials
+       (environment variables won't work as it's better if Celery forks worker into
+        separate process for the example to prove context propagation)
+    2. Install dependencies: pip install posthog celery redis
+    3. Start Redis: redis-server
+    4. Start the worker: celery -A examples.celery_integration worker --loglevel=info
+    5. Run the producer: python -m examples.celery_integration
+"""
+
+import time
+from typing import Any, Optional
+
+from celery import Celery
+from celery.signals import worker_process_init, worker_process_shutdown
+
+import posthog
+from posthog.client import Client
+from posthog.integrations.celery import PosthogCeleryIntegration
+
+
+# --- Configuration ---
+
+POSTHOG_PROJECT_API_KEY = "phc_..."
+POSTHOG_HOST = "http://localhost:8000"
+
+app = Celery(
+    "examples.celery_integration",
+    broker="redis://localhost:6379/0",
+)
+
+
+# --- Integration wiring ---
+
+def create_client() -> Client:
+    return Client(
+        project_api_key=POSTHOG_PROJECT_API_KEY,
+        host=POSTHOG_HOST
+    )
+
+
+def task_filter(task_name: Optional[str], task_properties: dict[str, Any]) -> bool:
+    if task_name is not None and task_name.endswith(".health_check"):
+        return False
+    return True
+
+
+def create_integration(client: Client) -> PosthogCeleryIntegration:
+    return PosthogCeleryIntegration(
+        client=client,
+        capture_exceptions=True,
+        capture_task_lifecycle_events=True,
+        propagate_context=True,
+        task_filter=task_filter,
+    )
+
+
+# Worker process setup.
+# Celery's default prefork pool runs tasks in child processes, so initialize
+# PostHog per child using worker_process_init.
+
+
+@worker_process_init.connect
+def on_worker_process_init(**kwargs) -> None:
+    worker_posthog_client = create_client()
+    worker_integration = create_integration(worker_posthog_client)
+    worker_integration.instrument()
+    
+    app._posthog_client = worker_posthog_client
+    app._posthog_integration = worker_integration
+
+
+@worker_process_shutdown.connect
+def on_worker_process_shutdown(**kwargs) -> None:
+    worker_integration = getattr(app, "_posthog_integration", None)
+    if worker_integration:
+        worker_integration.uninstrument()
+
+    worker_posthog_client = getattr(app, "_posthog_client", None)
+    if worker_posthog_client:
+        worker_posthog_client.shutdown()
+
+# --- Example tasks ---
+
+@app.task
+def health_check() -> dict[str, str]:
+    return {"status": "ok"}
+
+
+@app.task(bind=True, max_retries=3)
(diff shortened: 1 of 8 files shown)


dustinbyrne (commented):
Hey @parinporecha, love the direction here. I left a few questions/considerations below, let me know what you think!

parinporecha:
@marandaneto is this not needed?

examples/celery_integration.py
```diff
@@ -0,0 +1,185 @@
...
+# If workers run across multiple hosts, configure PostHog and instrument a
+# worker-local integration in worker_process_init.
+@worker_process_init.connect
+def on_worker_process_init(**kwargs) -> None:
+    # global integration
+
+    # configure_posthog()
+    # integration = create_integration()
+    # integration.instrument()
+    return
```
dustinbyrne, line 82: Is this still necessary? It looks like it's a no-op

examples/celery_integration.py
```diff
@@ -0,0 +1,185 @@
...
+
+
+# Use this signal to shutdown the integration and PostHog client
+# Calling shutdown() is important to flush any pending events
+@worker_process_shutdown.connect
+def on_worker_process_shutdown(**kwargs) -> None:
+    integration.shutdown()
+    posthog.shutdown()
```
dustinbyrne, line 90: Is this required setup configuration or can this be omitted? We're also calling doing this at the bottom of the example, but it's not clear why or if it's required.

marandaneto:
> @marandaneto is this not needed?

i think this is useful, yes, and @dustinbyrne probably missed the ping
some PRs got auto closed because i changed the main branch from master -> main
mind opening up the PR again against main? and with signed commits otherwise we cant merge it

posthog/client.py
```diff
@@ -1080,6 +1086,55 @@ def capture_exception(
...
+
+        Registered via os.register_at_fork(after_in_child=...) so it runs
+        exactly once in each child, before any user code, covering all code
+        paths (capture, flush, join, etc.).
+
+        Python threads do not survive fork() and queue.Queue internal locks
+        may be in an inconsistent state, so both are replaced.
+        Inherited queue items are intentionally discarded as they'll be
+        handled by the parent process's consumers.
+        """
```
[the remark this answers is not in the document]
dustinbyrne: I'm a little torn on this, because it feels a bit like the Celery integration and `flag_definition_cache_provider` are at odds with each other, unless we _also_ expand the `FlagDefinitionCacheProvider` to handle `_reinit_after_fork`. In practice, Celery is a great use case for caching flag definitions.

Perhaps we could do something like:
```python
integration = PosthogCeleryIntegration(
    client_factory=create_posthog_client,
)
```

where:
```python
def create_posthog_client():
    cache = MyCacheProvider(...)
    return Posthog(
        "...",
        personal_api_key="...",
        flag_definition_cache_provider=cache,
    )
```

Then on `worker_process_init`, the integration could call the factory and store the client for that process and we avoid the whole fork.

examples/celery_integration.py
parinporecha: This is needed for the celery workers. The one at the bottom of the example does it for the producer process. I'll add comments explaining it

examples/celery_integration.py
parinporecha: Yeah this shows how to set it up if the workers are running on a different host. I'll comment out this handler entirely and make that clearer

posthog/client.py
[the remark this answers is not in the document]
parinporecha: I'm more inclined towards updating the docs/example to explicitly mention that users with a custom `flag_definition_cache_provider` should reinit their client in the workers and reinstrument the integration with that new client instance. That's more in line with what the other providers (Sentry, DD, OTel) recommend.

We can replacing client with factory but that would be a deviation from that standard practice which I guess users would be more likely to be used to.

Issue #450, opened 2026-03-03
`before_send` is completely ignored during module-level initialization in `__init__.py`


sipa-echo-zaoa:
## Root Cause
The `Client` class constructor in `posthog/client.py` correctly accepts the `before_send` argument. 

However, in `posthog/__init__.py`, the module-level state variables and the `setup()` function completely omit `before_send`. When `setup()` lazily instantiates the `Client`, it does not look for a global `before_send` variable, nor does it pass it into the constructor:

```python
# From posthog/__init__.py -> setup()
default_client = Client(
    api_key,
    host=host,
    debug=debug,
    on_error=on_error,
    send=send,
    # ... before_send is completely missing here ...
    in_app_modules=in_app_modules,
)
```

**Why this matters:** The official [Django integration documentation](https://posthog.com/docs/libraries/django) explicitly instructs developers to initialize PostHog globally inside `apps.py`:

```python
# From the official Django docs:
from django.apps import AppConfig
import posthog

class YourAppConfig(AppConfig):
    name = "your_app_name"
    def ready(self):
        posthog.api_key = '<ph_project_api_key>'
        posthog.host = 'https://eu.i.posthog.com'
        # Developers naturally try to add `posthog.before_send = my_filter` here, but it silently fails.
```

## Steps to Reproduce
Following the standard initialization pattern, before_send fails to attach:

```python
import posthog

def my_filter(event):
    print("This will never fire")
    return event

posthog.api_key = 'my-key'
posthog.before_send = my_filter # This variable is ignored by setup()

# Triggering an event initializes the client via setup(), but without the hook
posthog.capture('user_123', 'test_event')
```
## Environment
* **posthog-python version:** 7.9.4

PR #472, opened 2026-03-25
fix: pass before_send through module-level setup

sipa-echo-zaoa:
## Summary
- pass the module-level `posthog.before_send` callback into the lazily initialized default client
- add a regression test covering module-level initialization so `posthog.capture(...)` uses the callback
- unblock Django apps that rely on global setup to suppress unwanted exception events like `Http404` via `before_send`

Closes PostHog/posthog-python#450

### DIFF
--- a/posthog/__init__.py
+++ b/posthog/__init__.py
@@ -55,7 +55,11 @@ from posthog.request import (
     set_socket_options as set_socket_options,
     SocketOptions as SocketOptions,
 )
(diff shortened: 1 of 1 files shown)


sipa-echo-zaoa:
> > @marandaneto Yes sorry, wasn't familiar with commits signing. Now i've set it up and they should be fine.
> 
> you piushed again without signed commits

Sorry, I've rebased the branch from Github that invalidated the commits verification :sweat_smile:.
I've rebased it locally and push forced it again.

PR #503, opened 2026-04-13
chore: remove unused decide helper and stale references

marandaneto:
## Problem
The Python SDK no longer uses the legacy `/decide` helper for feature flag evaluation, but a dead helper and several stale comments/docstrings still referred to it. This made the code and tests misleading.

## Changes
- remove the unused `decide()` helper from `posthog/request.py`
- update feature flag docstrings and response-normalization docs to refer to `/flags`
- update request tests to exercise `flags()` instead of the removed helper
- rename stale `/decide` comments and fixture values in feature flag tests where they now describe `/flags` fallback
- format `posthog/test/test_request.py` with Ruff

## Checklist
- [x] Tests added/updated where relevant
- [x] Verified locally (`uv run pytest posthog/test/test_request.py posthog/test/test_types.py -q`, `uv run pytest posthog/test/test_feature_flags.py -q`, and `uvx ruff format --check .`)
- [ ] Changeset added
- [ ] Release label added

### DIFF
--- a/posthog/request.py
+++ b/posthog/request.py
@@ -73,7 +73,7 @@ class HTTPAdapterWithSocketOptions(HTTPAdapter):
 
 
 def _build_session(socket_options: Optional[SocketOptions] = None) -> requests.Session:
-    """Build a session for general requests (batch, decide, etc.)."""
+    """Build a session for general requests (batch, remote config, etc.)."""
     adapter = HTTPAdapterWithSocketOptions(
         max_retries=Retry(
             total=2,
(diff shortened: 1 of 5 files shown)


posthog/request.py
```diff
@@ -262,18 +262,6 @@ def _process_response(
         raise APIError(res.status_code, res.text, retry_after=retry_after)
 
 
-def decide(
```
marandaneto, line 265: this wasnt being called anywhere anymore so i assume its left over?

dustinbyrne (approved):

posthog/request.py
marandaneto: @PostHog/team-feature-flags can you confirm before i pull the plug?

dmarticus (approved):
🙏 🙏 🙏

PR #523, opened 2026-04-17
feat(flags): support mixed targeting in local evaluation

patricio-posthog:
## Problem

Flags using mixed user+group targeting (the beta feature) cannot be evaluated locally by the SDK. The SDK reads `aggregation_group_type_index` at the flag level only. For mixed flags this is `None`, so all conditions are evaluated as person targeting, causing group conditions to fail and fall back to an HTTP call.

This breaks customers using flags in environments that cannot make HTTP calls (e.g., Temporal workflows).

## Changes

Updated `match_feature_flag_properties` to resolve aggregation per condition when a condition explicitly sets its own `aggregation_group_type_index` that differs from the flag level. Each condition uses the correct properties and bucketing value for its aggregation type. Backwards compatible with existing pure person and pure group flags.

All 115 existing tests pass unchanged. Added 6 new tests covering mixed targeting scenarios: person match, group match, no match, group skips to person, no groups passed, and correct bucketing per condition.


posthog/feature_flags.py
```diff
@@ -305,32 +308,58 @@ def match_feature_flag_properties(
...
+    group_type_mapping = group_type_mapping or {}
     # Some filters can be explicitly set to null, which require accessing variants like so
     flag_variants = (flag_filters.get("multivariate") or {}).get("variants") or []
     valid_variant_keys = [variant["key"] for variant in flag_variants]
 
     for condition in flag_conditions:
         try:
-            # if any one condition resolves to True, we can shortcircuit and return
-            # the matching variant
+            # Per-condition aggregation overrides only when the condition explicitly
```
haacked: `suggestion`: This comment says the override kicks in "only when the condition explicitly sets its own `aggregation_group_type_index`", but the actual trigger is `condition_aggregation != flag_aggregation`. A condition that explicitly sets the same value as the flag, or explicitly sets `None` when the flag is also `None`, won't trigger the override. A future maintainer will grep for an `"aggregation_group_type_index" in condition` check and not find one.

```suggestion
            # When a condition's aggregation_group_type_index differs from the
            # flag-level value (mixed targeting), resolve properties/bucketing
            # for that condition's aggregation. Otherwise use whatever the
            # caller already resolved.
```

posthog/feature_flags.py
```diff
@@ -305,32 +308,58 @@ def match_feature_flag_properties(
...
+            # When absent, use the properties/bucketing already resolved by the caller.
+            condition_aggregation = condition.get(
+                "aggregation_group_type_index", flag_aggregation
+            )
+
+            if condition_aggregation != flag_aggregation:
+                if condition_aggregation is not None:
+                    group_name = group_type_mapping.get(str(condition_aggregation))
+                    if not group_name or group_name not in groups:
+                        continue
```
haacked: `nit`: When `group_name not in groups`, the condition is silently skipped. The analogous path in `_compute_flag_locally` (`client.py:1456-1463`) logs a warning. A mixed flag silently evaluating to False because the caller forgot to pass `groups` is a worse debugging experience than the pure-group path. Consider a debug-level log (not warning, since skipping is expected in many calls):

```python
if not group_name or group_name not in groups:
    log.debug(
        f"[FEATURE FLAGS] Skipping group condition for flag {flag.get('key')}: "
        f"group {group_name!r} not in passed groups"
    )
    continue
```

posthog/feature_flags.py
```diff
@@ -305,32 +308,58 @@ def match_feature_flag_properties(
...
+                    if group_name not in group_properties:
+                        is_inconclusive = True
+                        continue
+                    effective_properties = group_properties[group_name]
+                    effective_bucketing = groups[group_name]
+                else:
```
haacked: `question`: For a pure-group flag (`flag_aggregation = 0`) with a condition that explicitly sets `aggregation_group_type_index: None` (a person-targeted condition inside an otherwise group-level flag), this inner `else` assigns `effective_properties = properties` and `effective_bucketing = bucketing_value`. But `_compute_flag_locally` at `client.py:1475-1480` passed `group_properties[group_name]` as `properties` and `group_key` as `bucketing_value` because it saw flag-level group aggregation. The person condition would then evaluate person property filters against group properties and bucket rollout on the group key. If the server never emits this shape today this is theoretical, but nothing here guards against it. Is this intentional, or should this path set `is_inconclusive = True` and continue so it falls back to the API?

posthog/feature_flags.py
```diff
@@ -305,32 +308,58 @@ def match_feature_flag_properties(
...
+                    effective_properties = properties
+                    effective_bucketing = bucketing_value
+            else:
```
haacked: `suggestion`: This block has three branches where two do the same thing. Both the outer `else` (same-as-flag) and the inner `else` at line 340 (person override) assign `effective_properties = properties` and `effective_bucketing = bucketing_value`; only the group branch actually differs. Setting the defaults first and only overriding in the group branch reads more directly:

```python
condition_aggregation = condition.get(
    "aggregation_group_type_index", flag_aggregation
)
effective_properties = properties
effective_bucketing = bucketing_value

if (
    condition_aggregation != flag_aggregation
    and condition_aggregation is not None
):
    group_name = group_type_mapping.get(str(condition_aggregation))
    if not group_name or group_name not in groups:
        continue
    if group_name not in group_properties:
        is_inconclusive = True
        continue
    effective_properties = group_properties[group_name]
    effective_bucketing = groups[group_name]
```

.sampo/changesets/mixed-targeting-local-eval.md
```diff
@@ -0,0 +1,5 @@
+---
+pypi/posthog: patch
```
haacked: `suggestion`: Consider bumping from `patch` to `minor`. This PR changes local evaluation results for an existing flag shape: a flag with a per-condition `aggregation_group_type_index` different from the flag-level value was previously evaluated entirely against caller-provided `properties`/`bucketing_value`, and now uses per-condition group data. Observable behavior change under a `feat:` commit reads more like `minor` than `patch`.

```suggestion
pypi/posthog: minor
```

posthog/test/test_feature_flags.py
```diff
@@ -3706,6 +3707,145 @@ def test_group_flag_dependency_ignores_device_id_bucketing_identifier(
         self.assertTrue(result)
         self.assertEqual(patch_flags.call_count, 0)
 
+    MIXED_FLAG = {
```
haacked: `nit`: `MIXED_FLAG` is the only class-level UPPER_CASE fixture in `TestLocalEvaluation`. Every other test in this class (including the two new non-parameterized tests in this PR) defines its flag dict inline in the test method, so this sits between test methods as a class attribute and reads like an accidental leak. Either move it to module scope above the class, or inline it into the parameterized test to match the file's convention.

marandaneto (approved):

posthog/feature_flags.py
```diff
@@ -305,32 +308,58 @@ def match_feature_flag_properties(
...
+            # sets its own aggregation_group_type_index (mixed targeting).
+            # When absent, use the properties/bucketing already resolved by the caller.
+            condition_aggregation = condition.get(
+                "aggregation_group_type_index", flag_aggregation
+            )
+
+            if condition_aggregation != flag_aggregation:
+                if condition_aggregation is not None:
+                    group_name = group_type_mapping.get(str(condition_aggregation))
+                    if not group_name or group_name not in groups:
```
dmarticus, line 335: Nit: this silently skips when the condition's `aggregation_group_type_index` doesn't map to a known group (or the caller didn't pass the group). The pure-group-flag path in `_compute_flag_locally` emits a `log.warning` for the equivalent case. A `log.debug` here would help diagnose misconfigured mixed flags without turning every per-condition skip into a warning.

posthog/feature_flags.py
```diff
@@ -305,32 +308,58 @@ def match_feature_flag_properties(
...
+                        continue
+                    if group_name not in group_properties:
+                        is_inconclusive = True
+                        continue
+                    effective_properties = group_properties[group_name]
+                    effective_bucketing = groups[group_name]
+                else:
+                    effective_properties = properties
```
dmarticus, line 348: Worth a short comment documenting the contract: this mixed-override branch assumes `flag_aggregation is None` (the only shape PostHog's UI produces today). If `flag_aggregation` is set to a group index and a condition overrides with `None`, this falls through to `effective_properties = properties`, but in that caller branch `properties` is `focused_group_properties`, not person properties — so person conditions would silently evaluate against group props. Not a regression and not a change you need to make here, but documenting the assumption will save the next person a confusing debugging session.

.sampo/changesets/mixed-targeting-local-eval.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Support mixed user+group targeting in local flag evaluation. Flags with per-condition `aggregation_group_type_index` now resolve properties and bucketing per condition instead of using the flag-level aggregation only.
```
dmarticus: Consider tightening to a single sentence, e.g. `Support mixed user+group targeting in local flag evaluation.` The second sentence is implementation detail that isn't especially user-facing.

dmarticus (approved):
## Review

Nice fix — scope is tight and the test coverage lines up well with the behaviors I care about (per-condition bucketing identity, skip-vs-inconclusive split, no-groups-passed). Approving.

I left three non-blocking inline comments on specific lines. One more observation that didn't fit inline:

- **Cohort properties on group-aggregated conditions (follow-up, out of scope).** A group-aggregated condition containing a cohort property would pass group props into `match_cohort`, but cohorts are person-defined. Today the UI probably doesn't let you build this shape, but if it ever does we'd evaluate the cohort against the wrong property bag. Worth a follow-up issue if you agree.

---

### Pastable prompt to address these comments

If you want to knock these out in one pass, feel free to paste this into your editor:

> In `posthog/feature_flags.py`, inside `match_feature_flag_properties`:
>
> 1. When a condition's `aggregation_group_type_index` resolves to an unknown group (not in `group_type_mapping`, or missing from `groups`), add a `log.debug(...)` before `continue` so misconfigured mixed flags are diagnosable. Mirror the message style used in `_compute_flag_locally`'s pure-group-flag path, but at debug level (not warning) since per-condition skips are expected.
>
> 2. Add a one-line comment above the `if condition_aggregation != flag_aggregation:` block noting that the mixed-override path assumes flag-level aggregation is `None`; the inverse shape (flag-level set, condition-level `None`) isn't supported by the current caller and would silently pass group props as person props. Documentation only — no code change.
>
> 3. In `.sampo/changesets/mixed-targeting-local-eval.md`, tighten the changeset body to a single user-facing sentence, e.g. `Support mixed user+group targeting in local flag evaluation.` The implementation detail in the current second sentence isn't something end users need.
>
> 4. (Optional follow-up, separate PR.) Cohort properties evaluated inside a group-aggregated condition will be matched against group props, not person props. File an issue or add a TODO if you agree it's worth tracking.
>
> No new tests required for any of these.

---

PR #532, opened 2026-04-23
feat: add get_feature_flags for bulk flag evaluation with events

dmarticus:
## Summary

Add a new `get_feature_flags(keys, distinct_id, ...)` method for evaluating a known subset of feature flags in one bulk pass while still emitting `$feature_flag_called` events per resolved flag. Also adds a `send_feature_flag_events` option to the existing `get_all_flags` / `get_all_flags_and_payloads` methods for opt-in per-flag event emission.

The motivation (per the original PR): customers loading a page server-side often need a specific ~10 flags. `get_feature_flag` per key is N network calls; `get_all_flags` is one call but emits no `$feature_flag_called` events and evaluates flags the caller doesn't care about. The new method is the sweet spot: at most one network round trip, per-flag usage events.

Ports [PostHog/posthog-js#3447](https://github.com/PostHog/posthog-js/pull/3447) from `posthog-node` to `posthog-python`.

## Changes

- **New `get_feature_flags` method** (`posthog/client.py`): evaluates a subset of flags in bulk. Locally-evaluated flags reuse the poller's cached definitions; keys that can't be resolved locally fall through to a single remote `/flags` call with `flag_keys_to_evaluate`.
- **Event deduplication**: reuses the existing `distinct_ids_feature_flags_reported` cache so single-flag and bulk paths share one source of truth.
- **New `send_feature_flag_events` option** on `get_all_flags` / `get_all_flags_and_payloads`. Defaults to `False` to preserve existing behavior.
- **Top-level proxy** (`posthog/__init__.py`): adds `posthog.get_feature_flags(...)` and threads `send_feature_flag_events` through the module-level `get_all_flags` / `get_all_flags_and_payloads` wrappers.
- **Tests** (`posthog/test/test_feature_flags.py`): new `TestGetFeatureFlagsBulk` class with 9 tests covering remote-only evaluation, local-only evaluation, hybrid scenarios, event deduplication, `send_feature_flag_events=False` suppression, missing-key handling, and the new `get_all_flags` option.

## Test plan

- [x] `uv run pytest posthog/test/test_feature_flags.py::TestGetFeatureFlagsBulk` — 9/9 pass
- [x] `uv run pytest posthog/test/` — 324/324 pass (no regressions; `send_feature_flag_events=False` default preserves historical behavior)
- [x] `uv run mypy posthog/client.py posthog/__init__.py` — no new errors vs. baseline
- [x] `uvx ruff format` — clean

---
[...]

PR #534, opened 2026-04-23
feat: add get_feature_flags for bulk flag evaluation with events

dmarticus:
## Summary

Add a new `get_feature_flags(keys, distinct_id, ...)` method for evaluating a known subset of feature flags in one bulk pass while still emitting `$feature_flag_called` events per resolved flag. Also adds a `send_feature_flag_events` option to the existing `get_all_flags` / `get_all_flags_and_payloads` methods for opt-in per-flag event emission.

The motivation (per the original PR): customers loading a page server-side often need a specific ~10 flags. `get_feature_flag` per key is N network calls; `get_all_flags` is one call but emits no `$feature_flag_called` events and evaluates flags the caller doesn't care about. The new method is the sweet spot: at most one network round trip, per-flag usage events.

Ports [PostHog/posthog-js#3447](https://github.com/PostHog/posthog-js/pull/3447) from `posthog-node` to `posthog-python`. Supersedes #532 (closed because its branch was based on a stale commit that predated several unsigned release commits on main, which blocked a clean rebase-and-push under the repo's signature policy).

## Changes

- **New `get_feature_flags` method** (`posthog/client.py`): evaluates a subset of flags in bulk. Locally-evaluated flags reuse the poller's cached definitions; keys that can't be resolved locally fall through to a single remote `/flags` call with `flag_keys_to_evaluate`. Threads `device_id` and participates in the same error handling (quota, timeout, connection, api-error, unknown) + stale-cache fallback as `_get_feature_flag_result`.
- **Event deduplication**: reuses the existing `distinct_ids_feature_flags_reported` cache so single-flag and bulk paths share one source of truth.
- **New `send_feature_flag_events` option** on `get_all_flags` / `get_all_flags_and_payloads`. Defaults to `False` to preserve existing behavior.
- **Top-level proxy** (`posthog/__init__.py`): adds `posthog.get_feature_flags(...)` and threads `send_feature_flag_events` through the module-level `get_all_flags` / `get_all_flags_and_payloads` wrappers.
- **Tests** (`posthog/test/test_feature_flags.py`): new `TestGetFeatureFlagsBulk` class with 9 tests covering remote-only evaluation, local-only evaluation, hybrid scenarios, event deduplication, `send_feature_flag_events=False` suppression, missing-key handling (with `$feature_flag_error=flag_missing`), and the new `get_all_flags` option.

## Test plan

- [x] `uv run pytest posthog/test/test_feature_flags.py::TestGetFeatureFlagsBulk` — 9/9 pass
- [x] `uv run pytest posthog/test/ --ignore=posthog/test/ai` — 459/459 pass (no regressions; `send_feature_flag_events=False` default preserves historical behavior)
- [x] `uv run mypy posthog/client.py posthog/__init__.py` — no new errors vs. baseline
- [x] `uvx ruff format` — clean

---

### DIFF
--- a/posthog/__init__.py
+++ b/posthog/__init__.py
@@ -617,6 +617,7 @@ def get_all_flags(
     disable_geoip=None,  # type: Optional[bool]
     device_id=None,  # type: Optional[str]
     flag_keys_to_evaluate=None,  # type: Optional[list[str]]
+    send_feature_flag_events=False,  # type: bool
 ) -> Optional[dict[str, FeatureFlag]]:
     """
     Get all flags for a given user.
@@ -629,6 +630,8 @@ def get_all_flags(
         only_evaluate_locally: Whether to evaluate only locally
         disable_geoip: Whether to disable GeoIP lookup
         flag_keys_to_evaluate: Optional list of flag keys to evaluate (evaluates all if None)
+        send_feature_flag_events: If True, emit a `$feature_flag_called` event for each
+            resolved flag. Defaults to False to preserve historical behavior.
 
     Details:
         Flags are key-value pairs where the key is the flag key and the value is the flag variant, or True, or False.
@@ -652,6 +655,7 @@ def get_all_flags(
         disable_geoip=disable_geoip,
         device_id=device_id,
         flag_keys_to_evaluate=flag_keys_to_evaluate,
+        send_feature_flag_events=send_feature_flag_events,
     )
 
 
@@ -756,6 +760,7 @@ def get_all_flags_and_payloads(
     disable_geoip=None,  # type: Optional[bool]
     device_id=None,  # type: Optional[str]
     flag_keys_to_evaluate=None,  # type: Optional[list[str]]
+    send_feature_flag_events=False,  # type: bool
 ) -> FlagsAndPayloads:
     return _proxy(
         "get_all_flags_and_payloads",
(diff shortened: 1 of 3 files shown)


posthog/__init__.py
```diff
@@ -617,6 +617,7 @@ def get_all_flags(
     disable_geoip=None,  # type: Optional[bool]
     device_id=None,  # type: Optional[str]
     flag_keys_to_evaluate=None,  # type: Optional[list[str]]
+    send_feature_flag_events=False,  # type: bool
```
marandaneto, line 620: iirc there was rfc to kill this flag due to its side effect
https://github.com/PostHog/requests-for-comments-internal/pull/1020 i think
so before adding this flag to more sdks, should we double check if we actually want to do this?

dmarticus: so my interpretation of that SDK is to remove it from methods like e.g. `capture`, since we really don't need to send flag data for non-flag calls. Me adding to to `get_all_flags` here was to make it so that people could still use `get_all_flags` and send the flag values as events. what I noticed this week when talking to a customer was that they basically call `get_all_flags` once as part of the initial page load, save those values into their own app memory, and then use them throughout. This would make it so that they don't have to refactor their code to use my proposed new `get_feature_flags()` method, and could instead just add a field to their existing code.

but I read that RFC and it's a much saner approach IMO, so maybe rather than me adding this new method that follows the existing patterns, I should do the heavier lift and just implement the RFC. The reason I'm hesitant is because it turns this into a heavier lift for the customer – rather than them just bumping the SDK and adding this new method, they have to bump to a bigger version and deprecate their other methods everywhere. Maybe in LLM world, that's just as easy, but I wanted to be cognizant of that.

dmarticus:
closing in favor of a more principled approach, see https://posthog.slack.com/archives/C07Q2U4BH4L/p1777054243923129

PR #538, opened 2026-04-27
fix: sanitize Django tracing headers

dustinbyrne:

Django middleware extracts PostHog tracing headers from incoming requests and uses them to set context values. These headers are user-controlled, so sanitize them before they enter context/event properties.

This keeps the existing session/distinct-id behavior, but now strips control characters, trims whitespace, caps header length, ignores non-string values, and falls back to the request user when a distinct-id header sanitizes to empty.

This brings us to parity with how we currently handle these headers in PostHog Django

- `uv run pytest posthog/test/integrations/test_middleware.py -q`

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/posthog/integrations/django.py
+++ b/posthog/integrations/django.py
@@ -1,4 +1,6 @@
-from typing import TYPE_CHECKING, cast
+import re
+from typing import TYPE_CHECKING, Optional, cast
+
 from posthog import contexts
 from posthog.client import Client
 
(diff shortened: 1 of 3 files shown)


turnipdabeets (approved):

posthog/integrations/django.py
```diff
@@ -18,7 +20,36 @@ def markcoroutinefunction(func):
 
 if TYPE_CHECKING:
     from django.http import HttpRequest, HttpResponse  # noqa: F401
-    from typing import Callable, Dict, Any, Optional, Union, Awaitable  # noqa: F401
+    from typing import Callable, Dict, Any, Union, Awaitable  # noqa: F401
+
+
+_MAX_TRACING_HEADER_LENGTH = 1000
```
ioannisj, line 26: Just curious, why this value? Is this to match what PostHog app is doing?

ioannisj (approved):

posthog/integrations/django.py
dustinbyrne: Yeah, we're doing this same thing in Django

posthog/integrations/django.py
```diff
@@ -126,15 +159,22 @@ def _build_tags(self, request, user_id, user_email):
...
+        distinct_id = (
+            _get_sanitized_tracing_header(request, "X-POSTHOG-DISTINCT-ID") or user_id
+        )
         if distinct_id:
             contexts.identify_context(distinct_id)
 
+        # Extract window ID from X-POSTHOG-WINDOW-ID header
+        window_id = _get_sanitized_tracing_header(request, "X-POSTHOG-WINDOW-ID")
+        if window_id:
+            contexts.set_context_window_id(window_id)
```
marandaneto: Curious if we actually need the window_id, this is only used for session replay since it may have N tabs (or windows/screens/etc - eg foldable devices)
Is there a need for that? isnt session_id alone enough?

dustinbyrne: My thinking was that if we wanted to associate, e.g., a backend exception with a session recording, we'd need the window ID?

The `x-posthog-window-id` header is only sent from web (I didn't copy it over to mobile), so this would only de-duplicate sessions from different tabs.

marandaneto: i think we only need the session id, exceptions dont have a window id afaik
but better to check with @PostHog/team-replay or maybe @pauldambra

dustinbyrne: You're right, got an answer via Slack. I'll adjust this.

marandaneto (approved):

PR #539, opened 2026-04-27
feat: add evaluate_flags() API for single-call flag evaluation

dmarticus:
## Problem

Phase 1 + Phase 2 of the Server SDK Feature Flag Evaluations RFC for `posthog-python`. Companion to the Node SDK PR ([PostHog/posthog-js#3476](https://github.com/PostHog/posthog-js/pull/3476)).

Today every flag check fires its own `/flags` request, and `capture(send_feature_flags=True)` silently fires yet another on every captured event. The flag values on a captured event can diverge from the ones the code actually branched on when person/group properties differ between calls. `send_feature_flags` also attaches every evaluated flag to every event, which bloats properties on high-volume events.

### New API (Phase 1)

`posthog.evaluate_flags(distinct_id, ...)` returns a `FeatureFlagEvaluations` snapshot:

```python
flags = posthog.evaluate_flags(distinct_id, person_properties={"plan": "enterprise"})
if flags.is_enabled("new-dashboard"):
    render_new_dashboard()
posthog.capture("page_viewed", distinct_id=distinct_id, flags=flags)
```

A single `/flags` request powers both branching and event enrichment. `is_enabled()` and `get_flag()` fire `$feature_flag_called` events (deduped through the existing cache) with the full metadata — `$feature_flag_id`, `$feature_flag_version`, `$feature_flag_reason`, `$feature_flag_request_id` — so experiment exposure tracking keeps working.

### Two layers of scoping

- **Network-level (`flag_keys` option):** scopes the underlying `/flags` request itself.

  ```python
  flags = posthog.evaluate_flags(distinct_id, flag_keys=["new-dashboard", "checkout-flow"])
  ```

- **Event-level (filter helpers):** narrow which flags get attached to a captured event without re-fetching.

  ```python
  posthog.capture(..., flags=flags.only_accessed())           # only flags the developer checked
  posthog.capture(..., flags=flags.only(["new-dashboard"]))   # specific keys
  ```

### Deprecation warnings (Phase 2)

The legacy single-flag surface keeps working but now emits `DeprecationWarning`s pointing at `evaluate_flags()`:

- `feature_enabled()`
- `get_feature_flag()`
- `get_feature_flag_payload()`
- `capture(send_feature_flags=...)` (only when truthy)

`feature_enabled` and `get_feature_flag` are restructured to call `_get_feature_flag_result` directly instead of routing through each other, so a single user-level call emits exactly one warning instead of cascading.

Phase 3 (removal in next major) ships separately.

### Local evaluation

Transparent. When the poller resolves a flag, the snapshot carries `locally_evaluated=True` and reason `"Evaluated locally"`, matching what `get_feature_flag()` emits today.

### Backwards compatibility

**No breaking changes.** All existing call paths return the same values they did before — the only behavior change is the new `DeprecationWarning` emissions, which can be silenced via Python's standard warnings filter.

### Internals

`_capture_feature_flag_called` was refactored: the dedup + capture portion is extracted into `_capture_feature_flag_called_if_needed`, which is shared between the single-flag path and the new `FeatureFlagEvaluations` object. Both paths now dedupe identically.

Response-level errors (`errors_while_computing_flags`, `quota_limited`) are propagated into `$feature_flag_called` events from the snapshot, matching the granularity of the single-flag path.

## Tests

`posthog/test/test_evaluate_flags.py` — 27 tests covering remote evaluation, local evaluation, filtering helpers, capture integration, `flag_keys` round-trip, empty-distinct_id safety, error-granularity propagation, and deprecation warning emission (with no-cascade verification).

Full suite: 489 passed. `ruff format` and `ruff check` clean.

### DIFF
--- a/.sampo/changesets/evaluate-flags-api.md
+++ b/.sampo/changesets/evaluate-flags-api.md
@@ -1,0 +1,16 @@
+---
+"posthog": minor
+---
+
+Add `evaluate_flags()` and a new `flags` option on `capture()` so a single `/flags` call can power both flag branching and event enrichment per request:
+
+```python
+flags = posthog.evaluate_flags(distinct_id, person_properties={"plan": "enterprise"})
+if flags.is_enabled("new-dashboard"):
+    render_new_dashboard()
+posthog.capture("page_viewed", distinct_id=distinct_id, flags=flags)
+```
+
+The returned `FeatureFlagEvaluations` snapshot exposes `is_enabled()`, `get_flag()`, `get_flag_payload()` for branching and `only_accessed()` / `only([keys])` filter helpers. Pass `flag_keys=[...]` to `evaluate_flags()` to scope the underlying `/flags` request itself.
+
+`feature_enabled()`, `get_feature_flag()`, `get_feature_flag_payload()`, and `capture(send_feature_flags=...)` now emit `DeprecationWarning` and will be removed in the next major version.
--- a/.sampo/changesets/evaluate-flags-api.md
+++ b/.sampo/changesets/evaluate-flags-api.md
@@ -13,4 +13,4 @@ posthog.capture("page_viewed", distinct_id=distinct_id, flags=flags)
 
 The returned `FeatureFlagEvaluations` snapshot exposes `is_enabled()`, `get_flag()`, `get_flag_payload()` for branching and `only_accessed()` / `only([keys])` filter helpers. Pass `flag_keys=[...]` to `evaluate_flags()` to scope the underlying `/flags` request itself.
 
-Existing `feature_enabled()`, `get_feature_flag()`, `get_feature_flag_payload()`, and `capture(send_feature_flags=...)` continue to work unchanged. They will be deprecated in a follow-up minor and removed in the next major version.
+Deprecates `feature_enabled()`, `get_feature_flag()`, `get_feature_flag_payload()`, and `capture(send_feature_flags=...)`. They continue to work but now emit a `DeprecationWarning` pointing at `evaluate_flags()`. Removal is planned for the next major version.
--- a/.sampo/changesets/evaluate-flags-api.md
+++ b/.sampo/changesets/evaluate-flags-api.md
@@ -13,4 +13,4 @@ posthog.capture("page_viewed", distinct_id=distinct_id, flags=flags)
 
 The returned `FeatureFlagEvaluations` snapshot exposes `is_enabled()`, `get_flag()`, `get_flag_payload()` for branching and `only_accessed()` / `only([keys])` filter helpers. Pass `flag_keys=[...]` to `evaluate_flags()` to scope the underlying `/flags` request itself.
 
-`feature_enabled()`, `get_feature_flag()`, `get_feature_flag_payload()`, and `capture(send_feature_flags=...)` now emit `DeprecationWarning` and will be removed in the next major version.
+Existing `feature_enabled()`, `get_feature_flag()`, `get_feature_flag_payload()`, and `capture(send_feature_flags=...)` continue to work unchanged. They will be deprecated in a follow-up minor and removed in the next major version.
--- a/posthog/__init__.py
+++ b/posthog/__init__.py
@@ -781,6 +781,7 @@ def evaluate_flags(
     only_evaluate_locally=False,  # type: bool
     disable_geoip=None,  # type: Optional[bool]
     flag_keys=None,  # type: Optional[list]
+    device_id=None,  # type: Optional[str]
 ) -> FeatureFlagEvaluations:
     """Evaluate all feature flags for a user in a single call and return a
     :class:`FeatureFlagEvaluations` snapshot. Branch on ``.is_enabled()`` /
@@ -803,6 +804,9 @@ def evaluate_flags(
             evaluated — the underlying ``/flags`` request asks the server for just
             this subset, which makes the response smaller and the request cheaper.
             Use this when you only need a handful of flags out of many.
+        device_id: Optional device ID override. If not provided, falls back to the
+            context device_id (which may be set via tracing headers). Used by
+            experience-continuity flags to match users across distinct_id changes.
 
     Examples:
         ```python
@@ -825,6 +829,7 @@ def evaluate_flags(
         only_evaluate_locally=only_evaluate_locally,
         disable_geoip=disable_geoip,
         flag_keys=flag_keys,
+        device_id=device_id,
     )
 
 
--- a/posthog/__init__.py
+++ b/posthog/__init__.py
@@ -39,6 +39,9 @@ from posthog.exception_utils import (
     DEFAULT_CODE_VARIABLES_IGNORE_PATTERNS,
     DEFAULT_CODE_VARIABLES_MASK_PATTERNS,
 )
+from posthog.feature_flag_evaluations import (
+    FeatureFlagEvaluations as FeatureFlagEvaluations,
+)
 from posthog.feature_flags import (
     InconclusiveMatchError as InconclusiveMatchError,
 )
@@ -767,6 +770,59 @@ def get_all_flags_and_payloads(
         disable_geoip=disable_geoip,
         device_id=device_id,
         flag_keys_to_evaluate=flag_keys_to_evaluate,
+    )
+
+
+def evaluate_flags(
+    distinct_id=None,  # type: Optional[str]
+    groups=None,  # type: Optional[Dict[str, str]]
+    person_properties=None,  # type: Optional[Dict[str, Any]]
+    group_properties=None,  # type: Optional[Dict[str, Dict[str, Any]]]
+    only_evaluate_locally=False,  # type: bool
+    disable_geoip=None,  # type: Optional[bool]
+    flag_keys=None,  # type: Optional[list]
+) -> FeatureFlagEvaluations:
+    """Evaluate all feature flags for a user in a single call and return a
+    :class:`FeatureFlagEvaluations` snapshot. Branch on ``.is_enabled()`` /
+    ``.get_flag()`` and pass the same snapshot to ``capture()`` via the
+    ``flags`` option so events carry the exact flag values the code branched on.
+
+    Prefer this over repeated ``get_feature_flag()`` calls and over
+    ``capture(send_feature_flags=True)`` — it consolidates flag evaluation into
+    a single ``/flags`` request per incoming request.
+
+    Args:
+        distinct_id: The user's distinct ID. If ``None``, falls back to the context
+            distinct_id. If still unresolvable, returns an empty snapshot.
+        groups: Mapping of group type to group key.
+        person_properties: Person properties to use for evaluation.
+        group_properties: Group properties keyed by group type.
+        only_evaluate_locally: If ``True``, never fall back to remote evaluation.
+        disable_geoip: Whether to disable GeoIP lookup.
+        flag_keys: Optional list of flag keys to scope the underlying ``/flags``
+            request to a subset.
+
+    Examples:
+        ```python
+        from posthog import evaluate_flags, capture
+        flags = evaluate_flags("user_123", person_properties={"plan": "enterprise"})
+        if flags.is_enabled("new-dashboard"):
+            render_new_dashboard()
+        capture("page_viewed", distinct_id="user_123", flags=flags)
+        ```
+
+    Category:
+        Feature flags
+    """
+    return _proxy(
+        "evaluate_flags",
+        distinct_id=distinct_id,
+        groups=groups,
+        person_properties=person_properties,
+        group_properties=group_properties,
+        only_evaluate_locally=only_evaluate_locally,
+        disable_geoip=disable_geoip,
+        flag_keys=flag_keys,
     )
 
 
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -1007,7 +1007,9 @@ class Client(object):
             exception: The exception to capture.
             distinct_id: The distinct ID of the user.
             properties: A dictionary of additional properties.
-            send_feature_flags: Whether to send feature flags with the exception.
+            flags: A ``FeatureFlagEvaluations`` snapshot from ``evaluate_flags()``.
+                Attaches those exact flag values to the captured `$exception` event.
+            send_feature_flags: Deprecated. Pass ``flags`` from ``evaluate_flags()`` instead.
             disable_geoip: Whether to disable GeoIP for this event.
 
         Examples:
@@ -1024,6 +1026,7 @@ class Client(object):
         """
         distinct_id = kwargs.get("distinct_id", None)
         properties = kwargs.get("properties", None)
+        flags_snapshot = kwargs.get("flags", None)
         send_feature_flags = kwargs.get("send_feature_flags", False)
         disable_geoip = kwargs.get("disable_geoip", None)
         # this function shouldn't ever throw an error, so it logs exceptions instead of raising them.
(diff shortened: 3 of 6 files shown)


posthog/feature_flag_evaluations.py
```diff
@@ -0,0 +1,254 @@
...
+        out of order. Pre-access before calling this if you want a guaranteed-empty
+        result.
+        """
+        if not self._accessed:
+            self._host.log_warning(
+                "FeatureFlagEvaluations.only_accessed() was called before any flags were accessed — "
+                "attaching all evaluated flags as a fallback. See "
+                "https://posthog.com/docs/feature-flags/server-sdks for details."
+            )
+            return self._clone_with(self._flags)
```
dustinbyrne: If this is a legitimate case of nothing being accessed, there's no way to avoid getting all flags. This could be more confusing than getting nothing?

E.g., I'm imaging a case where flags are first retrieved and consumed as needed. Should an early call to `capture(flags=flags.only_accessed())` include all flags? I think it's contradictory with the name `only_accessed`

posthog/client.py
dustinbyrne, line 516: Should we deprecate the other methods?

posthog/client.py
```diff
@@ -2190,6 +2227,193 @@ def get_all_flags_and_payloads(
...
+        locally_evaluated_keys: set[str] = set()
+
+        # Try local evaluation first when the poller has loaded definitions.
+        local_result, fallback_to_server = self._get_all_flags_and_payloads_locally(
+            distinct_id,
+            groups=dict(groups),
+            person_properties=person_properties,
+            group_properties=group_properties,
+            flag_keys_to_evaluate=flag_keys,
+        )
```
dustinbyrne, line 2375: We should pass `device_id` here, as it may be present in the context (via tracing headers).

I don't think it's necessary to add it as a parameter of `evaluate_flags` as well, but I'll leave it up to you.

posthog/feature_flag_evaluations.py
```diff
@@ -0,0 +1,254 @@
...
+                properties["$feature_flag_version"] = flag.version
+            if flag.reason:
+                properties["$feature_flag_reason"] = flag.reason
+
+        if self._request_id:
+            properties["$feature_flag_request_id"] = self._request_id
+        if self._evaluated_at and not (flag and flag.locally_evaluated):
+            properties["$feature_flag_evaluated_at"] = self._evaluated_at
+        if flag is None:
+            properties["$feature_flag_error"] = "flag_missing"
```
dustinbyrne: Just noting that we're losing granularity of `$feature_flag_error` values here

dustinbyrne (commented):
Looks good! I have a few questions/considerations

.sampo/changesets/evaluate-flags-api.md
```diff
@@ -0,0 +1,16 @@
+---
+"posthog": minor
```
marandaneto: this is wrong, see https://github.com/PostHog/posthog-python/blob/main/RELEASING.md

posthog/__init__.py
```diff
@@ -770,6 +773,59 @@ def get_all_flags_and_payloads(
...
+    Args:
+        distinct_id: The user's distinct ID. If ``None``, falls back to the context
+            distinct_id. If still unresolvable, returns an empty snapshot.
+        groups: Mapping of group type to group key.
+        person_properties: Person properties to use for evaluation.
+        group_properties: Group properties keyed by group type.
+        only_evaluate_locally: If ``True``, never fall back to remote evaluation.
+        disable_geoip: Whether to disable GeoIP lookup.
+        flag_keys: Optional list of flag keys to scope the underlying ``/flags``
+            request to a subset.
```
marandaneto: its unclear what this does to me

marandaneto:
how does auto captured events and errors set `flags` to those events if you need the instance from `evaluate_flags`

posthog/client.py
```diff
@@ -631,6 +637,7 @@ def capture(
         timestamp = kwargs.get("timestamp", None)
         uuid = kwargs.get("uuid", None)
         groups = kwargs.get("groups", None)
+        flags_snapshot = kwargs.get("flags", None)
```
marandaneto, line 639: if `send_feature_flags` is false, do we ignore `flags_snapshot`?

marandaneto:
> No breaking changes. Existing feature_enabled, get_feature_flag, get_feature_flag_payload, and capture(send_feature_flags=...) paths all continue to work unchanged. Phase 2 (runtime DeprecationWarnings pointing at evaluate_flags) and Phase 3 (removal in next major) ship in follow-up PRs — this minor is API-additive only so users have a quiet release to migrate before warnings start.

i'd say we should deprecate and add a warning now otherwise theres multiple ways of doing the same thing and users and agents will be very confused
theres an RFC about that already

.sampo/changesets/evaluate-flags-api.md
dmarticus: Thanks — fixed in 95eb1e9. The header now reads `pypi/posthog: minor` per RELEASING.md (which points at https://github.com/bruits/sampo as the format spec).

posthog/feature_flag_evaluations.py
dmarticus: Good catch — fixed in 95eb1e9. `only_accessed()` now honors its name and returns an empty snapshot when nothing has been accessed (no fallback, no warning). The previous behavior was a misguided safety net that ended up being more surprising than helpful for exactly the scenario you describe (early `capture()` before any branching). The Node-side equivalent has the same change queued.

Updated test: `test_only_accessed_returns_empty_when_no_flags_accessed`.

posthog/client.py
dmarticus: Reversed course on this — shipped Phase 2 in eda573d alongside Phase 1 in this PR (rather than splitting it into a follow-up).

`feature_enabled`, `get_feature_flag`, `get_feature_flag_payload`, and `capture(send_feature_flags=...)` now emit `DeprecationWarning`s pointing at `evaluate_flags()`. The methods continue to return the same values; users who pin warnings to errors will get a heads-up on first use, and the rest see them surface via pytest / `python -W` / IDE inspectors.

`feature_enabled` and `get_feature_flag` were restructured to call `_get_feature_flag_result` directly so a single user-level call emits exactly one warning instead of cascading three.

Phase 3 (removal in next major) is separate.

posthog/client.py
dmarticus: Good catch — wired up in 95eb1e9. `evaluate_flags()` now resolves `device_id` from context (`get_context_device_id()`) at the top of the method, then forwards it to the `get_flags_decision(...)` call so it lands in the `/flags` request body. I went with your suggestion and didn't add it as a method parameter — context-only is cleaner and matches the existing `distinct_id` resolution pattern.

posthog/feature_flag_evaluations.py
dmarticus: You're right — fixed in 95eb1e9. The snapshot now tracks response-level errors (`errors_while_computing_flags`, `quota_limited`) at construction and `_record_access` builds a comma-joined `$feature_flag_error` matching the single-flag path's granularity. So a missing flag during a quota-limited response now reports `quota_limited,flag_missing` instead of just `flag_missing`.

New test: `test_errors_while_computing_flags_propagates_to_event` covers both the standalone (`errors_while_computing_flags`) and combined (`errors_while_computing_flags,flag_missing`) cases.

Not covered yet: `TIMEOUT`/`CONNECTION_ERROR`/`api_error_NNN`. Those manifest as exceptions on the request and currently just log + leave the snapshot empty (so flags would surface as `flag_missing`). Could fold those into the snapshot in a follow-up if useful, but the most common cases (`errors_while_computing` and `quota_limited`) are now covered.

posthog/__init__.py
dmarticus: Sorry, the docstring was too terse. Rewrote it in 95eb1e9:

> `flag_keys`: Optional list of flag keys. When provided, only these flags are evaluated — the underlying `/flags` request asks the server for just this subset, which makes the response smaller and the request cheaper. Use this when you only need a handful of flags out of many.

This lives next to (and is distinct from) the in-memory `only([keys])` filter helper on the returned snapshot — `flag_keys` trims the network call, `only([...])` trims which flags get attached to a captured event.

posthog/client.py
dmarticus: No — `flags` always wins, regardless of `send_feature_flags`. Made the precedence explicit in 95eb1e9 with a comment block plus a runtime warning when both are passed:

```python
# Precedence: an explicit `flags` snapshot always wins, regardless of
# `send_feature_flags`. The snapshot guarantees the event carries the same
# values the developer branched on with no additional network call. The
# `send_feature_flags` path only runs when no snapshot is provided.
if flags_snapshot is not None:
    if send_feature_flags:
        self.log.warning(
            "[FEATURE FLAGS] Both `flags` and `send_feature_flags` were passed to "
            "capture(); using `flags` and ignoring `send_feature_flags`."
        )
    extra_properties = flags_snapshot._get_event_properties()
else:
    ...
```

New test: `test_capture_warns_and_uses_flags_when_both_flags_and_send_feature_flags_set` asserts the warning fires and no extra `/flags` request is made.

dmarticus:
> how does auto captured events and errors set `flags` to those events if you need the instance from `evaluate_flags`


Good question — the answer has two layers.

**Manual error captures**: `capture_exception(exception, ..., flags=flags)` now forwards the snapshot through to the inner `capture()` call so the `$exception` event carries the same flag context as your other events. This was a small gap I missed — fixed in db6d3c0:

```python
flags = posthog.evaluate_flags(distinct_id)
try:
    risky_thing()
except Exception as e:
    posthog.capture_exception(e, distinct_id=distinct_id, flags=flags)
```

The same applies to anywhere the developer is the one calling `capture()` — they can thread `flags=` through.

**Auto-captured exceptions** (the case you're really asking about — `sys.excepthook` in `ExceptionCapture`, plus the `new_context()` block-level exception handler): you're right that there's no surface today for the developer to attach a flag snapshot, because they're not the one calling capture. Two paths I see for a follow-up PR:

1. **Stash flags on the context.** `new_context()` already carries distinct_id, session_id, properties — adding `flags` would mean any exception captured within the block (block-level handler, or even `sys.excepthook` if the context is active) would automatically attach the snapshot. Most ergonomic for the request-scoped pattern.

2. **A "default flags provider" callback on the client** that auto-evaluates at capture time. More magic, more risk of latency surprises. Probably not the right answer.

Happy to do (1) as a separate PR — felt out of scope for this one since it touches the context module and needs its own design pass. Worth tracking as a follow-up issue?

dmarticus:
> > No breaking changes. Existing feature_enabled, get_feature_flag, get_feature_flag_payload, and capture(send_feature_flags=...) paths all continue to work unchanged. Phase 2 (runtime DeprecationWarnings pointing at evaluate_flags) and Phase 3 (removal in next major) ship in follow-up PRs — this minor is API-additive only so users have a quiet release to migrate before warnings start.
> 
> i'd say we should deprecate and add a warning now otherwise theres multiple ways of doing the same thing and users and agents will be very confused theres an RFC about that already


You're right — reversed course on this and shipped Phase 2 in eda573d alongside Phase 1 in this PR. The PR description has been updated.

`feature_enabled`, `get_feature_flag`, `get_feature_flag_payload`, and `capture(send_feature_flags=...)` now emit `DeprecationWarning`s pointing at `evaluate_flags()`. They keep working unchanged in this minor — Phase 3 (removal) lands in the next major.

`feature_enabled` and `get_feature_flag` were restructured to call `_get_feature_flag_result` directly so a single user-level call emits exactly one warning instead of cascading three.

.sampo/changesets/evaluate-flags-api.md
```diff
@@ -0,0 +1,16 @@
...
+```python
+flags = posthog.evaluate_flags(distinct_id, person_properties={"plan": "enterprise"})
+if flags.is_enabled("new-dashboard"):
+    render_new_dashboard()
+posthog.capture("page_viewed", distinct_id=distinct_id, flags=flags)
+```
+
+The returned `FeatureFlagEvaluations` snapshot exposes `is_enabled()`, `get_flag()`, `get_flag_payload()` for branching and `only_accessed()` / `only([keys])` filter helpers. Pass `flag_keys=[...]` to `evaluate_flags()` to scope the underlying `/flags` request itself.
+
+Deprecates `feature_enabled()`, `get_feature_flag()`, `get_feature_flag_payload()`, and `capture(send_feature_flags=...)`. They continue to work but now emit a `DeprecationWarning` pointing at `evaluate_flags()`. Removal is planned for the next major version.
```
dustinbyrne, line 16: Let's deprecate `get_feature_flag_result` as well. It lived a short life, but better to clean it up now.

marandaneto:
@dustinbyrne you'll need to coordinate this with the @PostHog/team-docs-wizard as well otherwise the wizard will be lost or instrumenting new apps with deprecated APIs
we should get the docs here https://posthog.com/docs/references/posthog-python?filter=feature-flags and here https://posthog.com/docs/libraries/python updated and most likely all the other code snippets spread around the codebase
can we also write a skill that we can add to https://github.com/PostHog/skills/tree/main/skills and point out in our docs so users can just run that skill in that codebase for the migration?
this is not a blocker for this PR but ideally somehow synced otherwise we'll just get support tickets I guess

marandaneto:
I think 'Stash flags on the context' makes sense
if we do this on a follow up PR, does that mean that new exceptions wont have flags context as they did before until we address this issue? or they currently dont have anyway so its not an issue?

dustinbyrne:
> @dustinbyrne you'll need to coordinate this with the @PostHog/team-docs-wizard as well otherwise the wizard will be lost or instrumenting new apps with deprecated APIs

Edwin gave me a short into on how the wizard sources context for these types of things. As I understand it, as long as it's documented on the website, the wizard will automatically start doing the right thing.

See https://github.com/PostHog/context-mill/blob/main/transformation-config/skills/feature-flags/config.yaml

For Python context it sources:
- https://posthog.com/docs/feature-flags/adding-feature-flag-code.md
- https://posthog.com/docs/feature-flags/best-practices.md
- https://posthog.com/docs/feature-flags/installation/python.md

So these are the documents that we will need to update. Same for every other language.

gewenyu99:
Updating the docs should do it. There isn't a feature flag specific thing that we run during setup, so it should be safe to get this in. There is a best practices skill, and I think this might change some of the feature flag best practices?

marandaneto (approved):

dustinbyrne (approved):
I can spend some time tomorrow updating documentation for this. If you update the other SDK branches to match this spec I'll review those as well! 👍

.sampo/changesets/evaluate-flags-api.md
dmarticus: done!

posthog/__init__.py
```diff
@@ -770,6 +773,61 @@ def get_all_flags_and_payloads(
...
 
+def evaluate_flags(
+    distinct_id=None,  # type: Optional[str]
+    groups=None,  # type: Optional[Dict[str, str]]
+    person_properties=None,  # type: Optional[Dict[str, Any]]
+    group_properties=None,  # type: Optional[Dict[str, Dict[str, Any]]]
+    only_evaluate_locally=False,  # type: bool
+    disable_geoip=None,  # type: Optional[bool]
+    flag_keys=None,  # type: Optional[list]
+) -> FeatureFlagEvaluations:
```
dustinbyrne, line 785: surfaced when documenting: we should include a `device_id` override for parity with the deprecated methods

this might be relevant for other sdks as well?

dmarticus: good shout! So most of the other SDKs didn't actually support this yet – it was only this one and JS that supported that operation. I _do_ think we should support `device_id` in this method going forward for other SDKs, but given that only Python and JS have it now, it's only a concern for those SDKs. Definitely will update JS, though.

Issue #173, opened 2025-01-23
posthog.project_api_key should be able to be None if (posthog.disabled == True)


PeterStolz:
Hi,
I recently integrated posthog into our backend and I do not like that a `posthog.capture` can crash if the api key is not set even when I set `posthog.disabled = True`.

```
Traceback (most recent call last):
  File "test.py", line 8, in <module>
    posthog.capture("test", event="asd")
  File "venv/lib/python3.12/site-packages/posthog/__init__.py", line 67, in capture
    return _proxy(
           ^^^^^^^
  File "venv/lib/python3.12/site-packages/posthog/__init__.py", line 497, in _proxy
    default_client = Client(
                     ^^^^^^^
  File "venv/lib/python3.12/site-packages/posthog/client.py", line 69, in __init__
    require("api_key", self.api_key, string_types)
  File "venv/lib/python3.12/site-packages/posthog/client.py", line 941, in require
    raise AssertionError(msg)
AssertionError: api_key must have (<class 'str'>,), got: None
```

To me this feels like a bug.
You can reproduce with
```
import posthog

posthog.disabled = True
posthog.host = "https://eu.i.posthog.com"
posthog.capture("test", event="asd")
```
It happens on posthog==3.8.4 and posthog==3.9.3
If you agree that this should not be default behavior when posthog.disabled is set to True I can create a PR

PR #541, opened 2026-04-29
fix: address straightforward open issue bugs

### LINKED ISSUE #526: Misleading error message on local evaluation 401: always blames personal_api_key even when project_api_key is the problem
## Summary

When local flag evaluation fails with a 401, the SDK logs a hardcoded message that always attributes the failure to \`personal_api_key\`, even when the actual cause is an invalid or rotated \`project_api_key\`.

## Affected code

**\`posthog/client.py:1352–1356\`** — the 401 handler in \`_fetch_feature_flags_from_api()\`:

\`\`\`python
except APIError as e:
    if e.status == 401:
        self.log.error(
            "[FEATURE FLAGS] Error loading feature flags: To use feature flags, "
            "please set a valid personal_api_key. More information: https://posthog.com/docs/api/overview"
        )
\`\`\`

## Why this is wrong

The local evaluation endpoint requires **two** keys:

\`\`\`
GET /api/feature_flag/local_evaluation/?token={project_api_key}&send_cohorts
Authorization: Bearer {personal_api_key}
\`\`\`

(\`request.py:344\` for the Bearer header; \`client.py:1295\` for the URL with \`token=\`)

If **either** key is invalid, the server returns 401. The SDK unconditionally tells the user to fix \`personal_api_key\`.

## Real-world impact

A customer (Zendesk #54814) rotated their \`project_api_key\` via the PostHog UI. Their local evaluation immediately started failing with 401. The error message told them their \`personal_api_key\` was invalid, so they regenerated their personal key and their feature flags secure API key — neither of which helped. The real fix was deploying the new \`project_api_key\`, which they eventually discovered independently. The misleading error cost significant debugging time.

## The server already distinguishes between the two cases

The backend returns **different \`detail\` fields** depending on which key failed:

- **Invalid \`?token=\` (project API key):** \`routing.py:462\` raises \`AuthenticationFailed()\` with no message → DRF default: \`"Incorrect authentication credentials."\`
- **Invalid personal API key (Bearer header):** \`auth.py:260\` raises \`AuthenticationFailed("Personal API key found in request Authorization header is invalid.")\`

The SDK already captures this in \`_process_response()\` (\`request.py:269\`):

\`\`\`python
raise APIError(res.status_code, payload["detail"], retry_after=retry_after)
\`\`\`

So \`e.message\` in the \`except APIError\` block already contains the server's distinguishing message. The current fix ignores it entirely.

## Suggested fix

Surface \`e.message\` from the server response rather than replacing it with a hardcoded string:

\`\`\`python
if e.status == 401:
    self.log.error(
        f"[FEATURE FLAGS] Error loading feature flags: authentication failed (HTTP 401): {e.message}. "
        "Check that both your project API key and your personal API key "
        "(or Feature Flags secure API key) are valid. "
        "More information: https://posthog.com/docs/api/overview"
    )
\`\`\`

This surfaces the server's specific message (which already identifies which key failed) while keeping the fallback guidance that covers both keys.

## Cross-SDK status

All server-side SDKs that support local evaluation use the same two-key pattern. Error messaging quality varies:

| SDK | 401 message | Issue |
|-----|-------------|-------|
| **posthog-python** | "please set a valid personal_api_key" | Worst — hardcoded, unambiguously wrong |
| **posthog-node** | "Your project key or personal API key is invalid" | Ambiguous but at least mentions both keys |
| **posthog-ruby** | Debug-level log only: \`"Failed to load feature flags: #{res}"\` | Silent — no user-visible error at all |
| **posthog-go** | "Unable to fetch feature flags, status: 401" | Generic — no mention of which key |
| **posthog-php** | Passes server's \`detail\` field, or silent | Depends on server response format |
| **posthog-dotnet** | Generic "Unauthorized" | Least specific |
| **posthog-elixir** | N/A — no local evaluation support | — |

Python is the most egregious (actively wrong), but most SDKs could be clearer.

marandaneto:
Fixes #526, #308, and #173.

This PR addresses the severe, straightforward bug reports from the open issue triage:
- local feature flag 401 errors now include the server error and mention both `project_api_key` and `personal_api_key` instead of always blaming only `personal_api_key`
- oversized queued events are acknowledged when dropped so `flush()` does not hang forever
- disabled/no-op usage no longer requires an API key before returning early

New feature requests and more invasive fixes were intentionally left out.

- `uv run pytest posthog/test/test_consumer.py posthog/test/test_client.py posthog/test/test_feature_flags.py posthog/test/test_module.py -q`
- `uv run ruff check .`

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/.sampo/changesets/quiet-otter-fixes.md
+++ b/.sampo/changesets/quiet-otter-fixes.md
@@ -1,0 +1,5 @@
+---
+pypi/posthog: patch
+---
+
+Fix disabled no-op setup, oversized queue drops, and local feature flag auth errors.
--- a/posthog/__init__.py
+++ b/posthog/__init__.py
@@ -857,7 +857,7 @@ def shutdown():
 def setup() -> Client:
     global default_client
     if not default_client:
-        if not api_key:
+        if not api_key and not disabled:
             raise ValueError("API key is required")
         default_client = Client(
             api_key,
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -166,7 +166,7 @@ class Client(object):
 
     def __init__(
         self,
-        project_api_key: str,
+        project_api_key: Optional[str],
         host=None,
         debug=False,
         max_queue_size=10000,
@@ -220,7 +220,12 @@ class Client(object):
         self.queue = queue.Queue(max_queue_size)
 
         # api_key: This should be the Team API Key (token), public
-        self.api_key = project_api_key.strip()
+        if project_api_key is None:
+            if not disabled:
+                raise ValueError("API key is required")
+            self.api_key = ""
+        else:
+            self.api_key = project_api_key.strip()
 
         self.on_error = on_error
         self.debug = debug
@@ -290,7 +295,7 @@ class Client(object):
         else:
             self.log.setLevel(logging.WARNING)
 
-        if not self.api_key:
+        if not self.api_key and not self.disabled:
             self.log.error(
                 "api_key is empty after trimming whitespace; check your project API key"
             )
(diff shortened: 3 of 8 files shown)


marandaneto:
Superseded by split draft PRs: #542, #543, and #544.

PR #544, opened 2026-04-30
fix: no-op clients without api key

### LINKED ISSUE #173: posthog.project_api_key should be able to be None if (posthog.disabled == True)

marandaneto:
Fixes #173.

When a direct client is created with a project API key that is missing or trims to an empty string, the SDK now logs the empty-key warning and marks that client as disabled so calls are no-ops. Feature flag loading and direct flag decision APIs also return early for disabled clients to avoid malformed requests with an empty token.

Module-level setup now preserves the client’s auto-disabled state after trimming a whitespace-only key.

- `uv run pytest posthog/test/test_client.py posthog/test/test_module.py -q`
- `uv run ruff format --check posthog/__init__.py posthog/client.py posthog/test/test_client.py posthog/test/test_module.py`
- `uv run ruff check posthog/__init__.py posthog/client.py posthog/test/test_client.py posthog/test/test_module.py`

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -220,7 +220,7 @@ class Client(object):
         self.queue = queue.Queue(max_queue_size)
 
         # api_key: This should be the Team API Key (token), public
-        self.api_key = project_api_key.strip()
+        self.api_key = (project_api_key or "").strip()
 
         self.on_error = on_error
         self.debug = debug
@@ -245,7 +245,7 @@ class Client(object):
         self.flag_definition_version = 0
         self._flags_etag: Optional[str] = None
         self._flag_definition_cache_provider = flag_definition_cache_provider
-        self.disabled = disabled
+        self.disabled = disabled or not self.api_key
         self.disable_geoip = disable_geoip
         self.historical_migration = historical_migration
         self.super_properties = super_properties
@@ -538,6 +538,9 @@ class Client(object):
         Category:
             Feature flags
         """
+        if self.disabled:
+            return normalize_flags_response({})
+
         groups = groups or {}
         person_properties = person_properties or {}
         group_properties = group_properties or {}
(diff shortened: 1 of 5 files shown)


posthog/client.py
```diff
@@ -245,7 +245,7 @@ def __init__(
         self.flag_definition_version = 0
         self._flags_etag: Optional[str] = None
         self._flag_definition_cache_provider = flag_definition_cache_provider
-        self.disabled = disabled
+        self.disabled = disabled or not self.api_key
```
ioannisj, line 248: Looks like get_feature_flag family is gated by `disabled` but not load_feature_flags() call, which can make a malformed net call with an empty api_key?

ioannisj (approved):
LG, left a minor comment on disabled checks.

posthog/client.py
marandaneto: good point

PR #546, opened 2026-04-30
chore: commit generated references with GitHub App

marandaneto:

The generated references workflow was failing when trying to push directly to `main` with `github-actions[bot]`, because repository rules require verified commits and PR/ruleset checks.

References are only needed after releases, so this removes the standalone manual `generate-references.yml` workflow and generates/commits references inline in the already-approved release job. The commit now uses the same GitHub App token and `planetscale/ghcommit-action` flow used by the release workflow, avoiding a second environment approval.

- Parsed `.github/workflows/release.yml` as YAML locally.
- Ran `git diff --check`.

## :pencil: Checklist

### If releasing new changes

- [ ] Added the `release` label to the PR


.github/workflows/release.yml
```diff
@@ -195,11 +194,34 @@ jobs:
...
+        uses: planetscale/ghcommit-action@25309d8005ac7c3bcd61d3fe19b69e0fe47dbdde # v0.2.20
+        with:
+          commit_message: "Update generated references"
+          repo: ${{ github.repository }}
+          branch: main
+          file_pattern: references/
         env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: gh workflow run generate-references.yml --ref main
+          GITHUB_TOKEN: ${{ steps.releaser.outputs.token }}
```
marandaneto, line 223: we need this to live here because of the GH token, otherwise it'd require a 2nd environment approval

marandaneto:
https://github.com/PostHog/posthog-python/actions/runs/25156855061/job/73740806572

.github/workflows/release.yml
```diff
@@ -195,11 +194,34 @@ jobs:
           NEW_VERSION: ${{ steps.sampo-release.outputs.new_version }}
         run: gh release create "$NEW_VERSION" --generate-notes
 
-      - name: Dispatch generate-references
+      - name: Generate references
         if: steps.commit-release.outputs.commit-hash != ''
+        run: |
```
ioannisj, line 198: Do we need a continue on error for these next steps? Low context from me here but if the package is released and externally visible I think we should not fail. Plus previous impl was just dispatching to another workflow so any failures were isolated there and did not affect this one

marandaneto: i dont disagree but this is exactly what happened before
generate references was failing and we didnt notice for quite a while

ioannisj: Could this be a warning message on slack?

marandaneto: since this is executed after release, github release etc the release will work anyway, and if references failed, we can just fix forward

marandaneto: yes, its gonna be a failed release, so the slack thread gets updated, we know that it partly failed, fix forward and thats it

ioannisj (approved):
Approving anw to unblock

.github/workflows/release.yml
marandaneto: we can change if this often fails and its breaking releases
we tried the other way, we had this broken for a few weeks, now we try the other way

PR #575, opened 2026-05-14
fix: guard null token usage fields in OpenAI converter

### LINKED ISSUE #574: TypeError in OpenAI converter when token usage detail fields are null
## Bug

`extract_openai_usage_from_response` in `posthog/ai/openai/openai_converter.py` crashes with:

```
TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

when the OpenAI API response includes `prompt_tokens_details.cached_tokens = null` or `completion_tokens_details.reasoning_tokens = null`.

## Root Cause

The `hasattr()` checks at lines 410-417 pass because the attributes exist on the response object — they're just `None`. The subsequent comparisons at lines 424/426 then fail:

```python
# Line 410-413: hasattr passes because attribute exists with value None
if hasattr(response.usage, "prompt_tokens_details") and hasattr(
    response.usage.prompt_tokens_details, "cached_tokens"
):
    cached_tokens = response.usage.prompt_tokens_details.cached_tokens  # <- None

# Line 424: crashes
if cached_tokens > 0:  # TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

Same issue exists for `reasoning_tokens` (line 426) and likely `web_search_count` (line 430).

## Reproduction

Any OpenAI-compatible API that returns a response with:
```json
{
  "usage": {
    "prompt_tokens": 500,
    "completion_tokens": 200,
    "prompt_tokens_details": { "cached_tokens": null },
    "completion_tokens_details": { "reasoning_tokens": null }
  }
}
```

This is common with OpenAI-compatible proxies/gateways and certain model providers that include the detail objects but set individual fields to null when not applicable.

## Suggested Fix

```python
# Line 424-427:
if cached_tokens is not None and cached_tokens > 0:
    result["cache_read_input_tokens"] = cached_tokens
if reasoning_tokens is not None and reasoning_tokens > 0:
    result["reasoning_tokens"] = reasoning_tokens
```

And similarly for `web_search_count` at line 430.

## Environment

- `posthog-python` version: latest (installed via pip)
- Python: 3.13
- OpenAI SDK: 2.17.0
- Observed via `posthog.ai.openai.AsyncOpenAI` wrapper around an OpenAI-compatible gateway

michael-ciridae:
## Problem

`extract_openai_usage_from_response` crashes with:
```
TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

when the OpenAI API response includes `prompt_tokens_details.cached_tokens = null` or `completion_tokens_details.reasoning_tokens = null`.

The `hasattr()` checks pass because the attributes exist on the response object — they're just `None`. The subsequent `> 0` comparisons then fail.

This is common with OpenAI-compatible proxies/gateways that include the token detail objects but set individual fields to null when not applicable (e.g. no caching, no reasoning tokens for a given model).

## Fix

- Add `is not None` guards before `> 0` comparisons in `extract_openai_usage_from_response`
- Add `is not None` guards before assigning to the usage dict in both streaming paths (chat completions and responses API)

## Impact

Without this fix, any `chat.completions.create()` or `.parse()` call that returns null token detail fields crashes the PostHog wrapper, which propagates up and kills the caller's request — even though the LLM response itself was valid.

Fixes #574

### DIFF
--- a/.sampo/changesets/noble-thunderbearer-vellamo.md
+++ b/.sampo/changesets/noble-thunderbearer-vellamo.md
@@ -1,0 +1,5 @@
+---
+pypi/posthog: patch
+---
+
+Fix OpenAI usage parsing when token detail fields are null
--- a/posthog/ai/openai/openai_converter.py
+++ b/posthog/ai/openai/openai_converter.py
@@ -432,9 +432,9 @@ def extract_openai_usage_from_response(response: Any) -> TokenUsage:
         output_tokens=output_tokens,
     )
 
-    if cached_tokens > 0:
+    if cached_tokens is not None and cached_tokens > 0:
         result["cache_read_input_tokens"] = cached_tokens
-    if reasoning_tokens > 0:
+    if reasoning_tokens is not None and reasoning_tokens > 0:
         result["reasoning_tokens"] = reasoning_tokens
 
     web_search_count = extract_openai_web_search_count(response)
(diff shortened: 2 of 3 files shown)


gregciridae:
Just ran into this issue again, thanks Michael :)

marandaneto:
@PostHog/team-llm-analytics is this safe to go?

marandaneto (approved):

marandaneto:
i pushed the changeset entry and a test, will release it

marandaneto:
force pushed commits so they are all signed, its a blocker for merging

PR #601, opened 2026-05-20
fix: reject leading-zero semver values in local evaluation

dmarticus:
## Summary

Per [semver 2.0.0 §2](https://semver.org/#spec-item-2), numeric identifiers MUST NOT include leading zeros. Values like `1.07.3` or `01.2.3` are not valid semver — the local feature flag evaluator currently parses them silently (via `int("07") → 7`), which means a person property of `1.07.3` would incorrectly match a `semver_eq` condition against `1.7.3`.

This PR makes the parser reject leading zeros in numeric identifiers. Both override values and flag values are validated; invalid inputs raise `InconclusiveMatchError` so the condition simply does not match.

## Changes

- `parse_semver` now uses a `_semver_numeric_identifier` helper that rejects strings with leading zeros (except literal `"0"`).
- `_wildcard_bounds` uses the same helper, so `01.*` and `1.07.*` are rejected.
- Updated the existing edge-case test that previously asserted `01.02.03` matched `1.2.3` — it now correctly rejects.
- Added `test_match_properties_semver_rejects_leading_zeros` covering override values, flag values, and all range operators.

## Test plan

- [x] `pytest posthog/test/test_feature_flags.py` — 128 passed
- [x] New rejection tests pass
- [x] `ruff format` clean
- [x] No new mypy errors on `feature_flags.py`

### DIFF
--- a/.sampo/changesets/strict-semver-leading-zeros.md
+++ b/.sampo/changesets/strict-semver-leading-zeros.md
@@ -1,0 +1,5 @@
+---
+pypi/posthog: patch
+---
+
+Reject semver values with leading zeros in local flag evaluation. Per semver 2.0.0 §2, numeric identifiers must not include leading zeros — values like `1.07.3` are not valid semver and should not match targeting conditions. Both override values and flag values are now validated; invalid inputs raise `InconclusiveMatchError` so the condition does not match.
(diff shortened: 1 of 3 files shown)


marandaneto:
missing changelog entry

marandaneto (approved):

PR #622, opened 2026-05-22
fix(ai): wrap streaming responses in AsyncStreamWrapper to support async with

### LINKED ISSUE #393: PostHog with pydantic-ai streaming raises TypeError: 'async_generator'
## Description

When using `posthog` with pydantic-ai's streaming functionality, a `TypeError` is raised because the response object from PostHog's wrapped client doesn't support the asynchronous context manager protocol that pydantic-ai expects.

Have a repo to reproduce the issues here:
https://github.com/noelw-a11y/posthog-pydantic/tree/main

We have observed this issue for other providers such as OpenAI and Anthropic too.

## Environment

- **Python version**: 3.12.2
- **posthog version**: >=7.3.1
- **openai version**: >=2.13.0
- **pydantic-ai version**: >=1.33.0

## Steps to Reproduce

In README.md:
https://github.com/noelw-a11y/posthog-pydantic/tree/main

## Expected Behavior

The streaming should work correctly, allowing pydantic-ai to iterate over the streaming response chunks.

## Actual Behavior

A `TypeError` is raised:

```
TypeError: 'async_generator' object does not support the asynchronous context manager protocol
```

## Error Traceback

```
Traceback (most recent call last):
  File "main.py", line 182, in <module>
    main()
  File "main.py", line 175, in main
    asyncio.run(test_streaming())
  File "/Users/noelwilson/.pyenv/versions/3.12.2/lib/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/Users/noelwilson/.pyenv/versions/3.12.2/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/noelwilson/.pyenv/versions/3.12.2/lib/python3.12/asyncio/base_events.py", line 685, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "main.py", line 79, in test_streaming
    async with agent.run_stream('Write a short 3-sentence story about a robot learning to code.') as result:
  File "/Users/noelwilson/.pyenv/versions/3.12.2/lib/python3.12/contextlib.py", line 210, in __aenter__
    return await anext(self.gen)
           ^^^^^^^^^^^^^^^^^^^^^
  File ".../pydantic_ai/agent/abstract.py", line 525, in run_stream
    async with node.stream(graph_ctx) as stream:
  File "/Users/noelwilson/.pyenv/versions/3.12.2/lib/python3.12/contextlib.py", line 210, in __aenter__
    return await anext(self.gen)
           ^^^^^^^^^^^^^^^^^^^^^
  File ".../pydantic_ai/_agent_graph.py", line 450, in stream
    async with ctx.deps.model.request_stream(
  File "/Users/noelwilson/.pyenv/versions/3.12.2/lib/python3.12/contextlib.py", line 210, in __aenter__
    return await anext(self.gen)
           ^^^^^^^^^^^^^^^^^^^^^
  File ".../pydantic_ai/models/openai.py", line 496, in request_stream
    async with response:
TypeError: 'async_generator' object does not support the asynchronous context manager protocol
```

## Root Cause Analysis

The issue occurs at `pydantic_ai/models/openai.py:496` where the code attempts to use:
```python
async with response:
```

However, when using `posthog.ai.openai.AsyncOpenAI`, the `response` object is an async generator that doesn't implement `__aenter__` and `__aexit__` methods required for the async context manager protocol.

## Workaround

Using the standard `openai.AsyncOpenAI` client (without PostHog) works correctly. The issue only occurs when PostHog's wrapper is used.

## Additional Context

- Non-streaming calls (using `agent.run()`) appear to work correctly with PostHog
- The issue is specific to streaming functionality (`agent.run_stream()`)
- A minimal reproduction repository is available at: [link to repo if public]

## Related Code Location

The error originates from:
- `pydantic_ai/models/openai.py`, line 496: `async with response:`
- The `response` comes from PostHog's wrapped `AsyncOpenAI` client's streaming method

devteamaegis:
## Problem

When using the PostHog AI wrappers (`posthog.ai.openai.AsyncOpenAI`,
`posthog.ai.anthropic.AsyncAnthropic`) with libraries that expect streaming
responses to be **async context managers** (e.g. pydantic-ai, any code that
uses `async with response:`), a `TypeError` is raised immediately:

```
TypeError: 'async_generator' object does not support the
asynchronous context manager protocol
```

Root cause: both `WrappedCompletions._create_streaming` and
`WrappedResponses._create_streaming` (OpenAI), and
`AsyncMessages._create_streaming` (Anthropic), return a bare
`async_generator()` object.  Async generators implement `__aiter__` /
`__anext__` but **not** `__aenter__` / `__aexit__`.

The OpenAI and Anthropic SDKs return `AsyncStream` objects that implement
both protocols. The PostHog wrappers lost that compatibility.

Fixes #393.

## Fix

Add `posthog/ai/stream.py` with a lightweight `AsyncStreamWrapper` class
that wraps an async generator and adds the missing context-manager protocol:

```python
async with response as stream:   # now works
    async for chunk in stream:
        ...
```

Key behaviors:
- `__aenter__` returns `self` so `async with response as r: async for c in r:` works as expected.
- `__aexit__` calls `aclose()` on the underlying generator, ensuring the `finally` block (PostHog event capture) runs even when the caller breaks out early.
- `__getattr__` proxies attribute access to the underlying generator for provider-specific metadata.
- Backward compatible: code using `async for chunk in response:` directly continues to work unchanged.

## Changes

| File | Change |
|------|--------|
| `posthog/ai/stream.py` | New — `AsyncStreamWrapper` |
| `posthog/ai/openai/openai_async.py` | Both `_create_streaming` methods return `AsyncStreamWrapper(async_generator())` |
| `posthog/ai/anthropic/anthropic_async.py` | `_create_streaming` returns `AsyncStreamWrapper(generator())` |
| `posthog/test/test_async_stream_wrapper.py` | 6 regression tests (no mocks, no external deps) |

## Tests

```
pytest posthog/test/test_async_stream_wrapper.py -v
# 6 passed
```

### DIFF
--- a/posthog/ai/anthropic/anthropic_async.py
+++ b/posthog/ai/anthropic/anthropic_async.py
@@ -8,6 +8,7 @@ except ImportError:
 
 import time
 import uuid
+from posthog.ai.stream import AsyncStreamWrapper
 from typing import Any, Dict, List, Optional
 
 from posthog import setup
@@ -225,7 +226,7 @@ class AsyncWrappedMessages(AsyncMessages):
                     stop_reason=stop_reason,
                 )
 
-        return generator()
+        return AsyncStreamWrapper(generator())
 
     async def _capture_streaming_event(
         self,
--- a/posthog/ai/openai/openai_async.py
+++ b/posthog/ai/openai/openai_async.py
@@ -1,8 +1,9 @@
 import time
 import uuid
-from typing import Any, Dict, List, Optional
+from typing import Any, AsyncIterator, Dict, List, Optional
 
 from posthog.ai.types import TokenUsage
+from posthog.ai.stream import AsyncStreamWrapper
 
 try:
     import openai
@@ -206,7 +207,7 @@ class WrappedResponses:
                     stop_reason=stop_reason,
                 )
 
-        return async_generator()
+        return AsyncStreamWrapper(async_generator())
 
     async def _capture_streaming_event(
         self,
@@ -486,7 +487,7 @@ class WrappedCompletions:
                     stop_reason=stop_reason,
                 )
 
-        return async_generator()
+        return AsyncStreamWrapper(async_generator())
 
     async def _capture_streaming_event(
         self,
--- a/posthog/ai/stream.py
+++ b/posthog/ai/stream.py
@@ -1,0 +1,61 @@
+"""Shared async streaming utilities for PostHog AI wrappers."""
+
+from typing import Any, AsyncGenerator, TypeVar
+
+T = TypeVar("T")
+
+
+class AsyncStreamWrapper:
+    """Wraps an async generator so it also implements the async context manager protocol.
+
+    The OpenAI and Anthropic SDKs return stream objects that support both
+    ``async for`` iteration **and** ``async with`` (i.e. they are both async
+    iterators and async context managers).  PostHog's streaming wrappers
+    previously returned a bare async generator, which only supports ``async
+    for``.  Libraries such as pydantic-ai call ``async with response:`` before
+    iterating, causing::
+
+        TypeError: 'async_generator' object does not support the
+        asynchronous context manager protocol
+
+    This class wraps the underlying async generator and adds the missing
+    ``__aenter__`` / ``__aexit__`` methods.  On ``__aexit__`` the generator is
+    closed so that the ``finally`` block inside the generator (which fires the
+    PostHog usage event) always executes, even when the caller breaks out of
+    the loop early.
+    """
+
+    def __init__(self, generator: AsyncGenerator[T, None]) -> None:
+        self._generator = generator
+
+    # ------------------------------------------------------------------ #
+    # Async iterator protocol                                              #
+    # ------------------------------------------------------------------ #
+
+    def __aiter__(self) -> "AsyncStreamWrapper":
+        return self
+
+    async def __anext__(self) -> T:
+        return await self._generator.__anext__()
+
+    # ------------------------------------------------------------------ #
+    # Async context manager protocol                                       #
+    # ------------------------------------------------------------------ #
+
+    async def __aenter__(self) -> "AsyncStreamWrapper":
+        return self
+
+    async def __aexit__(self, exc_type: Any, exc_val: Any, exc_tb: Any) -> bool:
+        # Close the generator so the finally block (PostHog event capture) runs
+        # even on early exit.  If the generator is already exhausted this is a
+        # no-op.
+        await self._generator.aclose()
+        return False
+
+    # ------------------------------------------------------------------ #
+    # Attribute proxy – forward any other attribute access to the         #
+    # underlying generator (e.g. .response on an Anthropic stream).      #
+    # ------------------------------------------------------------------ #
+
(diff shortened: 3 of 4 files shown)


marandaneto:
cc @PostHog/team-ai-observability

marandaneto:
Thanks for the PR — the underlying issue is real and the overall direction makes sense. Issue #393 happens because our async streaming wrappers return a bare async generator, while callers like pydantic-ai expect the provider stream to also support `async with response:`.

Before we merge, could you please address a few things?

1. **Close the original provider stream on context exit**
   `AsyncStreamWrapper.__aexit__()` currently closes only the PostHog wrapper generator. It does not close the original OpenAI/Anthropic SDK stream / HTTP response. Native SDK streams call `close()` on context exit, so the wrapper should also close the original provider stream when available, e.g. via `close()` or `aclose()`.

2. **Fix or remove the attribute proxy behavior**
   `__getattr__` currently proxies to the local async generator, not the original provider stream. That means the docstring/test claim about provider metadata like `.response` is misleading unless the wrapper also keeps a reference to the provider stream and proxies to that.

3. **Fix lint**
   Ruff currently fails because `AsyncIterator` is imported but unused in `posthog/ai/openai/openai_async.py`.

4. **Add regression tests around the actual wrapped clients**
   The direct `AsyncStreamWrapper` tests are useful, but we should also cover the real behavior:
   - `AsyncOpenAI.chat.completions.create(..., stream=True)` works with `async with`
   - `AsyncOpenAI.responses.create(..., stream=True)` works with `async with`
   - `AsyncAnthropic.messages.create(..., stream=True)` works with `async with`
   - early exit from the context closes the original provider stream

5. **Clarify / handle Anthropic `.messages.stream()` compatibility**
   `AsyncAnthropic.messages.stream()` is still an `async def`, so native Anthropic-style usage like `async with client.messages.stream(...)` still returns a coroutine and fails unless awaited first. If this PR intends to restore Anthropic stream API compatibility, that needs addressing; otherwise the PR description should narrow the claim to `messages.create(..., stream=True)`.

I verified the new wrapper tests and existing OpenAI/Anthropic tests pass, but lint currently fails on the unused import.

turnipdabeets:
Picked this up since it went stale — the underlying bug (#393) is real and worth fixing. Opened #645, which rebases the fix onto current `main` (this branch predated the #623/#618 refactors) and addresses @marandaneto's review:

- `__aexit__` now closes the underlying provider stream (not just the tracking generator), releasing the HTTP connection
- `__getattr__` proxies to the provider stream so `.response` metadata is real
- lint clean
- added real-client regression tests (`async with` on OpenAI chat/responses + Anthropic `messages.create(stream=True)`, plus early-exit stream closure)
- scoped out the Anthropic `messages.stream()` async-context mismatch (separate from #393) — noted in the new PR

Credited you as co-author on the commit. Thanks for the original fix and the clear writeup! Suggest closing this in favour of #645.

PR #623, opened 2026-05-24
Refactor AI wrapper callback helpers

marandaneto:

This refactors repeated AI wrapper and LangChain callback helper logic to make the code easier to maintain without changing behavior.

Changes:
- Adds a shared `OpenAIWrapperResource` base class for OpenAI wrapper resources.
- Updates sync and async OpenAI wrappers to reuse the shared fallback/delegation logic.
- Adds small LangChain capture helper methods to centralize debug logging and run capture behavior.

Not run (refactor-only change).
found issues with https://github.com/marandaneto/dry4python

## :pencil: Checklist

### If releasing new changes


marandaneto:
@PostHog/team-client-libraries review pls

posthog/ai/openai/wrapper_utils.py
```diff
@@ -21,3 +21,14 @@ def warn_on_fallback(wrapper_name: str, name: str) -> None:
         wrapper_name,
         name,
     )
+
+
+class OpenAIWrapperResource:
```
ioannisj: Robot says that this is public but should be private instead. Same for all Wrapped_ classes

marandaneto: Wrapped classes are already there, so i cant change it, its a breaking change
OpenAIWrapperResource can be renamed to private (_) prefix since its just a convention, will do

marandaneto: Fixed in 3246dfc — renamed the new internal `OpenAIWrapperResource` helper to `_OpenAIWrapperResource` and updated the sync/async OpenAI wrappers. Kept the existing `Wrapped*` classes unchanged because they already existed before this PR.

ioannisj (approved):

PR #632, opened 2026-05-28
fix: no-op global client for blank api key

marandaneto:
Module-level setup should be safe when the global API key is unset or blank. Instead of raising during setup, blank API keys now create a disabled no-op global client so module-level calls are ignored consistently.

Changes:
- Strip the module-level API key before creating the global client.
- Treat missing, empty, and whitespace-only global API keys as a disabled no-op client.
- Update the module-level API key/setup documentation to describe the no-op behavior in the return value docs.
- Add parameterized coverage for unset, empty, and whitespace-only API keys.
- Add a Sampo patch changeset for the release note.

- `.venv/bin/python -m pytest posthog/test/test_module.py -q`

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/posthog/__init__.py
+++ b/posthog/__init__.py
@@ -266,8 +266,8 @@ For new code, prefer creating an explicit ``Posthog``/``Client`` instance with
 the corresponding constructor arguments.
 
 Attributes:
-    api_key: Project API key/token used by the global client. Required before
-        calling any global capture or feature flag API.
+    api_key: Project API key/token used by the global client. Missing or blank
+        values create a disabled no-op global client.
     host: PostHog ingestion host. Defaults to the US ingestion endpoint when not
         set.
     on_error: Optional callback invoked by background consumers when event upload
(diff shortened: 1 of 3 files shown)


posthog/__init__.py
```diff
@@ -1066,18 +1066,16 @@ def setup() -> Client:
...
     Category:
         Initialization
     """
     global default_client
     if not default_client:
-        if not api_key:
-            raise ValueError("API key is required")
         default_client = Client(
-            api_key,
+            api_key or "",
```
ioannisj: I think it's a good idea to use .strip() on api_key first

ioannisj (approved):

posthog/__init__.py
```diff
@@ -1066,18 +1066,16 @@ def setup() -> Client:
     Returns:
         The global ``Client`` instance.
 
-    Raises:
-        ValueError: If ``api_key`` has not been configured.
+    If ``api_key`` is missing or blank, the returned client is disabled and
+    module-level calls become no-ops.
 
     Category:
```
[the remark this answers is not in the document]
marandaneto: Fixed in 3bd7eaa — folded the blank-key behavior into the Returns section.

posthog/__init__.py
marandaneto: Good call! Fixed in 3bd7eaa — setup now strips the module-level API key before creating the global client.

PR #643, opened 2026-06-02
feat: emit $is_server property on captured events

turnipdabeets:
Adds `$is_server: true` to every event captured by this SDK so PostHog ingestion can identify server-side events.

### DIFF
--- a/.sampo/changesets/emit-is-server-property.md
+++ b/.sampo/changesets/emit-is-server-property.md
@@ -1,0 +1,5 @@
+---
+pypi/posthog: patch
+---
+
+Emit `$is_server` property on captured events so PostHog can identify server-side events.
--- a/.sampo/changesets/emit-is-server-property.md
+++ b/.sampo/changesets/emit-is-server-property.md
@@ -2,4 +2,4 @@
 pypi/posthog: patch
 ---
 
-Emit `$is_server` property on captured events so PostHog can identify server-side events.
+Add a configurable `$is_server` event property (default `true`) so PostHog can identify server-side events. Set `is_server=False` when using posthog-python as a client/CLI so the device OS is attributed normally.
--- a/.sampo/changesets/emit-is-server-property.md
+++ b/.sampo/changesets/emit-is-server-property.md
@@ -1,5 +1,5 @@
 ---
-pypi/posthog: patch
+pypi/posthog: minor
 ---
 
 Add a configurable `$is_server` event property (default `true`) so PostHog can identify server-side events. Set `is_server=False` when using posthog-python as a client/CLI so the device OS is attributed normally.
--- a/posthog/__init__.py
+++ b/posthog/__init__.py
@@ -282,6 +282,9 @@ Attributes:
     poll_interval: Seconds between local feature flag definition refreshes.
     disable_geoip: Whether to disable server-side GeoIP enrichment. Defaults to
         True.
+    is_server: Whether events are emitted from a server-side runtime. Defaults to
+        True; set to False when using the SDK as a client/CLI so the device OS is
+        attributed to the person normally.
     feature_flags_request_timeout_seconds: Timeout in seconds for feature flag
         and remote config requests.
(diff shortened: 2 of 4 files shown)


posthog/test/test_client.py
```diff
@@ -159,6 +159,7 @@ def test_basic_capture(self):
             self.assertEqual(msg["distinct_id"], "distinct_id")
             self.assertEqual(msg["properties"]["$lib"], "posthog-python")
             self.assertEqual(msg["properties"]["$lib_version"], VERSION)
+            self.assertEqual(msg["properties"]["$is_server"], True)
```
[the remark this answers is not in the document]
turnipdabeets: Good catch — addressed both. `$is_server` is now set *after* the `super_properties` merge so a user-provided super property can no longer override the SDK classification, and added `test_is_server_not_overridden_by_super_properties` asserting it stays `True` even when `super_properties` sets `$is_server: False`.

.sampo/changesets/emit-is-server-property.md
```diff
@@ -0,0 +1,5 @@
+---
+pypi/posthog: patch
```
ioannisj: This should probably be minor since we are adding a new constructor param?

posthog/client.py
```diff
@@ -1328,6 +1330,11 @@ def _enqueue(self, msg, disable_geoip):
         if self.super_properties:
             msg["properties"] = {**msg["properties"], **self.super_properties}
 
+        # Set after the super_properties merge so this SDK's server classification
+        # can't be silently overridden by a user-provided super property.
+        if self.is_server:
+            msg["properties"]["$is_server"] = True
```
ioannisj, line 1339: Curious, why do we omit vs set to False?

posthog/__init__.py
```diff
@@ -313,6 +313,7 @@ def get_tags() -> Dict[str, Any]:
 project_api_key = None  # type: Optional[str]
 poll_interval = 30  # type: int
 disable_geoip = True  # type: bool
+is_server = True  # type: bool
```
ioannisj, line 319: We are missing a docstring entry above

ioannisj (approved):
Approving to unblock, left a couple of comments though

.sampo/changesets/emit-is-server-property.md
ioannisj: Your call here but if we change we'll need to update all other SDKs that provide a config option as well

posthog/__init__.py
ioannisj, line 319: Bool would be hard to change in the future without breaking changes if needed. Wondering if we should be defining as String (`server`, `client` for now) which will give us flexibility for the future. Based on the comment [here](https://github.com/PostHog/posthog-go/pull/215#issuecomment-4606720070) for example this could also be "cli" (we'd need to rename the property of course to maybe something like $runtime_type)

I'll hold back review on other SDKs since they follow the same patters so these points can be discussed here?

posthog/client.py
marandaneto: makes sense to me, we dont send on the client side so its the same semantics

posthog/__init__.py
marandaneto: CLI is also a client side, its all about if its running in the users machine or not at the end of the day, does not matter if its a desktop app, a mobile app, a CLI, etc
its either local or cloud, not sure if we need more than a boolean but good point

.sampo/changesets/emit-is-server-property.md
turnipdabeets: update all to use minor

posthog/__init__.py
turnipdabeets: yea I think bool is ok, opening up to strings can make it complicated and we really just want to know if its server or not.

turnipdabeets: updated, should be there for all others as well

PR #657, opened 2026-06-10
feat(exceptions): opt-in Ed25519 signing of $exception events

Gilbert09:
## Problem

Exceptions captured by a backend can be forwarded to systems that act on them (e.g. an internal app that auto-triages data-import failures with an unattended agent). Because the PostHog ingest key is **public**, anyone can capture a forged `$exception` that's indistinguishable from a genuine one — there's no way for a consumer to prove an exception actually came from the customer's backend. That's an injection vector for anything downstream that takes the exception content as input.

## Changes

Adds **opt-in Ed25519 signing of `$exception` events**. A backend service configures an Ed25519 private key; the SDK then signs every captured exception over a canonical projection of its `$exception_list` and attaches:

- `$exception_signature` — base64 Ed25519 signature
- `$exception_signature_key_id` — short key fingerprint (`base64url(sha256(raw_pubkey))[:16]`) for lookup/rotation
- `$exception_signature_version`

PostHog error-tracking ingestion verifies this against the project's registered **public** key and stamps a trusted `$exception_verified` flag (separate changes in PostHog/posthog). PostHog never holds a key that can forge.

- New `posthog/exception_signing.py`: a deterministic **length-prefixed canonical encoding** of `$exception_list` (type + message + each frame's function/filename/lineno/module — excludes everything ingestion mutates and anything float-valued), Ed25519 signing, and key-id derivation. JSON canonicalisation was deliberately avoided in favour of explicit length-prefixing to keep the bytes identical across the SDK and the Rust verifier.
- Config: `enable_exception_signing` + `exception_signing_private_key` (PEM), wired through `Client.__init__`, the module proxies, and `setup()`.
- Signing runs in `_enqueue` **after** `before_send`, so it covers the final sent content and a `before_send` callback can't strip it. Failures leave the event unsigned, never dropped.
- Crypto (`cryptography`) is an **optional `[exception-signing]` extra** so the base SDK stays lean.

**Backend use only** — never ship a private key in a browser/mobile app.

## How tested

Agent-assisted. `pytest posthog/test/test_exception_signing.py` (14 tests) + the existing before_send/exception-capture suites pass. Includes a **cross-language parity vector** — a fixed keypair signing a fixed exception list, asserting the exact canonical bytes + signature — that the Rust verifier (PostHog/posthog) reproduces, guarding against any drift between the two implementations. Verified sign→verify round-trips with a generated keypair; confirmed signing happens after before_send.

### DIFF
--- a/.sampo/changesets/exception-signing.md
+++ b/.sampo/changesets/exception-signing.md
@@ -1,0 +1,5 @@
+---
+pypi/posthog: minor
+---
+
+feat(exceptions): add opt-in Ed25519 signing of `$exception` events. Set `enable_exception_signing=True` and provide an Ed25519 private key in `exception_signing_private_key`, then register the matching public key in your PostHog project. The SDK signs each captured exception over a canonical projection of its `$exception_list`, so error-tracking ingestion can verify it genuinely came from your backend (rather than being forged through the public ingest key) and mark it verified. Backend use only — never ship a private key in a browser/mobile app. Requires the new `[exception-signing]` extra (`pip install posthoganalytics[exception-signing]`).
--- a/posthog/__init__.py
+++ b/posthog/__init__.py
@@ -338,6 +338,13 @@ code_variables_mask_patterns = DEFAULT_CODE_VARIABLES_MASK_PATTERNS
 code_variables_ignore_patterns = DEFAULT_CODE_VARIABLES_IGNORE_PATTERNS
 in_app_modules = None  # type: Optional[list[str]]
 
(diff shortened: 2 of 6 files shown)


marandaneto:
@PostHog/team-error-tracking i think this is you first

Gilbert09:
Closing as superseded by PostHog/posthog#62888 — verification is now transport-level: configure the SDK with the project's secret API key (`api_key="phs_..."`) and all captured events are marked `$verified` during ingestion. No SDK changes required.

PR #662, opened 2026-06-12
feat(exception-capture): add client-side token bucket rate limiting

hpouillot:

Addresses the long-standing `TODO` in `posthog/exception_capture.py`: exception autocapture has no client-side rate limiting, so a crash loop can flood the ingestion queue.

This ports the `BucketedRateLimiter` from posthog-js ([`packages/core/src/utils/bucketed-rate-limiter.ts`](https://github.com/PostHog/posthog-js/blob/main/packages/core/src/utils/bucketed-rate-limiter.ts)) and applies it to exception autocapture as an **opt-in** feature:

- **disabled by default** — without the flag, behavior is identical to released versions (the limiter is not even constructed)
- one token bucket **per exception type** (the Python equivalent of `$exception_list[0].type`, falling back to `"Exception"`)
- rate-limited exceptions are skipped before reaching the ingestion queue, logging `Skipping exception capture because of client rate limiting.` like the other SDKs
- defaults are more generous than the browser/Node SDKs' 10 / 1 / 10s because one server process aggregates exceptions across many users' requests

New `Client` options (also available as module-level settings):

| Option | Default | Description |
|---|---|---|
| `enable_exception_autocapture_rate_limiting` | `False` | Opt into client-side rate limiting |
| `exception_autocapture_bucket_size` | `50` | Max burst of captures per exception type (clamped to 0–100) |
| `exception_autocapture_refill_rate` | `10` | Tokens restored per refill interval |
| `exception_autocapture_refill_interval_seconds` | `10` | Seconds between refills |

Deviations from the JS source, since this SDK runs in threaded server processes:

- guarded by a `threading.Lock` (the `on_bucket_rate_limited` callback fires outside the lock)
- injectable monotonic clock for tests

One JS quirk is preserved for cross-SDK parity: the call that drains the bucket is itself reported as limited, so a burst lets `bucket_size - 1` events through.

- `posthog/test/test_bucketed_rate_limiter.py`: 25 tests porting the posthog-js spec ([`bucketed-rate-limiter.spec.ts`](https://github.com/PostHog/posthog-js/blob/main/packages/core/src/utils/bucketed-rate-limiter.spec.ts)) — consumption, refill math, partial intervals, bucket isolation, callback semantics, `stop()`, timestamp carry-over — plus Python-specific tests: parameter clamping, a 10-thread concurrency test asserting exactly `bucket_size - 1` of 200 contended consumes pass, `ExceptionCapture` integration (15 same-type exceptions on a size-10 bucket → 9 captured; a different type still passes), disabled-by-default (100 captures pass untouched, no limiter constructed), enabled defaults, and config pass-through from `Client` kwargs to the limiter.
- Existing `posthog/test/test_exception_capture.py` (15 tests) still passes.
- `ruff format`, `ruff check`, and `mypy` are clean.

## :pencil: Checklist

### If releasing new changes


marandaneto (approved):

posthog/exception_capture.py
```diff
@@ -44,7 +68,20 @@ def exception_receiver(self, exc_info, extra_properties):
 
     def capture_exception(self, exception, metadata=None):
         try:
+            if self._rate_limiter is not None:
+                exception_type = self._exception_type(exception)
```
cat-ph, line 73: is this the correct order with chained exceptions? I think that might consume the top-level one instead, i.e. `RuntimeError from ZeroDivisionError` consumes RuntimeError instead of ZeroDivisionError

```py
client = MagicMock()
capture = ExceptionCapture(client, rate_limiting_enabled=True, bucket_size=2)

capture.capture_exception(exc_info_for(RuntimeError("wrapped", ZeroDivisionError())))
capture.capture_exception(exc_info_for(RuntimeError("wrapped", KeyError())))

assert client.capture_exception.call_count == 2
```

I think this fails, but we'd want it to pass?

cat-ph (approved):

posthog/exception_capture.py
hpouillot: Good catch ! Fixed it in latest commit

PR #672, opened 2026-06-16
fix: increase default flush interval

marandaneto:

The Python SDK currently flushes partial background batches after 500ms by default, which is much more aggressive than the other backend SDK defaults. This increases the default background flush interval to 5 seconds while keeping the existing `flush_at = 100` count trigger.

Default backend SDK context:

| SDK | Default flush interval | Count trigger |
|---|---:|---:|
| `posthog-python` before this PR | 500 ms | `flush_at = 100` |
| `posthog-python` after this PR | 5,000 ms | `flush_at = 100` |
| `posthog-dotnet` | 30,000 ms | `FlushAt = 20` |
| `posthog-go` | 5,000 ms | `BatchSize = 250` |
| `posthog-elixir` | 10,000 ms | `max_batch_events = 100` |
| `posthog-js` / `posthog-node` | 10,000 ms | `flushAt = 20` |
| `posthog-php` | No timer | `batch_size = 100`; flushes on batch size/manual/destruct |
| `posthog-ruby` | No fixed timer | `batch_size = 100`; worker sends queued events immediately |
| `posthog-rs` | No internal queue timer | Captures send immediately; batching is explicit |

- `uv run python .github/scripts/check_public_api.py`
- `uv run pytest posthog/test/test_consumer.py posthog/test/test_client.py::TestClient::test_user_defined_flush_at`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

An agent updated the Python client and consumer default flush interval from 0.5s to 5s, aligned the SDK compliance adapter default `flush_interval_ms` to 5000, regenerated the public API snapshot, and added a Sampo changeset. The default-values table above was gathered from the locally cloned PostHog SDK repositories to give reviewers cross-SDK context for the chosen 5s value.


marandaneto:
@PostHog/team-client-libraries @cat-ph 
i think 0.5s is too low
5s is more reasonable
rust is adding 5s as well
wdyt?

cat-ph (approved):

dustinbyrne (approved):

PR #685, opened 2026-06-19
fix: improve strict public API types

### LINKED ISSUE #559: Improve Python SDK typing for strict pyright/mypy projects
## Summary

While integrating the PostHog Python SDK into a strict Python client, we hit several quality-gate failures around strict typing and optional SDK initialization. We were able to work around them locally, but the experience was rough enough that I wanted to report it upstream.

The main friction points were:

- `pyright` reported partially unknown types when using the PostHog client object, especially around `shutdown`.
- `mypy` could not find library stubs for `posthog`, which forced us to avoid a normal static import.
- Optional initialization paths were easy to make type-checker-hostile because the SDK client type is not available in a typed way.

## Environment

- Package: `posthog`
- Language: Python
- Type checkers:
  - `pyright`
  - `mypy`
- Project using it: Python client
- Python observed in test run: `3.14.3`

### Pyright

```text
analytics.py: Type of "shutdown" is partially unknown
```

This happened when registering shutdown with `atexit` from a lazily initialized PostHog client.

### Mypy

```text
analytics.py: Unused "type: ignore" comment
Cannot find implementation or library stub for module named "posthog"
```

The direct import:

```python
from posthog import Posthog
```

caused `mypy` to complain about missing stubs. Adding a `type: ignore` was not acceptable in our repository's quality policy, and in one run it also became an unused ignore.

### Downstream optional-client fallout

After the SDK wrapper passed lint/type checks, our test suite exposed a separate crash caused by analytics code assuming a fully initialized serving client:

```text
AttributeError: 'ServingClient' object has no attribute '_base_url'
```

That one was our bug, but it was triggered in the analytics call path and made the integration more fragile than expected.

## Workaround used

We ended up avoiding the static import and defining a local protocol for the subset of the SDK we use:

```python
from importlib import import_module
from typing import Any, Protocol, cast


class _PostHogClient(Protocol):
    def capture(self, *, distinct_id: str, event: str, properties: dict[str, Any]) -> None: ...

    def capture_exception(self, exc: BaseException, distinct_id: str) -> None: ...

    def shutdown(self) -> None: ...


posthog_module = import_module("posthog")
posthog_class = posthog_module.Posthog
client = cast(_PostHogClient, posthog_class(project_token, **kwargs))
```

This works, but it is more ceremony than we expected for a small optional analytics integration.

## Expected behavior

It would be helpful if the Python SDK shipped enough typing metadata for strict projects to use the normal import path without local protocols/casts:

```python
from posthog import Posthog

client = Posthog(project_token, enable_exception_autocapture=True)
client.capture(...)
client.capture_exception(...)
client.shutdown()
```

Ideally this should pass both `pyright` and `mypy` without downstream projects needing custom stubs, local protocols, or ignore comments.

## Why this matters

Strict Python projects often prohibit `type: ignore`/`noqa` style suppressions in CI. Without typed exports or bundled stubs, integrating PostHog requires either loosening quality policy or hiding the SDK behind local dynamic-import wrappers.

marandaneto:

Strict Pyright users could still hit partially unknown types when accessing public PostHog APIs, even after the issue-specific `Posthog.shutdown` path was fixed.

Fixes #559.

- Compared the SDK event-facing types against the capture `/batch/` backend contract in `rust/capture/src/v0_request.rs` and `rust/common/types/src/event.rs`. The SDK intentionally keeps some wrapper timestamp types stricter than the backend accepts.
- `python -m mypy --config-file mypy.ini posthog/client.py posthog/__init__.py`
- `uv run ruff check posthog/client.py posthog/__init__.py`
- `uv run --extra dev python .github/scripts/check_public_api.py`
- `uv run --extra test pytest posthog/test/test_feature_flag.py posthog/test/test_feature_flag_result.py posthog/test/test_module.py::TestModule::test_flush`
- `.github/scripts/check_strict_types.sh`
- Added a CI `Strict type smoke` job that runs `.github/scripts/check_strict_types.sh`. The script installs the package into a temporary downstream venv, runs strict Pyright against representative public API usage, and checks all public client/module function member access for unknown types.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

An AI coding agent investigated the strict typing gaps around #559, identified remaining Pyright `Unknown` types on public APIs, and made a typing-only fix. The changes focus on explicit public signatures in `posthog.client.Client` and the module-level wrappers in `posthog.__init__`, plus a patch changeset for release notes and a strict downstream Pyright smoke check in CI.


posthog/__init__.py
```diff
@@ -664,16 +662,16 @@ def feature_enabled(
...
+    key: str,
+    distinct_id: ID_TYPES,
+    groups: Optional[Dict[str, str]] = None,
+    person_properties: Optional[Dict[str, Any]] = None,
+    group_properties: Optional[Dict[str, Dict[str, Any]]] = None,
+    only_evaluate_locally: bool = False,
+    send_feature_flag_events: bool = True,
+    disable_geoip: Optional[bool] = None,
+    device_id: Optional[str] = None,
+) -> Optional[FlagValue]:
```
marandaneto, line 698: why is this changed?

posthog/__init__.py
```diff
@@ -717,15 +715,15 @@ def get_feature_flag(
...
-) -> Optional[dict[str, FeatureFlag]]:
+    distinct_id: ID_TYPES,
+    groups: Optional[Dict[str, str]] = None,
+    person_properties: Optional[Dict[str, Any]] = None,
+    group_properties: Optional[Dict[str, Dict[str, Any]]] = None,
+    only_evaluate_locally: bool = False,
+    disable_geoip: Optional[bool] = None,
+    device_id: Optional[str] = None,
+    flag_keys_to_evaluate: Optional[list[str]] = None,
+) -> Optional[dict[str, FlagValue]]:
```
marandaneto, line 750: why is this changed?

posthog/__init__.py
```diff
@@ -821,17 +818,17 @@ def get_feature_flag_result(
...
+    distinct_id: ID_TYPES,
+    match_value: Optional[FlagValue] = None,
+    groups: Optional[Dict[str, str]] = None,
+    person_properties: Optional[Dict[str, Any]] = None,
+    group_properties: Optional[Dict[str, Dict[str, Any]]] = None,
+    only_evaluate_locally: bool = False,
+    send_feature_flag_events: bool = True,
+    disable_geoip: Optional[bool] = None,
+    device_id: Optional[str] = None,
+) -> Optional[Any]:
```
marandaneto: why is this changed?

posthog/client.py
```diff
@@ -2755,7 +2757,7 @@ def _get_all_flags_and_payloads_locally(
         self,
         distinct_id: ID_TYPES,
         *,
-        groups: Dict[str, Union[str, int]],
+        groups: Dict[str, str],
```
marandaneto: why is this changed?

posthog/__init__.py
```diff
@@ -61,7 +61,8 @@
 )
 from posthog.types import (
     BeforeSendCallback as BeforeSendCallback,
-    FeatureFlag,
+    FeatureFlag as FeatureFlag,
+    FlagValue,
     FlagsAndPayloads,
 )
```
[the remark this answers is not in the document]
marandaneto: Good catch! Fixed in d408314 — `FlagValue` now uses the explicit `FlagValue as FlagValue` re-export form, and I added a strict Pyright smoke check for importing it from `posthog`.

posthog/client.py
marandaneto: Fixed in d408314 — this helper no longer narrows group keys to `str`; it now accepts a read-only mapping with `str | int` values, preserving the previous accepted shape while still allowing strict public callers with `dict[str, str]`.

posthog/__init__.py
marandaneto: This return type changed because the wrapper proxies to `Client.get_feature_flag()`, which returns a flag value (`bool | str | None`), not a `FeatureFlag` object. The old annotation was inaccurate and caused downstream strict type checkers to infer the wrong API contract.

posthog/__init__.py
marandaneto: This changed for the same reason: `get_all_flags()` returns a mapping of flag keys to flag values (`bool | str`), not `FeatureFlag` objects. The updated annotation matches the runtime return value from `Client.get_all_flags()`.

posthog/__init__.py
marandaneto: This changed because feature flag payloads are not guaranteed to be strings at runtime; they can be decoded JSON values. The client method returns `Optional[Any]`, so the module-level wrapper now matches that actual API behavior.

posthog/client.py
```diff
@@ -2755,7 +2757,7 @@ def _get_all_flags_and_payloads_locally(
         self,
         distinct_id: ID_TYPES,
         *,
-        groups: Dict[str, Union[str, int]],
+        groups: Mapping[str, Union[str, int]],
```
marandaneto, line 2986: why did we change from Dict to Mapping?

marandaneto: This keeps the helper from narrowing the previous accepted shape (`str | int` group keys), while still allowing callers that pass a normal `dict[str, str]`. `dict` is invariant, so `dict[str, str]` is not accepted where `Dict[str, str | int]` is expected; `Mapping` is covariant for values and works for this read-only helper.

.github/workflows/ci.yml
```diff
@@ -77,6 +77,64 @@ jobs:
...
                   make public_api_check
 
+    strict-type-smoke:
+        name: Strict type smoke
+        runs-on: ubuntu-latest
+
+        steps:
+            - uses: actions/checkout@85e6279cec87321a52edac9c87bce653a07cf6c2
+              with:
+                  fetch-depth: 1
```
marandaneto, line 87: do we need this?

.github/workflows/ci.yml
```diff
@@ -77,6 +77,64 @@ jobs:
...
+                  enabled: bool | None = posthog.feature_enabled("flag", "user")
+                  payload: object | None = client.get_feature_flag_payload("flag", "user")
+
+                  _ = (flag_value, all_flags, enabled, payload)
+                  PY
+
+                  cat > "$tmp/pyrightconfig.json" <<JSON
+                  {
+                    "typeCheckingMode": "strict",
+                    "pythonVersion": "3.14",
```
marandaneto: we set up python 3.11 but use 3.14 here why

.github/workflows/ci.yml
```diff
@@ -77,6 +77,64 @@ jobs:
...
+                  fetch-depth: 1
+
+            - name: Set up Python 3.11
+              uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55
+              with:
+                  python-version: 3.11.11
+
+            - name: Check strict downstream Pyright usage
+              shell: bash
+              run: |
```
marandaneto: can we move this to a bash script in a file

.github/workflows/ci.yml
```diff
@@ -77,6 +77,64 @@ jobs:
...
+                  import posthog
+                  from posthog import FlagValue, Posthog
+
+                  client = Posthog("phc_test")
+                  atexit.register(client.shutdown)
+
+                  flag_value: FlagValue | None = client.get_feature_flag("flag", "user")
+                  all_flags: dict[str, FlagValue] | None = posthog.get_all_flags("user")
+                  enabled: bool | None = posthog.feature_enabled("flag", "user")
+                  payload: object | None = client.get_feature_flag_payload("flag", "user")
```
marandaneto: we have more public APIs, do we need to call all APIs to valida this?

.github/workflows/ci.yml
marandaneto: Yes, I think it is useful for this PR because it exercises the downstream scenario from #559 that mypy and the public API snapshot do not cover: installing the package and running strict Pyright against public SDK usage. Kept it small and moved the implementation to `.github/scripts/check_strict_types.sh` in 37cd471/9bd5b40.

.github/workflows/ci.yml
marandaneto: Fixed in 37cd471 — Pyright now defaults to Python 3.11 in `.github/scripts/check_strict_types.sh`, matching the CI setup-python version.

.github/workflows/ci.yml
marandaneto: Done in 37cd471 — the workflow now just runs `.github/scripts/check_strict_types.sh`.

.github/workflows/ci.yml
marandaneto: Good point. I expanded the script in 9bd5b40 so it does not call every API (to avoid side effects), but it does generate and strict-check member access for all public client/module functions. The representative call file still covers the key strict downstream usage from #559.

posthog/__init__.py
```diff
@@ -3,7 +3,7 @@
 
 from typing_extensions import Unpack
 
-from posthog.args import ExceptionArg, OptionalCaptureArgs, OptionalSetArgs
+from posthog.args import ID_TYPES, ExceptionArg, OptionalCaptureArgs, OptionalSetArgs
```
ioannisj: Similar to https://github.com/PostHog/posthog-python/pull/685#discussion_r3442792172 should we do the same for ID_TYPES?

posthog/__init__.py
```diff
@@ -61,7 +61,8 @@
 )
 from posthog.types import (
     BeforeSendCallback as BeforeSendCallback,
-    FeatureFlag,
+    FeatureFlag as FeatureFlag,
+    FlagValue as FlagValue,
     FlagsAndPayloads,
```
ioannisj: Same for this?

posthog/client.py
```diff
@@ -2169,18 +2169,18 @@ def _locally_evaluate_flag(
...
-        disable_geoip=None,
+        groups: Optional[Dict[str, str]] = None,
+        person_properties: Optional[Dict[str, Any]] = None,
+        group_properties: Optional[Dict[str, Dict[str, Any]]] = None,
+        only_evaluate_locally: bool = False,
+        send_feature_flag_events: bool = False,
+        disable_geoip: Optional[bool] = None,
         device_id: Optional[str] = None,
-    ):
+    ) -> Optional[Any]:
```
ioannisj: Should tihs be Optional[Object]?

posthog/__init__.py
marandaneto: Done in d96fc4d — `ID_TYPES` now uses the explicit `ID_TYPES as ID_TYPES` re-export form.

posthog/__init__.py
marandaneto: Done in d96fc4d — `FlagsAndPayloads` now uses the explicit `FlagsAndPayloads as FlagsAndPayloads` re-export form too.

posthog/client.py
marandaneto: Agreed. Changed both `Client.get_feature_flag_payload()` and the module-level wrapper to return `Optional[object]` in d96fc4d, and refreshed the public API snapshot.

posthog/__init__.py
```diff
@@ -840,17 +842,17 @@ def get_feature_flag_result(
...
+    distinct_id: ID_TYPES,
+    match_value: Optional[FlagValue] = None,
+    groups: Optional[Dict[str, str]] = None,
+    person_properties: Optional[Dict[str, Any]] = None,
+    group_properties: Optional[Dict[str, Dict[str, Any]]] = None,
+    only_evaluate_locally: bool = False,
+    send_feature_flag_events: bool = True,
+    disable_geoip: Optional[bool] = None,
+    device_id: Optional[str] = None,
+) -> Optional[object]:
```
marandaneto, line 855: should that be `Optional[FeatureFlagResult]`? since this returns the result from `_get_feature_flag_result`

posthog/client.py
```diff
@@ -2390,18 +2390,18 @@ def _locally_evaluate_flag(
...
-        disable_geoip=None,
+        groups: Optional[Dict[str, str]] = None,
+        person_properties: Optional[Dict[str, Any]] = None,
+        group_properties: Optional[Dict[str, Dict[str, Any]]] = None,
+        only_evaluate_locally: bool = False,
+        send_feature_flag_events: bool = False,
+        disable_geoip: Optional[bool] = None,
         device_id: Optional[str] = None,
-    ):
+    ) -> Optional[object]:
```
marandaneto, line 2404: same as https://github.com/PostHog/posthog-python/pull/685/changes#r3452135316

posthog/__init__.py
marandaneto: This should stay `Optional[object]`: the wrapper calls `get_feature_flag_payload()`, which returns only the payload, not the `FeatureFlagResult`. Clarified the docstring in e753e61.

posthog/client.py
marandaneto: Same here: `_get_feature_flag_result()` is used internally, but this method returns `feature_flag_result.payload`. Clarified the docstring in e753e61.

ioannisj (approved):

PR #688, opened 2026-06-19
feat: mask sensitive data inside objects and URLs in code variables

ablaszkiewicz:
# Overview

- **It's faster.** Much faster on nasty inputs (~57× on the worst case, comfortably sub-1ms) and faster than where we started even on ordinary exceptions. We compile the masking patterns once and cache them, fast-path the common value types, and collapse all the secret-name matching into a single regex pass.

- **A little bit stricter limits — it's best-effort.** If some exception has one bazillion triple-nested variables
and crazy collections, we just stop very early instead of grinding through all of it. Hard caps on depth, collection width, total nodes, and string length.

- **Connection URL detection.** We scrub `user:pass@host` credentials out of DSNs / connection strings
(`postgres://…`, `redis://…`, etc.), so a URL sitting in a local variable can't leak its password.

- **Much more support for all the crazy ways a secret can end up in the final event.** Secrets don't only live in plain attributes, so we now also catch sensitively-named `@property`, `cached_property`, `__slots__`, class-level
attributes, descriptors and namedtuple fields, plus weird non-string dict keys. We never call a getter, and never
trust a custom `__repr__` that could rename a field out of the mask — when in doubt, fail closed.

- **Brand new, ultra clean tests.** Rewrote the entire suite. One idea per assertion.
# Benchmark

## Examples

| | **small** | **medium** | **maxed** |
  |---|---|---|---|
  | **Represents** | a typical, everyday exception | a "busy" but still realistic frame | a deliberately pathological worst case |
  | **Realistic?** | yes — the common case | yes — heavy but plausible | no — a stress test |
  | **Top-level vars** | 9 | 11 | 57 |
  | **Biggest structure** | a 3-key dict, a 5-int list | nested dict (depth 4), 40-row list | fan-out tree: ~111k nodes + 1M leaves |
  | **Long strings** | none | 4 KB + 4.5 KB | 5 KB, 20 KB, 4.8 KB |
  | **Limits it hits** | none | per-value output truncation (1 KB) | **all of them** |
  
  ## Benchmark vs `main`
  
  | Workload | `main` | `current` | Speedup |
  |---|--:|--:|--:|
  | small (9 vars) | 21 µs | 18 µs | 1.15× faster |
  | medium (11 vars) | 194 µs | 182 µs | 1.07× faster |
  | maxed (57 vars) | 497 ms | 0.86 ms | **576× faster** |


posthog/test/test_exception_capture.py
```diff
@@ -809,3 +809,339 @@
...
+    assert "ZeroDivisionError" in output
+    assert "code_variables" in output
+
+    # The secret must never appear, in any form.
+    assert "uHjH9WJuEV0VT2NKoP7zpQ" not in output
+    assert "$$_posthog_redacted_based_on_masking_rules_$$" in output
+
+    # Surrounding, non-sensitive context is preserved.
+    assert "PostgresSourceConfig" in output
+    assert "db.example.com" in output
```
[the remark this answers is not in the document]
ablaszkiewicz: Fixed

posthog/exception_utils.py
```diff
@@ -979,49 +1006,145 @@ def _pattern_matches(name, patterns):
...
-            return "<circular ref>"
-        _seen.add(obj_id)
+    if isinstance(value, str):
+        if len(value) > _MAX_VALUE_LENGTH_FOR_PATTERN_MATCH:
+            return CODE_VARIABLES_TOO_LONG_VALUE
+        if _pattern_matches(value, compiled_mask):
+            return CODE_VARIABLES_REDACTED_VALUE
+        if mask_url_credentials:
+            return _redact_url_credentials(value)
+        return value
```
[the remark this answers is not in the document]
ablaszkiewicz: Fixed

posthog/test/test_exception_capture.py
```diff
@@ -809,3 +809,339 @@ def test_mask_sensitive_data_large_tuple_replaced():
...
+    ) == ("postgresql://" + CODE_VARIABLES_REDACTED_VALUE + "@db.example.com:26257/db")
+    # Password-only userinfo is also caught.
+    assert "p4ss" not in _redact_url_credentials("redis://:p4ss@cache:6379")
+    # A URL with a port but no credentials is left untouched.
+    assert (
+        _redact_url_credentials("https://api.example.com:8080/v1")
+        == "https://api.example.com:8080/v1"
+    )
+    # Non-URL strings pass through unchanged.
+    assert _redact_url_credentials("just a string") == "just a string"
```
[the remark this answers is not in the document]
ablaszkiewicz: I'll stay consistent with what's already in this file

posthog/test/test_exception_capture.py
```diff
@@ -809,3 +809,508 @@
...
+        __slots__ = ()
+
+        def __repr__(self):
+            return "Conn(url=postgresql://user:leakpw@db.example.com/app)"
+
+    compiled_mask = _compile_patterns(DEFAULT_CODE_VARIABLES_MASK_PATTERNS)
+    result = _safe_repr(Conn(), compiled_mask)
+
+    assert "leakpw" not in result
+    assert "db.example.com" in result
```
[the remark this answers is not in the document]
ablaszkiewicz: Fixed

posthog/test/test_exception_capture.py
```diff
@@ -809,3 +809,551 @@ def test_mask_sensitive_data_large_tuple_replaced():
     result = _mask_sensitive_data(large_tuple, compiled_mask)
 
     assert result == CODE_VARIABLES_TOO_LONG_VALUE
+
+
+def test_mask_sensitive_data_traverses_object_attributes():
```
hpouillot: Edge-case but it won't work with deeply nested structure:

```python
from dataclasses import dataclass
 from posthog.exception_utils import _compile_patterns, _mask_sensitive_data, DEFAULT_CODE_VARIABLES_MASK_PATTERNS

 @dataclass
 class Box:
     password: str
     def __repr__(self):
         return f"Box({self.password})"

 class Node:
     def __init__(self, child):
         self.child = child

 for n in [24,25,26]:
     value = Box("hunter2")
     for _ in range(n):
         value = Node(value)
     compiled_mask = _compile_patterns(DEFAULT_CODE_VARIABLES_MASK_PATTERNS)
     result = _mask_sensitive_data(value, compiled_mask)
     print('nodes', n, 'leaked?', 'hunter2' in str(result))
     if 'hunter2' in str(result): print(result)
```

ablaszkiewicz: Fixed and added regression test

hpouillot: Hum, this seems to leak as well (properties are not traversed)

 ```python
   from posthog.exception_utils import (
       _compile_patterns,
       _mask_sensitive_data,
       DEFAULT_CODE_VARIABLES_MASK_PATTERNS,
   )

   class Config:
       @property
       def password(self):
           return "hunter2"

       def __repr__(self):
           return f"Config({self.password})"

   mask = _compile_patterns(DEFAULT_CODE_VARIABLES_MASK_PATTERNS)
   result = _mask_sensitive_data(Config(), mask)

   print(result)
   print("leaked?", "hunter2" in str(result))
   print(Config().__dict__)
 ```

ablaszkiewicz: Refactored everything. No leaks

.changeset/brave-otters-mask.md
```diff
@@ -0,0 +1,5 @@
+---
+'pypi/posthog': minor
```
hpouillot: let's make it a patch ?

ablaszkiewicz: Hm ok yeah. My thinking was that I'd make this a minor because this changes default behavior (we strip connection strings by default) but after 2nd thought I highly doubt that someone relied on it 😆

ablaszkiewicz: Done!

hpouillot (approved):

PR #690, opened 2026-06-22
chore: relocate pending changesets to .sampo/changesets

ablaszkiewicz:
Fix recent releases


marandaneto:
whaat

marandaneto:
@ablaszkiewicz lets also delete the .changeset folder so agents dont get confused

marandaneto (approved):

cat-ph (approved):
oof, good catch! I think the deleted changesets were already released? do we need to backfill changelog with their descriptions? 😬

ablaszkiewicz:
@cat-ph. Most of them (just no the 2 I moved) were released. I wouldn't touch changelog which is now auto generated. The two remaining will be added to the next patch

marandaneto:
> @cat-ph. Most of them (just no the 2 I moved) were released. I wouldn't touch changelog which is now auto generated. The two remaining will be added to the next patch

i think we can manually append the changeset to the correct version if already released, otherwise its gonna list in the wrong version
we can also update the github releases manually if needed

marandaneto:
i am auditing other sdks, maybe there were more cases with sampo releases

PR #692, opened 2026-06-23
feat: entropy-based secret detection for exception code variables

ablaszkiewicz:
Currently we detect secrets based on keys and values having common secret phrases like `api-key`, `password`, etc...

@hpouillot got an amazing idea to do an entropy based secrets detection. This PR does that and extends our detection with popular secrets format like `sk-ant`, `ey...`, `gh_...`, `glpat_...`.

There is a problem however - sometimes high entropy strings are classified as secrets. We do our best to detect genuine high entropy non-secrets and stop the entropy detection process if it's:
- UUID,
- file or URL path,
- something [in these brackets] or in {these brackets}.

There are much more cases and rules - it's in the PR.

I also tightened our search limits based on real world exceptions.

Did a benchmark on 3 real exceptions. This computes how long does it take to capture variables from frames
  | Case | Before (master) (ms) | After (ms) | Δ (ms) |
  |:---|---:|---:|---:|
  | eventA | 0.062 | 0.083 | +0.021 (+34%) | 
  | eventB | 0.040 | 0.053 | +0.013 (+31%) |
  | eventC | 0.040 | 0.053 | +0.012 (+31%) |

### DIFF
--- a/posthog/exception_utils.py
+++ b/posthog/exception_utils.py
@@ -1249,6 +1249,9 @@ def _safe_repr(value, config):
         return CODE_VARIABLES_REDACTED_VALUE
     if _matcher_matches(rendered, config.mask):
         return CODE_VARIABLES_REDACTED_VALUE
+    # A __repr__ that is itself a bare secret would otherwise bypass detection.
+    if config.detect_secrets and _looks_like_secret(rendered):
+        return CODE_VARIABLES_REDACTED_VALUE
     if config.mask_url_credentials:
         return _redact_url_credentials(rendered)
     return rendered
--- a/posthog/exception_utils.py
+++ b/posthog/exception_utils.py
@@ -1118,18 +1118,45 @@ _UUID_RE = re.compile(
 # Lowercase word-like path segment; two around a '/' marks a filesystem path.
 _PATH_WORD_RE = re.compile(r"\A[a-z][a-z.]*\Z")
 
-# Well-known credential formats, matched regardless of entropy.
+# Well-known credential formats, matched regardless of entropy. High-confidence,
+# distinctive-prefix patterns adapted from the gitleaks / detect-secrets rule sets.
 _KNOWN_SECRET_PATTERNS = [
+    # AI / LLM providers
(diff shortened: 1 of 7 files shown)


posthog/test/test_code_variables.py
```diff
@@ -908,3 +910,180 @@ def trigger_error():
...
         assert "code_variables" in output
         assert "p4ssRUNTIME" in output  # URL masking disabled -> credential retained
+
+
+# --- entropy-based secret detection (last resort) ------------------------------------
+
+
+# Synthetic, format-correct fakes (no real credentials). Vendor keys are assembled from
+# prefix + body so no complete secret literal lives in source (which trips secret scanners).
+def _key(prefix, body):
```
ablaszkiewicz, line 920: I had to split secrets into two parts. Otherwise GitHub wasn't happy

posthog/exception_utils.py
cat-ph, line 1270: do we want to run `_looks_like_secret(rendered)` here too maybe? 🤔 otherwise something like this bypasses it:

```py
from posthog.exception_utils import _MaskingConfig, _encode_variable, VariableSizeLimiter

class OpaqueToken:
    __slots__ = ()

    def __repr__(self):
        return "n8fK2pQ9vX7mL4wR8tY3uZ6bC1dE5gH"  # high-entropy fake token

config = _MaskingConfig.build(
    [], [],
    mask_url_credentials=False,
    detect_secrets=True,
)

print(_encode_variable(OpaqueToken(), config, VariableSizeLimiter()))
```

cat-ph (approved):
love this!! 🚀🚀

posthog/exception_utils.py
ablaszkiewicz: Good catch. Fixed and added a test

marandaneto:
i've used https://github.com/Yelp/detect-secrets/tree/master/detect_secrets/plugins (see regexes) in the past, not sure if its up to date tho

hpouillot (approved):

ablaszkiewicz:
Thanks @marandaneto I borrowed some stuff from them 8)

marandaneto (approved):

PR #695, opened 2026-06-25
feat(flags): add official PostHog OpenFeature provider

gustavohstrassburger:
## What

Adds an **official PostHog provider for the [OpenFeature](https://openfeature.dev) Python SDK**, built in this repo and shipped as a **separate distribution** (`openfeature-provider-posthog`) under the OpenFeature namespace package `openfeature.contrib.provider.posthog`.

It wraps a configured `posthog.Posthog` client and resolves all five OpenFeature flag types via the modern, non-deprecated `get_feature_flag_result` (one call → value + variant + payload + reason):

| OpenFeature method | Resolves to |
| --- | --- |
| `get_boolean_value` | flag `enabled` |
| `get_string_value` | multivariate variant key |
| `get_integer_value` / `get_float_value` | variant parsed as a number |
| `get_object_value` | flag's JSON **payload** (full object support) |

**Evaluation context** maps `targeting_key` → `distinct_id`, reserved attributes `groups` / `group_properties` → PostHog groups, and every other attribute → `person_properties`.

## Behavior decisions (recommended defaults, easy to change)

- Missing `targeting_key` → `TargetingKeyMissingError` (OpenFeature-idiomatic; SDK returns the caller's default). Opt into anonymous eval via `default_distinct_id="anonymous"`.
- Type mismatch (e.g. `get_string_value` on a boolean flag, non-numeric variant for int, non-object payload) → `TypeMismatchError` → default returned, per spec.
- `send_feature_flag_events=True` by default (keeps `$feature_flag_called` / experiments working); toggleable.
- Build backend setuptools (repo-consistent); `posthog>=6.0.0` floor.

## Layout

```
openfeature-provider/
├── pyproject.toml                # dist: openfeature-provider-posthog (MIT, py>=3.10)
├── README.md / LICENSE / .gitignore / uv.lock
├── openfeature/contrib/provider/posthog/   # PEP 420 namespace; __init__ only at leaf
│   ├── __init__.py  ├── provider.py  └── py.typed
└── tests/            # 20 unit + end-to-end tests (mocked client)
```

## Repo wiring

- **CI**: new `openfeature-provider` job mirroring `django5-integration` — `uv sync`, pytest, ruff format/check, mypy, then `uv build` + `twine check` in the sub-project's own env.
- **`mypy.ini`**: excludes `openfeature-provider/.*` from the root mypy pass (the namespace tree is type-checked in its own env where `openfeature-sdk` is installed).
- **README**: link to the provider.

The new top-level `openfeature/` tree is isolated from the `posthog` wheel (explicit packages list) and from the `posthog.*`-scoped public-API snapshot — verified.

## Verification

- ✅ 20/20 tests pass
- ✅ ruff format + lint clean under both root (0.12.2) and sub-project (0.15.17)
- ✅ mypy clean (sub-project env + full root run, 127 files)
- ✅ `uv build` + `twine check` pass; wheel ships only the leaf package + `py.typed`, no clobbering `openfeature/__init__.py`, no tests
- ✅ no repo-wide regression: posthog wheel/sdist contain zero `openfeature/` entries; `public_api_check` snapshot up to date

## Follow-ups (not in this PR)

- **Confirm the PyPI distribution name** `openfeature-provider-posthog` before first publish. A third-party `posthog-openfeature-provider-python` (MPL-2.0, deprecated API, no object support) already exists on PyPI; this one is official, MIT, and uses the modern API with full object/JSON support.
- **Publish workflow**: the repo's release flow only ships `posthog`/`posthoganalytics`; this dist needs its own version/tag + PyPI publish job. No `posthoganalytics` twin needed.

## Docs

Documentation (install + usage snippets) lives in the PostHog docs, not in the package README (kept minimal so it doesn't drift). Companion docs PR: **PostHog/posthog.com#18006** — adds `/docs/feature-flags/installation/openfeature`.


openfeature-provider/openfeature/contrib/provider/posthog/provider.py
```diff
@@ -0,0 +1,265 @@
...
+        if not result.enabled:
+            return Reason.DISABLED
+        text = (result.reason or "").lower()
+        if "condition" in text or "match" in text or "variant" in text:
+            return Reason.TARGETING_MATCH
+        if "default" in text:
+            return Reason.DEFAULT
+        # Enabled, but no recognizable reason text: treat an absent reason as a
+        # targeting match, and an unfamiliar reason string as unknown.
+        return Reason.TARGETING_MATCH if result.reason is None else Reason.UNKNOWN
```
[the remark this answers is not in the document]
gustavohstrassburger: Fixed in 00ee43a. `_map_reason` no longer early-returns `DISABLED` on `not enabled`: it returns `TARGETING_MATCH` when the flag is enabled, `DEFAULT` when disabled (the common "no condition matched" case, since archived/missing flags surface as `FlagNotFoundError`), and `DISABLED` only when the reason text actually says the flag is off. The raw PostHog reason is also preserved in `flag_metadata["posthog_reason"]`.

openfeature-provider/tests/test_provider_unit.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+def test_float_variant_parse(fake_client):
+    fake_client.get_feature_flag_result.return_value = make_result(
+        enabled=True, variant="3.5"
+    )
+    details = _provider(fake_client).resolve_float_details(
+        "n", 0.0, EvaluationContext("u")
+    )
+    assert details.value == 3.5
```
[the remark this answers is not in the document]
gustavohstrassburger: Fixed in 00ee43a. The boolean reason-mapping cases (`test_boolean_targeting_match` + `test_boolean_disabled`) are collapsed into a parametrized `test_boolean_reason_mapping` (now also covering the `DEFAULT` case), and the number-parsing tests into `test_number_variant_parse` / `test_number_variant_parse_failure` with added cases (`"3"`→int/float, `None` variant).

openfeature-provider/openfeature/contrib/provider/posthog/provider.py
```diff
@@ -0,0 +1,265 @@
...
+    def initialize(self, evaluation_context: EvaluationContext) -> None:
+        # Preload locally-evaluated flag definitions only when the injected
+        # client is configured for local evaluation. We do not otherwise mutate
+        # the caller-owned client, and a preload failure must not make the
+        # OpenFeature client un-ready (remote evaluation still works).
+        if getattr(self._client, "personal_api_key", None):
+            try:
+                self._client.load_feature_flags()
+            except Exception:
+                pass
```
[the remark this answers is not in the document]
gustavohstrassburger: Fixed in 00ee43a. `initialize` now logs a `WARNING` (with `exc_info`) instead of swallowing the exception, so a misconfigured `personal_api_key` / unreachable host / missing permission is visible while still falling back to remote evaluation. Added `test_initialize_logs_warning_on_preload_failure` covering it.

gustavohstrassburger:
Tested this locally end-to-end against a running PostHog instance (not just the mocked unit/e2e tests) — registered the provider with the OpenFeature SDK and evaluated a real boolean flag through `client.get_boolean_details(...)`, and it worked. 🎉

Repro script I used:

```python
"""
Example demonstrating how to use the PostHog OpenFeature provider.

This example shows:
1. Initializing the PostHog client
2. Registering the PostHogProvider with the OpenFeature SDK
3. Evaluating a boolean feature flag via the OpenFeature API
"""

import os

import posthog
from openfeature import api
from openfeature.evaluation_context import EvaluationContext
from openfeature.contrib.provider.posthog import PostHogProvider

# Initialize the PostHog client
posthog_client = posthog.Posthog(
    project_api_key=os.getenv("POSTHOG_API_KEY", "phc_PIPWvLdxL4N9RUvpyENExMOFEz2jXuk5ehmyXFG3A2k"),
    host=os.getenv("POSTHOG_HOST", "http://localhost:8010"),
)

# Register the PostHog provider with OpenFeature
provider = PostHogProvider(posthog_client)
api.set_provider(provider)

# Get the OpenFeature client
client = api.get_client()

# The flag key to evaluate
flag_key = os.getenv("POSTHOG_FLAG_KEY", "my-flag")

# Evaluation context: targeting_key maps to PostHog's distinct_id
ctx = EvaluationContext(targeting_key="test-user")

# Evaluate the boolean flag
result = client.get_boolean_details(flag_key, False, ctx)

print(f"Flag: {flag_key}")
print(f"Value: {result.value}")
print(f"Reason: {result.reason}")
if result.error_message:
    print(f"Error: {result.error_message}")

posthog_client.shutdown()
```

(The `phc_` key above is a local-dev project key pointing at `localhost:8010`.) Confirms the full path works as installed: namespace import, provider registration, `targeting_key` → `distinct_id` mapping, and the boolean resolution returning the expected value/reason.

openfeature-provider/openfeature/contrib/provider/posthog/provider.py
```diff
@@ -0,0 +1,280 @@
...
+        )
+
+    def resolve_string_details(
+        self,
+        flag_key: str,
+        default_value: str,
+        evaluation_context: Optional[EvaluationContext] = None,
+    ) -> FlagResolutionDetails[str]:
+        result = self._resolve(flag_key, evaluation_context)
+        if result.variant is None:
```
haacked, line 154: `suggestion`: Unenrolled users get `TYPE_MISMATCH` errors when they shouldn't. When a user matches no condition on a multivariate flag, PostHog returns `enabled=False, variant=None`, so the `result.variant is None` check here raises `TypeMismatchError`. The OpenFeature SDK then returns the correct default value but marks it with `error_code=TYPE_MISMATCH` and `reason=ERROR`, so normal non-enrollment gets flagged as type-mismatch errors.

You can tell the two cases apart by `enabled`: when `enabled=True, variant=None`, it's a genuine mismatch (boolean flag read as string), while `enabled=False, variant=None` means the flag is off or nobody matched. Return the default with a non-error reason in that case:

```python
        result = self._resolve(flag_key, evaluation_context)
        if result.variant is None:
            if not result.enabled:
                return FlagResolutionDetails(
                    value=default_value,
                    reason=self._map_reason(result),
                    flag_metadata=self._flag_metadata(result),
                )
            raise TypeMismatchError(
                f"Flag '{flag_key}' has no string variant (boolean flag)."
            )
```

The same `variant is None` branch is in [`_resolve_number`](https://github.com/posthog/posthog-python/blob/e5809e1196f0ed706dc9fe86f4a3a2fa2c35e724/openfeature-provider/openfeature/contrib/provider/posthog/provider.py#L212); keep the non-numeric-variant case there raising `TypeMismatchError`. One tradeoff: a disabled boolean flag read as a string would reclassify from `TYPE_MISMATCH` to `DISABLED`, since `enabled=False, variant=None` can't tell the two apart.

openfeature-provider/openfeature/contrib/provider/posthog/provider.py
```diff
@@ -0,0 +1,280 @@
...
+    def _split_context(
+        evaluation_context: Optional[EvaluationContext],
+    ) -> tuple[dict[str, Any], dict[str, Any], dict[str, Any]]:
+        if evaluation_context is None or not evaluation_context.attributes:
+            return {}, {}, {}
+        attrs = evaluation_context.attributes
+        groups = attrs.get(GROUPS_KEY) or {}
+        group_properties = attrs.get(GROUP_PROPERTIES_KEY) or {}
+        person_properties = {k: v for k, v in attrs.items() if k not in _RESERVED_KEYS}
+        groups = groups if isinstance(groups, dict) else {}
```
haacked, line 257: `suggestion`: Nothing tests the non-dict fallback here. `test_context_split` only passes well-formed dicts, so when a caller sets `groups` to a string or list, the `isinstance(groups, dict)` guard that coerces it to `{}` is unverified. Drop that guard in a later refactor and a non-dict would flow straight into `get_feature_flag_result` with no test to catch it.

Add a case mirroring `test_context_split` that passes `groups="acme"` and asserts the forwarded `kwargs["groups"]` is `None`.

.github/workflows/ci.yml
```diff
@@ -256,6 +256,49 @@ jobs:
...
+
+        steps:
+            - uses: actions/checkout@85e6279cec87321a52edac9c87bce653a07cf6c2
+              with:
+                  fetch-depth: 1
+
+            - name: Set up Python 3.12
+              uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55
+              with:
+                  python-version: 3.12
```
haacked: `suggestion`: The package declares support for Python 3.10 through 3.14 (`requires-python` plus classifiers in `pyproject.toml`), but this job only tests on 3.12. mypy targets 3.10, so type errors are caught, but nothing exercises the floor or ceiling at runtime. A 3.10-only incompatibility or a 3.14 behavior change would ship unverified.

The `tests` and `import-check` jobs already matrix across all five versions.

```yaml
        strategy:
            matrix:
                python-version: ['3.10', '3.11', '3.12', '3.13', '3.14']
        steps:
            ...
            - name: Set up Python ${{ matrix.python-version }}
              uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55
              with:
                  python-version: ${{ matrix.python-version }}
```

openfeature-provider/openfeature/contrib/provider/posthog/provider.py
```diff
@@ -0,0 +1,280 @@
...
+    # -- typed resolvers ------------------------------------------------------
+
+    def resolve_boolean_details(
+        self,
+        flag_key: str,
+        default_value: bool,
+        evaluation_context: Optional[EvaluationContext] = None,
+    ) -> FlagResolutionDetails[bool]:
+        result = self._resolve(flag_key, evaluation_context)
+        return FlagResolutionDetails(
```
haacked: `nit`: The four typed resolvers end with the same `FlagResolutionDetails` tail except for `value`. A small helper keeps the `reason`/`flag_metadata` wiring in one place:

```python
    def _details(self, value: _T, result: FeatureFlagResult) -> FlagResolutionDetails[_T]:
        return FlagResolutionDetails(
            value=value,
            variant=result.variant,
            reason=self._map_reason(result),
            flag_metadata=self._flag_metadata(result),
        )
```

haacked (commented):
Clean, well-tested provider, nothing blocking. A few non-blocking suggestions inline, the correctness one (TYPE_MISMATCH on unmatched experiments) is the one worth a decision.

I'll leave approval to the client libraries team who may want a look.

openfeature-provider/README.md
```diff
@@ -0,0 +1,103 @@
+# PostHog provider for OpenFeature (Python)
```
marandaneto: i'd move all install/snippets to https://posthog.com/docs/feature-flags and just point to the docs
single source of truth for docs, readme gets outdated

marandaneto:
since its a new distribution, you'd need to adapt https://github.com/PostHog/posthog-python/blob/main/.github/workflows/release.yml to publish multiple packages
you can check ruby/rails, the js monorepo, etc, since they also publish multiple packages within the release process

openfeature-provider/pyproject.toml
```diff
@@ -0,0 +1,70 @@
...
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
+]
+dependencies = [
+    # FeatureFlagResult / get_feature_flag_result are the modern, non-deprecated
+    # single-call API. Our reason mapping tolerates an absent reason field.
+    "posthog>=6.0.0",
```
marandaneto: we need to make sure that all PRs build this from the local branch so we know we're not breaking compatibility between both versions since they depend on each other
i'd also recommend making a pinned dependency here

most likely you need something like:

   [tool.uv.sources]
   posthog = { path = "..", editable = true }

openfeature-provider/pyproject.toml
```diff
@@ -0,0 +1,70 @@
+[build-system]
+requires = ["setuptools>=61.0", "wheel"]
+build-backend = "setuptools.build_meta"
+
+[project]
+name = "openfeature-provider-posthog"
+version = "0.1.0"
```
marandaneto: needs to set up `sampo` here, see other comment related to the release process

marandaneto:
Optional nice-to-have: add a CI smoke test that installs the built wheel into a fresh env and imports openfeature.contrib.provider.posthog, since namespace packaging issues are easy to miss with source-tree tests.

marandaneto:
The unmatched-user/default behavior should include object flags as well?
     - Existing comment covers string/number variant is None.
     - resolve_object_details has the same issue when enabled=False and payload=None: it raises TypeMismatchError, so
       OpenFeature reports Reason.ERROR / TYPE_MISMATCH instead of a normal default.
     - See openfeature-provider/openfeature/contrib/provider/posthog/provider.py:196.
     - If they fix non-enrollment handling, apply it consistently to string/number/object and add detail tests checking
       no error_code.

marandaneto:
left a bunch of comments, main blocker is the release process and a few other comments/suggestions, Phil added good points as well

openfeature-provider/openfeature/contrib/provider/posthog/provider.py
gustavohstrassburger: Fixed in 8f24613. `enabled=False, variant=None` now returns the default with a normal reason (DEFAULT/DISABLED) instead of raising `TypeMismatchError`; the genuine boolean-read-as-string case (`enabled=True, variant=None`) still raises. Applied the same in `_resolve_number`, and (per Manoel) in `resolve_object_details`. Added detail tests asserting `error_code is None` on the non-enrollment path.

openfeature-provider/openfeature/contrib/provider/posthog/provider.py
gustavohstrassburger: Fixed in 8f24613. Added `test_context_split_non_dict_groups_coerced_to_none`, parametrized over `"acme"`, a list, and an int, asserting the forwarded `groups`/`group_properties` kwargs are `None`.

openfeature-provider/openfeature/contrib/provider/posthog/provider.py
gustavohstrassburger: Done in 8f24613 — extracted `_details(value, result)` and routed all four typed resolvers (and the default-return paths) through it.

.github/workflows/ci.yml
gustavohstrassburger: Agree the matrix should mirror the `tests`/`import-check` jobs (3.10–3.14). I do not edit `.github/workflows/**` in this automated review-fix turn (guardrail), so I have left this for a maintainer to apply directly.

openfeature-provider/README.md
gustavohstrassburger: Done in 8f24613 — slimmed the README to a minimal quickstart and pointed to https://posthog.com/docs/feature-flags as the single source of truth. Kept only the OpenFeature-specific registration snippet + context mapping (not yet on the docs site); those can move there once the provider is documented.

openfeature-provider/pyproject.toml
gustavohstrassburger: The local-branch build is already wired: `[tool.uv.sources] posthog = { path = "..", editable = true }` is in this pyproject, so the provider CI job builds `posthog` from the branch. On pinning: I kept `posthog>=6.0.0` rather than an exact pin, since exact-pinning a library runtime dep causes resolution conflicts for downstream consumers. Happy to raise the floor (e.g. to the current major) if you prefer a tighter bound.

openfeature-provider/pyproject.toml
gustavohstrassburger: The release wiring (Sampo for the new package + adapting `release.yml` to publish a second distribution) is a `.github/workflows/**` + release-config change I will not make in this automated turn. Flagged for a maintainer in a summary comment on the PR.

gustavohstrassburger: Implemented in 5236b4b. Sampo already auto-discovers the package as `pypi/openfeature-provider-posthog` (it has its own pyproject with name+version), and `release.yml` now builds/publishes/tags it — gated so it only fires when a changeset bumps the provider, leaving posthog-only releases untouched. Added `openfeature-provider/CHANGELOG.md` for Sampo to maintain. First publish still needs a one-time PyPI trusted-publisher registration for the new project.

openfeature-provider/pyproject.toml
gustavohstrassburger: Pinned in 5236b4b: `posthog>=7.0.0,<8.0.0` (the tested major). The local-branch build via `[tool.uv.sources] posthog = { path = "..", editable = true }` is already in place, so PRs build against the branch.

openfeature-provider/README.md
gustavohstrassburger: Done earlier in 8f24613 — README slimmed to a minimal quickstart pointing at https://posthog.com/docs/feature-flags as the source of truth.

gustavohstrassburger:
@marandaneto — implemented your remaining suggestions in `5236b4b`:

- **Release (`release.yml`)** now publishes `openfeature-provider-posthog` as a second distribution. Sampo already auto-discovers it as `pypi/openfeature-provider-posthog` (own pyproject with name + version), and `short_tags = "posthog"` means it tags as `openfeature-provider-posthog-v{version}`. The new build/publish/tag steps are **gated** — they only run when this release actually bumped the provider (a changeset targeting it), so `posthog`-only releases skip them, and they run **after** the posthog publish/tag/release so a provider issue can never block the core release.
- **Sampo**: added `openfeature-provider/CHANGELOG.md` for it to maintain.
- **CI smoke test**: the provider job now installs the built wheel into a clean env and imports `openfeature.contrib.provider.posthog` — catches namespace-packaging regressions source-tree tests miss.
- **Pin**: `posthog>=7.0.0,<8.0.0` (tested major); local-branch build via `[tool.uv.sources]` was already in place.

**One manual prerequisite for the first provider release** (can't be done in a PR): register a PyPI **trusted publisher** for `openfeature-provider-posthog` pointing at `release.yml` + the `Release` environment, exactly like posthog/posthoganalytics. After that, adding a changeset that bumps the provider triggers its first publish.

Verified locally: 37 tests, ruff, mypy, `uv build` + clean-env wheel install/import all pass; both workflow YAMLs parse. The release flow itself can only be fully exercised on a real release run.

marandaneto:
@gustavohstrassburger 

Still not fixed / still worth commenting:

 1. Provider release/versioning is still broken
     - Release workflow assumes Sampo can process changesets for pypi/openfeature-provider-posthog.
     - But Sampo does not know that package. I verified in a temp worktree:

    ```bash
      sampo add -p pypi/openfeature-provider-posthog -b patch -m test
      # Failed: Package 'pypi/openfeature-provider-posthog' not found in the workspace.
    ```

    So openfeature-provider/pyproject.toml won’t be bumped by Sampo, and the release workflow’s detection at
    .github/workflows/release.yml:245 likely never triggers for this package.

 2. Provider CI is still only Python 3.12
     - openfeature-provider/pyproject.toml claims Python 3.10–3.14 support.
     - .github/workflows/ci.yml:268-271 still only sets up 3.12.

those are still open, the 2. is optional but i think its cool to test across all major versions like we do for the main package

openfeature-provider/pyproject.toml
gustavohstrassburger: Correction to my earlier reply — you were right, Sampo does **not** auto-discover this package. I checked `sampo-core`'s pip adapter: it only discovers packages from the root pyproject's `[tool.uv.workspace] members` (plus the root). Fixed properly in b40251e by declaring a uv workspace (`members = ["openfeature-provider"]`) so `sampo add -p pypi/openfeature-provider-posthog` works and `sampo release` bumps the provider. Details in the PR comment.

.github/workflows/release.yml
```diff
@@ -227,6 +227,53 @@ jobs:
...
+      - name: Tag openfeature-provider release
+        if: steps.of-provider.outputs.released == 'true'
+        env:
+          GH_TOKEN: ${{ steps.releaser.outputs.token }}
+          PROVIDER_VERSION: ${{ steps.of-provider.outputs.version }}
+          COMMIT_HASH: ${{ steps.commit-release.outputs.commit-hash }}
+        run: |
+          gh api "repos/${{ github.repository }}/git/refs" \
+            -f "ref=refs/tags/openfeature-provider-posthog-v${PROVIDER_VERSION}" \
+            -f "sha=${COMMIT_HASH}"
```
marandaneto: i think a build matrix is better as suggested before
for example https://github.com/PostHog/posthog-ruby/blob/4579a7f2c1bb253bf67cb560f14a72190118ff4f/.github/workflows/release.yml#L222-L233
those are also N packages

openfeature-provider/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Development
+
+```bash
+cd openfeature-provider
+uv sync --extra dev
+uv run pytest
+uv run ruff check .
+uv run mypy .
+```
```
marandaneto: move to https://github.com/PostHog/posthog-python/blob/main/CONTRIBUTING.md or create its own CONTRIBUTING

openfeature-provider/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+of_client = api.get_client()
+ctx = EvaluationContext(targeting_key="user-123", attributes={"plan": "pro"})
+enabled = of_client.get_boolean_value("my-flag", False, ctx)
+```
+
+The OpenFeature `targeting_key` maps to PostHog's `distinct_id`; other context
+attributes become person properties, with reserved keys `groups` and
+`group_properties` mapping to PostHog groups. You own the `Posthog` client
+lifecycle — call `client.shutdown()` when done.
```
marandaneto: this gets outdated super fast, lets point to our docs only, you can link to the specific open feature section instead

gustavohstrassburger:
## Review feedback — all addressed ✅

Thanks @marandaneto and @haacked (and Greptile) for the thorough review. Summary of what changed and how.

### Greptile
- **Reason mapping** (`DISABLED` returned for every disabled flag) → now maps a not-enabled/unmatched result to `Reason.DEFAULT`, reserving `DISABLED` for when the reason text actually says the flag is off.
- **Silent `initialize` swallow** → a failed `load_feature_flags()` now logs a `WARNING` (with `exc_info`) instead of `pass`, so a misconfigured `personal_api_key`/host is visible while still falling back to remote eval.
- **Repeated tests** → collapsed into `@pytest.mark.parametrize`.

### @haacked
- **Unenrolled users got `TYPE_MISMATCH`** → string/integer/float/**object** resolvers now return the caller's default with a normal reason when `enabled=False` (no condition matched / flag off); only a genuine mismatch (`enabled=True` but uncoercible) raises. Added detail tests asserting `error_code is None` on the non-enrollment path.
- **`_details` nit** → extracted a shared helper for the `reason`/`flag_metadata` wiring across the typed resolvers.
- **Non-dict `groups` untested** → added a coercion test (non-dict `groups`/`group_properties` forwarded as `None`).
- **CI only on 3.12** → the provider CI job now matrixes Python **3.10–3.14**, mirroring the main package.

### @marandaneto
- **"Set up Sampo" / `sampo add ... not found`** → root `pyproject.toml` now declares a uv workspace (`[tool.uv.workspace] members = ["openfeature-provider"]`), so Sampo discovers `pypi/openfeature-provider-posthog`. The root `uv.lock` delta is purely additive (member + `openfeature-sdk`); the main `posthog` build is unaffected.
- **Build from the local branch** → the provider resolves posthog via `{ workspace = true }`; added `openfeature-provider/CHANGELOG.md` for Sampo.
- **Publish N packages / use a build matrix** → `release.yml` restructured into an approval-gated `version-bump` job + a no-environment `publish` **matrix** over `posthog`, `posthoganalytics`, and `openfeature-provider-posthog` (`fail-fast`, `max-parallel: 1`), each gated on whether its version changed. Single approval preserved.
- **Wheel-install smoke test** → CI installs the built wheel into a clean env and imports `openfeature.contrib.provider.posthog`.
- **README drifts / point to docs only** → the separate provider README is **removed entirely**; the package (`[project.urls] Documentation`) and the root README point straight to the docs. Docs content lives in **PostHog/posthog.com#18006** (`/docs/feature-flags/installation/openfeature`).
- **Dev instructions** → moved to `openfeature-provider/CONTRIBUTING.md`.
- **Pin posthog** → `posthog>=7.0.0,<8.0.0`.

### Status
- 37 provider tests pass; ruff/mypy clean; provider CI matrix (3.10–3.14), build + wheel smoke test green.
- The branch was rebased onto current `main` (7.21.1 + `posthog.mcp`), which also resolved a merge conflict that had been blocking CI.

### One follow-up for maintainers (not blocking this PR)
The `publish` matrix job runs without a deployment environment (so the matrix doesn't multiply the approval prompt — approval stays on `version-bump`). For OIDC to work, each package's PyPI **trusted publisher** should point at the `publish` job rather than the `Release` environment — a one-time PyPI-side config.

.github/workflows/release.yml
```diff
@@ -249,14 +213,162 @@ jobs:
...
+    strategy:
+      fail-fast: true
+      max-parallel: 1
+      matrix:
+        package:
+          - name: posthog
+            version_file: pyproject.toml
+            build: uv run make build_release
+            packages_dir: dist
+            tag_prefix: "v"
```
marandaneto:
```suggestion
            tag_prefix: "posthog-v"
```

marandaneto: i think it makes sense like this now

.github/workflows/release.yml
```diff
@@ -249,14 +213,162 @@ jobs:
...
+            changelog: CHANGELOG.md
```
marandaneto, line 246: i think we should move this to posthog/CHANGELOG.md now
and the root changelog points to the inner changelogs
similar to https://github.com/PostHog/posthog-ruby/blob/main/CHANGELOG.md

.github/workflows/release.yml
```diff
@@ -249,14 +213,162 @@ jobs:
...
+
+      - name: Create ${{ matrix.package.name }} GitHub Release
+        if: steps.detect.outputs.has-new-version == 'true' && matrix.package.github_release
+        env:
+          GH_TOKEN: ${{ steps.releaser.outputs.token }}
+          TAG: ${{ matrix.package.tag_prefix }}${{ steps.detect.outputs.version }}
+          CHANGELOG_FILE: ${{ matrix.package.changelog }}
+        run: |
+          CHANGELOG_ENTRY=$(awk -v defText="see ${CHANGELOG_FILE}" '/^## /{if (flag) exit; flag=1; next} flag; END{if (!flag) print defText}' "${CHANGELOG_FILE}" | sed '/[^[:space:]]/,$!d' | tac | sed '/[^[:space:]]/,$!d' | tac)
+          gh release create "$TAG" --notes "$CHANGELOG_ENTRY"
```
marandaneto: here

openfeature-provider/CHANGELOG.md
```diff
@@ -0,0 +1,5 @@
+# Changelog
```
marandaneto:
```suggestion
# openfeature-provider-posthog
```

make sure the https://github.com/PostHog/posthog-python/pull/695/changes#r3498456802 works here since its fragile

marandaneto (approved):
left a few comments still, but approving to unblock

marandaneto:
probably requires setting up pypi for the new package (trusted publisher)

PR #700, opened 2026-06-27
fix: fall back when payload compression fails

marandaneto:

When `gzip=True`, local gzip failures should not prevent the SDK from sending the batch. The SDK should fall back to the original uncompressed JSON payload without a `Content-Encoding` header.

This catches local gzip errors during request construction and keeps the upload path uncompressed. A unit test covers the fallback.

- `uv run --extra test pytest posthog/test/test_request.py -q`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented as part of a cross-SDK consistency pass for client-side compression failure fallback. Kept the change scoped to request construction and added a focused regression test.

### DIFF
--- a/posthog/request.py
+++ b/posthog/request.py
@@ -240,13 +240,16 @@ def post(
     log.debug("making request: %s to url: %s", data, url)
     headers = {"Content-Type": "application/json", "User-Agent": USER_AGENT}
     if gzip:
-        headers["Content-Encoding"] = "gzip"
-        buf = BytesIO()
-        with GzipFile(fileobj=buf, mode="w") as gz:
-            # 'data' was produced by json.dumps(),
-            # whose default encoding is utf-8.
-            gz.write(cast(str, data).encode("utf-8"))
-        data = buf.getvalue()
+        try:
+            buf = BytesIO()
+            with GzipFile(fileobj=buf, mode="w") as gz:
+                # 'data' was produced by json.dumps(),
+                # whose default encoding is utf-8.
+                gz.write(cast(str, data).encode("utf-8"))
+            data = buf.getvalue()
+            headers["Content-Encoding"] = "gzip"
+        except OSError as exc:
+            log.warning("failed to gzip request body, sending uncompressed: %s", exc)
 
     res = (session or _get_session()).post(
         url, data=data, headers=headers, timeout=timeout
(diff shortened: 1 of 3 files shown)


posthog/request.py
```diff
@@ -240,13 +240,16 @@ def post(
...
+        try:
+            buf = BytesIO()
+            with GzipFile(fileobj=buf, mode="w") as gz:
+                # 'data' was produced by json.dumps(),
+                # whose default encoding is utf-8.
+                gz.write(cast(str, data).encode("utf-8"))
+            data = buf.getvalue()
+            headers["Content-Encoding"] = "gzip"
+        except OSError as exc:
+            log.warning("failed to gzip request body, sending uncompressed: %s", exc)
```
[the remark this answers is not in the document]
marandaneto: Fixed in b42076f — broadened the compression-failure catch so zlib.error also falls back to uncompressed.

posthog/test/test_request.py
```diff
@@ -179,6 +179,27 @@ def test_post_sends_bytes_payload_with_gzip(self):
...
+                path="/batch/",
+                gzip=True,
+                session=mock_session,
+                batch=[],
+            )
+
+        data = mock_session.post.call_args.kwargs["data"]
+        headers = mock_session.post.call_args.kwargs["headers"]
+        self.assertIsInstance(data, str)
+        self.assertNotIn("Content-Encoding", headers)
```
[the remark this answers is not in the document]
marandaneto: Fixed in b42076f — expanded the fallback test to cover both OSError and zlib.error via subTest.

posthog/request.py
```diff
@@ -240,13 +240,16 @@ def post(
...
-        data = buf.getvalue()
+        try:
+            buf = BytesIO()
+            with GzipFile(fileobj=buf, mode="w") as gz:
+                # 'data' was produced by json.dumps(),
+                # whose default encoding is utf-8.
+                gz.write(cast(str, data).encode("utf-8"))
+            data = buf.getvalue()
+            headers["Content-Encoding"] = "gzip"
+        except Exception as exc:
```
ioannisj: should we narrow this down?

ioannisj (approved):

posthog/request.py
marandaneto: i dont see how it'd throw any other error if not compressing related but we could

marandaneto: Good catch, narrowed the gzip fallback to OSError and zlib.error in fb61e42.

PR #702, opened 2026-06-27
feat(capture): v1 wire serialization transforms (capture v1, 2/6)

eli-r-ph:

Second PR in the stacked **Capture V1** series (stacked on #701). Adds the pure, no-I/O serialization layer that turns a legacy-shaped queued message into a `/i/v1/analytics/events` wire event. Still inert — nothing calls these functions yet (transport + wiring come in later PRs).

New module `posthog/capture_v1.py`:

- **`to_v1_event(msg)`** — dict-to-dict transform proven against the server contract (`rust/capture/src/v1/analytics/types.rs`) and posthog-go's `capture_v1.go`:
  - Lifts the sentinel properties into the typed `options` object, including the `$ignore_sent_at` -> `disable_skew_correction` rename.
  - Promotes `$session_id`/`$window_id` to top-level string fields.
  - **Relocates top-level `$set`/`$set_once` into `properties`** — v1 has no top-level form, and the legacy `set()`/`set_once()` builders emit them at the top level, so without this person-property updates would silently vanish.
  - Strips `$lib`/`$lib_version` (the server injects them from the required `PostHog-Sdk-Info` header).
  - **Coerces options to native JSON types or omits them.** v1 deserializes `options` strictly; a wrong type (e.g. `cookieless_mode: "true"`) 400s the *entire batch*, so a bad value is dropped rather than forwarded.
  - Pure: the input message is not mutated (safe for retries/callbacks).
- **`build_v1_batch_body(events, historical_migration)`** — the `api_key`/`sent_at`-free envelope with a tz-aware RFC3339 `created_at` (`historical_migration` omitted when false).
- Shared constants: endpoint path, required header names, result codes, retryable/terminal status sets (used by the transport PR).

New `posthog/test/test_capture_v1.py` (47 cases, parameterized): bool/string coercion across all variants; per-sentinel lift + rename + coercion; bad-coercion omitted-but-removed; top-level string sentinels; `options == {}` when empty; `$lib` stripping; non-wire keys (e.g. `type`) not leaked; input-not-mutated; top-level `$set`/`$set_once` relocation incl. merge-precedence with an existing `properties.$set`; `$groups` left untouched; tz-aware/passthrough/default timestamps; envelope shape + RFC3339 `created_at` + `historical_migration` toggling.

`ruff format`/`check` clean; `mypy` clean on the new module; regenerated `references/public_api_snapshot.txt`.

## :pencil: Checklist

- [x] No breaking change (additive; transforms are not yet wired in).

## 🤖 Agent context

Authored with Cursor (Claude Opus 4.8) per the agreed plan. Architecture note: posthog-go builds v1 events from typed message structs, but posthog-python's queue holds already-assembled legacy dicts, so this is a dict-to-dict transform applied at send time — the wire shape and sentinel table are identical to go's. The strict-type/coerce-or-omit and `$set` relocation behaviors were verified directly against the Rust `Event`/`Options` structs and their serde tests.

### DIFF
--- a/references/public_api_snapshot.txt
+++ b/references/public_api_snapshot.txt
@@ -477,17 +477,6 @@ attribute posthog.capture_exception_code_variables = False
 attribute posthog.capture_mode.CAPTURE_MODE_ENV_VAR = 'POSTHOG_CAPTURE_MODE'
 attribute posthog.capture_mode.CaptureMode.V0 = 'v0'
 attribute posthog.capture_mode.CaptureMode.V1 = 'v1'
-attribute posthog.capture_v1.CAPTURE_V1_PATH = '/i/v1/analytics/events'
-attribute posthog.capture_v1.HEADER_ATTEMPT = 'PostHog-Attempt'
-attribute posthog.capture_v1.HEADER_REQUEST_ID = 'PostHog-Request-Id'
-attribute posthog.capture_v1.HEADER_REQUEST_TIMESTAMP = 'PostHog-Request-Timestamp'
-attribute posthog.capture_v1.HEADER_SDK_INFO = 'PostHog-Sdk-Info'
-attribute posthog.capture_v1.RESULT_DROP = 'drop'
-attribute posthog.capture_v1.RESULT_OK = 'ok'
-attribute posthog.capture_v1.RESULT_RETRY = 'retry'
-attribute posthog.capture_v1.RESULT_WARNING = 'warning'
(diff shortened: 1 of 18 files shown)


references/public_api_snapshot.txt
```diff
@@ -1309,6 +1322,7 @@ module posthog.ai.utils
 module posthog.args
 module posthog.bucketed_rate_limiter
 module posthog.capture_mode
+module posthog.capture_v1
```
marandaneto: most likely we can make all of this internal? eg _methodName _constant etc

marandaneto (approved):

PR #703, opened 2026-06-27
feat(capture): v1 transport + partial-retry send loop (capture v1, 3/6)

eli-r-ph:

Third PR in the stacked **Capture V1** series (stacked on #702). Adds the HTTP transport and partial-retry send loop for `POST /i/v1/analytics/events`, on top of the pure transforms from #702. Still inert — `_send_v1_batch` has no caller until the consumer-wiring PR.

New in `posthog/capture_v1.py` (transport helpers are private; only `CaptureV1Error` is public):

- **`_post_v1(...)`** — a single attempt. Bearer auth (no `api_key` in the body), the required v1 headers (`PostHog-Sdk-Info`, `PostHog-Attempt`, `PostHog-Request-Id`, `PostHog-Request-Timestamp`), and optional body compression. Returns the raw response; this is also the monkeypatch seam the test harness adapter will drive.
- **`_parse_v1_response(...)`** — classifies one response without raising: 2xx parses the per-uuid `results` map (an unparseable 2xx body is flagged `malformed`), non-2xx best-effort extracts an error message, and `Retry-After` (delta-seconds or HTTP-date) is parsed in both cases.
- **`_send_v1_batch(...)`** — the v1 sibling of `Consumer._send`. Loops up to `max_retries + 1` attempts, but **shrinks the batch to only the events the server tagged `retry`** after each 2xx. `ok`/`warning`/absent events succeed silently; `drop` events are logged (a request the server accepted-but-dropped is not a delivery failure, so it is not raised). Raises `CaptureV1Error` (an `APIError` subclass, so the consumer's existing `on_error(exc, batch)` keeps working) on a batch-level terminal/transport failure or once retries are exhausted.
- **`CaptureV1Error`** (public).

Also adds `posthog/capture_compression.py`: `CaptureCompression` enum (`none`/`gzip`/`deflate`/`zstd`) with internal `_resolve_capture_compression`. ZSTD needs the optional `posthog[zstd]` extra (`zstandard` package); explicit kwarg without it raises, env var without it warns and falls back.

Behavior choices, verified against the Rust contract and posthog-go's `capture_v1_send.go`:

- **429 is terminal in v1** (v0 retries it) — the backend signals overload via retryable 5xx + `Retry-After`.
- **A stable `PostHog-Request-Id` and `created_at` span all attempts**; `PostHog-Attempt` increments — so the backend can correlate/dedupe a retried batch.
- **A 2xx with an unparseable body is terminal** — re-sending against a broken success would loop forever.
- Backoff reuses the v0 policy (`Retry-After` wins, else capped exponential) so both wire protocols back off identically.

Also factors a shared `gzip_compress` helper out of `request.post` — pure refactor, no v0 behavior change (covered by the existing `test_request.py`).

`posthog/test/test_capture_v1.py` now has 74 cases (47 transform + 27 transport). New transport coverage: `post_v1` header/url/no-key-in-body/gzip-magic; `parse_v1_response` success/malformed/missing-results/error-body-variants/text-fallback/`Retry-After`; and `send_v1_batch` driven by a stubbed `post_v1` with mocked sleeps — all-ok, absent-uuid-accepted, partial-retry-shrinks-to-retry-uuids, stable-request-id + incrementing-attempt, drop-logged-not-raised, retry-exhausted-raises, malformed-2xx-terminal, 400/429-terminal-not-retried, 503-then-success (honors `Retry-After`), 503-exhausted-raises, transport-error-then-success, transport-error-exhausted-reraises.

`ruff format`/`check` clean; `mypy` clean on `capture_v1.py` + `request.py`; `test_request.py` (61) still green after the gzip refactor; regenerated `references/public_api_snapshot.txt`.

## :pencil: Checklist

- [x] No breaking change (additive; `send_v1_batch` has no caller yet).

## 🤖 Agent context

Authored with Cursor (Claude Opus 4.8) per the agreed plan. posthog-go's `sendV1` is woven into its client lifecycle (channels, `notifyFailure`/`notifySuccess`, `maxAttempts`); this port instead fits posthog-python's `Consumer._send` shape — a synchronous loop that raises on failure so the existing `upload()` -> `on_error(exc, batch)` path fires unchanged — while preserving go's partial-retry algorithm, status matrix, stable-request-id semantics, and per-event drop/retry handling.

### DIFF
--- a/posthog/capture_compression.py
+++ b/posthog/capture_compression.py
@@ -105,8 +105,8 @@ def _resolve_capture_compression(
     if raw is None or raw.strip() == "":
         return fallback
 
-    resolved = _ALIASES.get(raw.strip().lower())
-    if resolved is None:
+    env_resolved = _ALIASES.get(raw.strip().lower())
+    if env_resolved is None:
         log.warning(
             "Unrecognized %s=%r; falling back to %s. Expected one of %s.",
             CAPTURE_COMPRESSION_ENV_VAR,
@@ -115,7 +115,7 @@ def _resolve_capture_compression(
             sorted(_ALIASES),
         )
         return fallback
-    if resolved is CaptureCompression.ZSTD and not _zstd_available():
+    if env_resolved is CaptureCompression.ZSTD and not _zstd_available():
         log.warning(
             "%s=%r requires the zstandard package (install posthog[zstd]); "
             "falling back to %s.",
@@ -124,4 +124,4 @@ def _resolve_capture_compression(
             fallback.value,
         )
         return fallback
-    return resolved
+    return env_resolved
--- a/posthog/capture_compression.py
+++ b/posthog/capture_compression.py
@@ -1,0 +1,90 @@
+import logging
+import os
+from enum import Enum
+from typing import Optional, Union
+
+log = logging.getLogger("posthog")
+
+CAPTURE_COMPRESSION_ENV_VAR = "POSTHOG_CAPTURE_COMPRESSION"
+
+
+class CaptureCompression(str, Enum):
+    """Selects the request-body compression for capture-v1 uploads.
+
+    Only honored when ``capture_mode`` is ``V1``; the legacy ``/batch/`` path
+    keeps using its own ``gzip`` flag. ``NONE`` sends the body uncompressed.
+    ``GZIP`` and ``DEFLATE`` (zlib, RFC 1950) are both stdlib / zero-dependency
+    and map to the matching ``Content-Encoding`` token the v1 server decodes
+    (``br``/``zstd`` are accepted by the server too but need extra dependencies,
+    so they are intentionally left out for now). Inheriting from ``str`` keeps
+    the members comparable to and serializable as their token values.
+    """
+
+    NONE = "none"
+    GZIP = "gzip"
+    DEFLATE = "deflate"
+
+
+# Accepted spellings for both the kwarg and the env var. ``identity`` mirrors
+# the HTTP token for "no encoding".
+_ALIASES: dict[str, CaptureCompression] = {
+    "none": CaptureCompression.NONE,
+    "identity": CaptureCompression.NONE,
+    "gzip": CaptureCompression.GZIP,
+    "deflate": CaptureCompression.DEFLATE,
+}
+
+
+def _coerce_explicit(
+    value: Union[CaptureCompression, str],
+) -> CaptureCompression:
+    """Normalize an explicitly-supplied compression to a ``CaptureCompression``.
+
+    An explicit but unrecognized value is a programming error, so it raises
+    ``ValueError`` rather than silently defaulting (unlike the env var, which is
+    operator-supplied and defaults defensively).
+    """
+    if isinstance(value, CaptureCompression):
+        return value
+    if isinstance(value, str):
+        resolved = _ALIASES.get(value.strip().lower())
+        if resolved is not None:
+            return resolved
+    raise ValueError(
+        f"invalid capture_compression {value!r}; expected a CaptureCompression "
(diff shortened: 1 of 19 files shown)


marandaneto:
2 things here

1. drop results are silently treated as success

 posthog/capture_v1.py:510-526

 Python collects drop results, but if there are no retry events it returns successfully and never surfaces the dropped
 event. If a batch has both drop and retry, and the retry later succeeds, the earlier drop is also lost.

 Go does this differently: posthog-go/capture_v1_send.go:173-174 calls failure callback immediately for drop.

 Rust also surfaces 2xx drop/final retry outcomes through on_error: posthog-rs/src/client/transport.rs:835-850.

 This means Python users won’t be told that the backend explicitly rejected an event.


2. Retry-After replaces backoff instead of acting as a minimum

 posthog/capture_v1.py:404-413

 Python sleeps exactly Retry-After when present. Go/Rust treat Retry-After as a minimum and wait for max(configured_backoff, retry_after):

 - Go: posthog-go/capture_v1_send.go:183-188
 - Rust: posthog-rs/src/client/retry.rs:102-107

 With a small Retry-After, Python retries earlier than the normal exponential backoff, which drifts from the canonical
 implementations.

eli-r-ph:
👋 looked into it, PR parity issues are addressed, thanks for the callout! Also surfaced a couple of other small parity Issues across the 3 SDKs (some extending into prior defaults that effect v1, not v1-specific stuff) so opening a couple other tiny PRs to try to normalize so these things can be gated/smoke tested in a backend SDK-agnostic way as the porting continues 👍

eli-r-ph:
👋 update on this:
* All parity issues addressed in this stack now ✅ 
* This also prompted a couple of spot changes in the Rust and Go SDKs to tighten up the parity:
    * Unified Retry-After and max retry attempt time clamp (default 30s since 2 of the 3 SDKs already agree on this)
    * When this can be overridden by users in an SDK, it's a default; otherwise hardcoded
    * Ensured retry backoff calc logic is in parity across Go & Rust SDKs + this PR series (when not completely overridable as in Go SDK, where this is just a default again) 

Spot changes to those SDKs:
https://github.com/PostHog/posthog-go/pull/255
https://github.com/PostHog/posthog-go/pull/256
https://github.com/PostHog/posthog-rs/pull/177

posthog/capture_compression.py
```diff
@@ -0,0 +1,90 @@
...
+        resolved = _ALIASES.get(value.strip().lower())
+        if resolved is not None:
+            return resolved
+    raise ValueError(
+        f"invalid capture_compression {value!r}; expected a CaptureCompression "
+        f"or one of {sorted(_ALIASES)}"
+    )
+
+
+def resolve_capture_compression(
```
marandaneto: can we make this internal? eg _resolve_capture_compression

marandaneto (approved):

posthog/capture_compression.py
```diff
@@ -0,0 +1,90 @@
...
+import os
+from enum import Enum
+from typing import Optional, Union
+
+log = logging.getLogger("posthog")
+
+CAPTURE_COMPRESSION_ENV_VAR = "POSTHOG_CAPTURE_COMPRESSION"
+
+
+class CaptureCompression(str, Enum):
```
robbie-c, line 21: May want to add `zstd` to this list, it's quite a bit faster/better than `gzip` or `zlib`/`deflate`, and is available for free on some platforms (it is on node, not sure about python I'm afraid!)

eli-r-ph: I'm rolling that in now, good call. Will shim slightly b/c this will be an external lib until Python 3.14 when it lands in stdlib 👍 (this is why I skipped initially)

PR #705, opened 2026-06-27
chore(capture): v1 compliance adapter + CI job (capture v1, 5/6)

eli-r-ph:

Fifth PR in the stacked **Capture V1** series (stacked on #704). This is the one that proves the v1 implementation against the cross-SDK conformance harness: it teaches the compliance adapter to speak capture-v1 and adds a CI job that runs the harness `capture_v1` suite against it. Mirrors the posthog-go adapter at the same harness pin.

Adapter (`sdk_compliance_adapter/adapter.py`):

- **Protocol selected by `CAPTURE_MODE`** (baked into a new `Dockerfile.v1`). One adapter process speaks one mode: `/health` advertises `capture_v1` vs `capture_v0`, and `/init` creates the `Client` with the matching `capture_mode`. The harness uses the advertised capability to pick which suite to run.
- **`/init` now accepts `disable_geoip` and `historical_migration`.**
- **`/capture` folds the harness's v1 `options` back into the SDK's `$`-prefixed sentinel properties** (`cookieless_mode` -> `$cookieless_mode`, `disable_skew_correction` -> `$ignore_sent_at`, etc.), so the SDK lifts them onto the wire `options` object — exercising the #702 transform end to end.
- **A patched `_post_v1` records v1 attempts for `/state`** (retry attempt read from the call, only terminal/non-`retry` per-event results counted as sent), mirroring the existing `patched_batch_post`.

CI (`.github/workflows/sdk-compliance.yml`):

- **Bumps the harness pin** to **`0.10.0` (`02c049e`)** on both jobs (rebased onto main).
- **Splits into two jobs**: `compliance` (v0, `Dockerfile`) and `compliance-v1` (v1, `Dockerfile.v1`), each with its own report artifact.
- `docker-compose.yml` gains a `sdk-adapter-v1` service for local v1 runs.

`ruff format`/`check` clean; `py_compile` clean on the adapter. The v0 path is behavior-preserving — `capture_mode="v0"` resolves identically to the prior implicit default, and `disable_geoip`/`historical_migration` keep their previous effective defaults — so the existing v0 suite should pass unchanged at the new pin.

The harness integration itself is validated by the two CI jobs this PR adds (the multi-container harness orchestration is exactly what the reusable workflow runs per job). The adapter is a direct port of the posthog-go v1 adapter, which already passes the `capture_v1` suite at this same pin.

## :pencil: Checklist

- [x] No breaking change (adapter/CI only; v0 path preserved).

## 🤖 Agent context

Authored with Cursor (Claude Opus 4.8) per the agreed plan. The v1 suite asserts against the mock server's recorded requests rather than `/state`, so the adapter's `/state` tracking for v1 is best-effort parity with v0 (and with posthog-go) rather than load-bearing for the v1 assertions. Local harness runs were not performed — the added CI jobs are the gate, matching how the harness is intended to be exercised.

### DIFF
--- a/.github/workflows/sdk-compliance.yml
+++ b/.github/workflows/sdk-compliance.yml
@@ -13,9 +13,19 @@ on:
 
 jobs:
   compliance:
-    name: PostHog SDK compliance tests
+    name: PostHog SDK compliance tests (capture v0)
     uses: PostHog/posthog-sdk-test-harness/.github/workflows/test-sdk-action.yml@02c049e529001d02f37a534745678e057d371fb0
     with:
       adapter-dockerfile: "sdk_compliance_adapter/Dockerfile"
       adapter-context: "."
       test-harness-version: "0.10.0"
+      report-name: "sdk-compliance-report-v0"
+
+  compliance-v1:
+    name: PostHog SDK compliance tests (capture v1)
+    uses: PostHog/posthog-sdk-test-harness/.github/workflows/test-sdk-action.yml@02c049e529001d02f37a534745678e057d371fb0
+    with:
+      adapter-dockerfile: "sdk_compliance_adapter/Dockerfile.v1"
+      adapter-context: "."
+      test-harness-version: "0.10.0"
+      report-name: "sdk-compliance-report-v1"
--- a/.sampo/changesets/capture-v1-mode.md
+++ b/.sampo/changesets/capture-v1-mode.md
@@ -1,0 +1,9 @@
+---
+pypi/posthog: minor
+---
+
+Add an opt-in `capture_mode` for the Capture V1 ingestion protocol (`POST /i/v1/analytics/events`). Set `capture_mode="v1"` on the client (or the `POSTHOG_CAPTURE_MODE=v1` environment variable) to use Bearer auth, per-event results, and partial retry. Defaults to `"v0"` (the legacy `/batch/` endpoint), so existing setups are unaffected.
+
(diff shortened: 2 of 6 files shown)


.github/workflows/sdk-compliance.yml
```diff
@@ -13,9 +13,21 @@ on:
...
+      test-harness-version: "0.8.0"
+      report-name: "sdk-compliance-report-v0"
+
+  compliance-v1:
+    name: PostHog SDK compliance tests (capture v1)
+    uses: PostHog/posthog-sdk-test-harness/.github/workflows/test-sdk-action.yml@be8b8d5a3f94a249659844e94832e874f049c1e4
+    with:
+      adapter-dockerfile: "sdk_compliance_adapter/Dockerfile.v1"
+      adapter-context: "."
+      test-harness-version: "0.8.0"
```
marandaneto: update to latest version
https://github.com/PostHog/posthog-sdk-test-harness/releases/tag/0.10.0

posthog/request.py
```diff
@@ -321,6 +322,7 @@ def flags(
         gzip,
         timeout,
         session=_get_flags_session(),
+        api_key_field="token",
```
marandaneto: isnt token and api_key aliases? so we dont need that

marandaneto (approved):

posthog/request.py
eli-r-ph: ah good call, updated 👍

PR #707, opened 2026-06-28
fix: Retry flags only on transport errors

marandaneto:

Align Python `/flags/?v=2` retry behavior with the other SDK fixes: retry only bounded network/transport/timeout failures, and do not retry HTTP/API status responses like 408, 429, or 5xx.

This keeps transient network handling while avoiding repeated requests when PostHog has already returned an API response.

- `uv run --extra test --extra dev ruff check posthog/request.py posthog/client.py posthog/__init__.py posthog/test/test_request.py posthog/test/test_client.py`
- `uv run --extra test pytest posthog/test/test_request.py posthog/test/test_client.py`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with pi. The retry policy was changed from urllib3 status-code retries to explicit `/flags` transport retries, with default max retries set to 1, `0` disabling retries, and exponential backoff starting at 300ms.

### DIFF
--- a/posthog/__init__.py
+++ b/posthog/__init__.py
@@ -67,6 +67,7 @@ from posthog.flag_definition_cache import (
     FlagDefinitionCacheProvider as FlagDefinitionCacheProvider,
 )
 from posthog.request import (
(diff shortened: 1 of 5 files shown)


marandaneto:
Closing in favor of a fork-backed branch because repository rules require signed commits on origin feature branches and the follow-up public API snapshot update could not be pushed directly. Replacement PR coming next.

PR #708, opened 2026-06-28
fix: Retry flags only on transport errors

marandaneto:

Align Python `/flags/?v=2` retry behavior with the other SDK fixes: retry only bounded network/transport/timeout failures, and do not retry HTTP/API status responses like 408, 429, or 5xx.

This keeps transient network handling while avoiding repeated requests when PostHog has already returned an API response.

- `uv run --extra test --extra dev ruff check posthog/request.py posthog/client.py posthog/__init__.py posthog/test/test_request.py posthog/test/test_client.py`
- `uv run --extra test --extra dev python .github/scripts/check_public_api.py`
- `uv run --extra test pytest posthog/test/test_request.py posthog/test/test_client.py`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with pi. The retry policy was changed from urllib3 status-code retries to explicit `/flags` transport retries, with default max retries set to 1, `0` disabling retries, and exponential backoff starting at 300ms.


posthog/client.py
```diff
@@ -895,10 +902,15 @@ def _get_flags_decision(
...
+        if self.feature_flags_request_max_retries != 1:
+            flag_request_options["max_retries"] = self.feature_flags_request_max_retries
+
         resp_data = flags(
             self.api_key,
             self.host,
             timeout=self.feature_flags_request_timeout_seconds,
+            **flag_request_options,
             **request_data,
         )
```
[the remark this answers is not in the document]
marandaneto: Good catch — fixed in fae8487 by passing `max_retries` directly to `flags()` on every feature flag request.

posthog/test/test_client.py
```diff
@@ -984,6 +984,21 @@ def test_basic_capture_with_feature_flags_returns_active_only(self, patch_flags)
...
+        patch_flags.return_value = {"featureFlags": {}, "featureFlagPayloads": {}}
+        client = Client(
+            FAKE_TEST_API_KEY,
+            feature_flags_request_max_retries=0,
+            personal_api_key=FAKE_TEST_API_KEY,
+        )
+
+        client.get_all_flags("distinct_id")
+
+        self.assertEqual(patch_flags.call_args.kwargs["max_retries"], 0)
```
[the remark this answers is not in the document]
marandaneto: Fixed in fae8487 — the forwarding coverage is now parameterized for the default value, disabled retries, and a non-default retry count.

posthog/request.py
```diff
@@ -41,8 +42,7 @@
     if hasattr(socket, attr):
         KEEP_ALIVE_SOCKET_OPTIONS.append((socket.SOL_TCP, getattr(socket, attr), value))
 
-# Status codes that indicate transient server errors worth retrying
-RETRY_STATUS_FORCELIST = [408, 500, 502, 503, 504]
```
marandaneto, line 46: idk if that was intentional and why not all the other sdks do that, so i am assuming we should remove it and just retry for transient errors which is what all sdks should do
if all sdks should retry for those http errors as well, then i'd port this to all the other sdks

marandaneto: @PostHog/team-feature-flags ?

marandaneto: same here https://github.com/PostHog/posthog-elixir/pull/155

marandaneto:
@PostHog/team-feature-flags ping

posthog/request.py
gustavohstrassburger: I think there is a difference in what we're considering a transient network error here.
This was originally introduced by @haacked in #392.
It seems intentional and makes sense from the PR perspective, so I'll let him reply to this.

marandaneto: not based on this recent change https://github.com/PostHog/posthog-ruby/pull/198/changes that should also retry on those http errors, if that was intentional
either python or ruby are the best approach, we can decide and port the best approach across sdks

marandaneto: Updated this PR to match the new SDK test harness contract from PostHog/posthog-sdk-test-harness#30: `/flags` now retries the contract-defined transient HTTP statuses 502 and 504, while continuing not to retry other HTTP/API statuses such as 408, 429, 500, or 503.

marandaneto:
this will require https://github.com/PostHog/posthog-sdk-test-harness/pull/30 to be merged/released

ioannisj (approved):
LG

marandaneto:
Closing this fork PR in favor of the same-repository PR: https://github.com/PostHog/posthog-python/pull/724

PR #717, opened 2026-07-01
ci: Allow flags board workflow to inherit secrets

marandaneto:
The feature flags project-board caller sometimes failed CI around Semgrep's broad `secrets: inherit` finding. Replacing it with explicit secret mapping caused GitHub to reject the reusable workflow at startup because the central `.github` workflow does not declare `workflow_call.secrets`.

This keeps the required inheritance for that reusable workflow and documents the Semgrep exception inline.

- Parsed `.github/workflows/call-flags-project-board.yml` with PyYAML and verified `secrets` resolves to `inherit`.
- Ran `git diff --check`.
- Ran Semgrep locally with a rule matching `secrets: inherit` and verified `# nosemgrep` suppresses the intentional exception.
- Ran the PostHog `.github` Semgrep rules locally against the workflow file.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Pi investigated the failing workflow behavior, confirmed explicit secret mapping is incompatible with the pinned central reusable workflow, and kept `secrets: inherit` as the required GitHub Actions configuration while adding a scoped Semgrep suppression with context for reviewers.

### DIFF
--- a/.github/workflows/call-flags-project-board.yml
+++ b/.github/workflows/call-flags-project-board.yml
@@ -14,6 +14,8 @@ jobs:
             pr_number: ${{ github.event.pull_request.number }}
             pr_node_id: ${{ github.event.pull_request.node_id }}
             is_draft: ${{ github.event.pull_request.draft }}
-        secrets:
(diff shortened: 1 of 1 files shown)


ioannisj (approved):
Trust that this is safe to do

PR #727, opened 2026-07-02
feat: add secret_key config, deprecate personal_api_key

turnipdabeets:
## Problem

The credential used for local feature flag evaluation and remote config accepts **either** a Personal API Key (`phx_...`) **or** a Project Secret API Key (`phs_...`), but the config field is named `personal_api_key`. Passing a project secret key to a field called "personal api key" is confusing now that project secret keys exist.

## Change

- Add **`secret_key`** as the canonical config field — accepts a Personal API Key or a Project Secret API Key.
- Keep **`personal_api_key`** as a **deprecated alias** that resolves into `secret_key`. Non-breaking: existing code keeps working and gets a `DeprecationWarning`.
- Prefer `secret_key` when both are provided.
- Update docstrings and user-facing error messages to reference `secret_key`.

## Context

From the Slack discussion + PostHog/posthog-js#4046. This covers the **rename** only. The other two items from that thread — making flag refresh opt-in, and a config to enable/disable local eval independent of the env var — are separate.

Part of a coordinated rename across the backend SDKs (python, node, php, ruby, go, dotnet, rs).


turnipdabeets:
**Marked ready for review.**

Naming is settled — this is a generic **`secret_key`** config that accepts **either** a Personal API Key or a Project Secret API Key (both coexist; same `Authorization: Bearer` header). `personal_api_key` stays as a deprecated **config-name** alias that forwards into it — *not* a deprecation of personal keys as a credential. It's a name-level change and is crash-safe with respect to using the key (adds no new error path).

Merge timing still coordinates with the project-secret-key rollout (feature-flagged today) — ready-for-review ≠ merge-now.

posthog/client.py
```diff
@@ -234,6 +234,7 @@ def __init__(
         timeout=15,
         thread=1,
         poll_interval=30,
+        secret_key=None,
         personal_api_key=None,
```
dustinbyrne: this is a breaking change if anyone is using positional arguments in the client constructor

given how many we have, it seems unlikely, but there's no downside to moving this addition to the end

dustinbyrne (commented):
some lingering references
```sh
$  git -P grep -l 'personal_api_key' -- . ':(exclude)references/**'
.env.example
example.py
examples/async_redis_flag_cache.py
examples/celery_integration.py
examples/redis_flag_cache.py
examples/remote_config.py
openfeature-provider/openfeature/contrib/provider/posthog/provider.py
openfeature-provider/tests/conftest.py
openfeature-provider/tests/test_provider_unit.py
posthog/__init__.py
posthog/ai/prompts.py
posthog/client.py
posthog/flag_definition_cache.py
posthog/request.py
posthog/test/ai/test_prompts.py
posthog/test/test_client.py
posthog/test/test_client_fork.py
posthog/test/test_feature_flags.py
posthog/test/test_flag_definition_cache.py
```

posthog/client.py
turnipdabeets: Done — moved `secret_key` to the end of the constructor params (2b87f8a) so no existing positional argument shifts. Thanks!

dustinbyrne (approved):

PR #729, opened 2026-07-03
fix: mirror django request user agent into $raw_user_agent

lricoy:

The Django middleware tags all events in a request with `$user_agent`, but PostHog's server-side classification — web analytics bot detection (`$virt_is_bot` and friends) and the materialized-column fast path — keys on the standardized `$raw_user_agent` property (the one posthog-js sends). Server-side-tracked pageviews forwarding a real client UA are exactly the traffic where bot classification matters, and today they miss it.

PostHog/posthog#68253 removes the `$user_agent` query-time fallback for performance (it forced a full properties-blob read on every classification query — see the PR for production measurements), so this change mirrors the header into `$raw_user_agent`, aligning the middleware with the standardized property. Related: PostHog/posthog#68232.

`$user_agent` is still sent for backwards compatibility.

Agent-authored. Extended `test_extract_tags_basic` to assert both properties carry the request's User-Agent header; ran the middleware extract_tags tests locally (11 passed).

## :pencil: Checklist

- [x] Changeset added in `.sampo/changesets/`.

## 🤖 Agent context

Built with Claude Code as part of a cross-repo effort to standardize on `$raw_user_agent` (PostHog/posthog#68232, PostHog/posthog#68253, with sibling PRs to posthog-android and posthog-ruby).

### DIFF
--- a/.sampo/changesets/mirror-raw-user-agent.md
+++ b/.sampo/changesets/mirror-raw-user-agent.md
@@ -1,5 +1,5 @@
 ---
-"posthog": patch
+"pypi/posthog": patch
 ---
 
 Django middleware also sends the request user agent as `$raw_user_agent`, the standardized property PostHog's server-side classification (e.g. bot detection) reads
--- a/.sampo/changesets/mirror-raw-user-agent.md
+++ b/.sampo/changesets/mirror-raw-user-agent.md
@@ -1,0 +1,5 @@
+---
+"posthog": patch
+---
+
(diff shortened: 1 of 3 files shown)


.sampo/changesets/mirror-raw-user-agent.md
```diff
@@ -0,0 +1,5 @@
+---
+"posthog": patch
```
marandaneto:
```suggestion
"pypi/posthog": patch
```

dustinbyrne (approved):

Issue #146, opened 2024-11-26
Backpressure when queue is full


mariusandra:
We [uncovered a case](https://posthog.slack.com/archives/C03K4JWRPPA/p1732554829330079?thread_ts=1732201357.190059&cid=C03K4JWRPPA) where a key customer was losing events due to the way this library is configured/built.

With the `flush_interval=2` (default is `0.5`), `flush_at=100` and `max_queue_size=1000` (default 10k), their queue got full during peak hours and they started dropping events.

Their recommended solution is very reasonable:

> Ideally the library would provide back pressure by blocking/waiting for acknowledgement when flushing a batch of events vs having to configure a max_queue_size.

We should implement some way to block when the queue is full, or some way to let the user know the queue size.

emmayusufu:
Looking to take this. Played with the customer's config (`flush_interval=2, flush_at=100, max_queue_size=1000`) against `_enqueue` around line 1330 — `self.queue.put(msg, block=False)` always drops when full.

What I'd add: an `on_queue_full` param with three values.

- `"drop"` stays default, no break.
- `"block"` switches to `put(block=True, timeout=...)`, with `queue_full_block_timeout` (default `None`) for callers who want a bounded wait. If the timeout expires, falls back through the existing warn-and-drop branch so it still terminates.
- `"raise"` lets `queue.Full` propagate.

One snag worth raising before opening the PR — the `capture` wrapper at line 142 swallows exceptions unless `debug=True`. So `"raise"` only raises when debug is on. Two ways to handle: document it on the param, or carve `queue.Full` out of the wrapper as a special case. I'd document. The wrapper carve-out changes contract for everyone using the lib, just to support one strategy.

Tests are written locally, six of them, all green.

PR #731, opened 2026-07-06
fix: say the event is dropped when the queue is full

emmayusufu:
The queue-full warning now says the event is being dropped (with the queue maxsize and event name) instead of only reporting that the queue is full, and drops the stale analytics-python name.

Follows the discussion below: the original backpressure modes and the on_drop callback are gone, the public API is unchanged.

Context: #146

### DIFF
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -1712,7 +1712,11 @@ class Client(object):
             self.log.debug("enqueued %s.", msg["event"])
             return sent_uuid
         except Full:
-            self.log.warning("analytics-python queue is full")
+            self.log.warning(
+                "event queue is full (maxsize %d), dropping event %s",
+                self.queue.maxsize,
+                msg["event"],
+            )
             return None
 
     def flush(self, timeout_seconds: Optional[float] = 10) -> None:
--- a/.sampo/changesets/queue-full-log-message.md
+++ b/.sampo/changesets/queue-full-log-message.md
@@ -1,0 +1,5 @@
+---
+pypi/posthog: patch
+---
+
+Clarify the queue-full warning to say the event is being dropped, instead of only reporting that the queue is full.
--- a/posthog/test/test_client.py
+++ b/posthog/test/test_client.py
@@ -2114,9 +2114,11 @@ class TestClient(unittest.TestCase):
         for i in range(10):
             client.capture("test event", distinct_id="distinct_id")
 
-        msg_uuid = client.capture("test event", distinct_id="distinct_id")
+        with self.assertLogs("posthog", level="WARNING") as logs:
+            msg_uuid = client.capture("test event", distinct_id="distinct_id")
         # Make sure we are informed that the queue is at capacity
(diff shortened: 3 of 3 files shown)


emmayusufu:
Addressed all three: negative `queue_full_block_timeout` now raises `ValueError` at init instead of surfacing from `Queue.put` at capture time, the queue-full outcome tests are folded into one `@parameterized.expand` table (drop, block with timeout, raise with and without debug) plus a second parameterised table for the invalid configs, and the standalone default-drop test is gone since `test_overflow` already covers the implicit default.

dustinbyrne (commented):
Hey @emmayusufu, thanks for putting this together. I've been thinking about this quite a bit. Most critically, I think data loss is our only real option here. Consider either of the following scenarios in which the buffer cannot be drained quickly enough:

1. The client is producing events faster than they can be processed and flushed out of the buffer.
2. An intermittent network issue or outage prevents captured events from reaching PostHog due to timeouts, server errors, etc.

Either scenario would catastrophic outcomes if the buffer becomes full and every subsequent capture becomes blocking. Because a full buffer already indicates that the consumer cannot keep up, blocking can amplify the problem by tying up application threads.

There is no lossless outcome once production remains above the drain rate, whatever the cause. We either drop events, allow unbounded memory growth, persist them elsewhere, or transfer back pressure into the application's worker pool. For our SDK, I don't believe the latter option is safe.

Again, I appreciate you taking the time to open this PR. I’d be interested to hear your thoughts, particularly if you had a workload in mind where blocking would be preferable to dropping.

emmayusufu:
Hi @dustinbyrne,

I do think you are right, for an analytics SDK, blocking the caller to save event data might not be best approach because If the app is already outrunning the flush or the network is down, tying up its threads makes it worse, so dropping events is defintely the safe approach
So I removed the block and raise modes, and added an on_drop callback that receives the event when the queue is full. I think that should be able to give people visibility to count or log drops without the SDK ever blocking.

posthog/client.py
```diff
@@ -1647,6 +1656,11 @@ def _enqueue(self, msg, disable_geoip):
             return sent_uuid
         except Full:
             self.log.warning("analytics-python queue is full")
+            if self.on_drop is not None:
+                try:
+                    self.on_drop(msg)
+                except Exception:
+                    self.log.exception("on_drop callback raised")
```
dustinbyrne: I'd prefer if we could keep the public API unchanged - otherwise this becomes a non-standardization compared to other SDKs (unless we port it; though we'd need more evidence that this is useful to end users, and not a foot-gun)

Perhaps what we really need is just to clean up the log message:
> analytics-python queue is full

I believe we append something like `[PostHog] ` to these log lines, so that's decent, but we should at least indicate that events are being dropped - not just that the queue is full.

emmayusufu:
Hey @dustinbyrne I have reworked  it down to just the log cleanup

dustinbyrne (approved):
Thank you @emmayusufu!

PR #734, opened 2026-07-10
fix(flags): local eval dependency fail to match

matheus-vb:

Under local evaluation a release condition of the form "flag X evaluates to `false`" never matched, so the dependent flag returned `false` for every locally evaluated user. The chain walk in `evaluate_flag_dependency` rejected any falsy flag in the chain before the value comparison ran, and since the referenced flag is always last in the chain the users whose flag X was `false` (exactly the ones the condition should match) were dropped. 

This reworks the chain walk to only evaluate and cache each member and to decide the outcome solely from the referenced flag's cached value, matching the Go and C# SDKs and the Rust `/flags` service. A malformed `value: null` condition now falls back to the server instead of silently returning `true`.

Added end to end tests in `posthog/test/test_feature_flags.py` covering `value: false` with variant and `true` controls, the multi level A/B/D chain, an inactive base flag, an inconclusive base that falls back to `/flags`, and `value: null`. Reran the bug report reproduction and the full suite passes.

## :pencil: Checklist

### If releasing new changes

### DIFF
--- a/.sampo/changesets/stalwart-iceseeker-sampsa.md
+++ b/.sampo/changesets/stalwart-iceseeker-sampsa.md
@@ -1,0 +1,5 @@
+---
+pypi/posthog: patch
+---
+
+Fix local evaluation of flag dependencies with a `flag_evaluates_to: false` condition: such conditions never matched, forcing the dependent flag to `false` for every locally-evaluated user.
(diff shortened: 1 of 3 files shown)


ioannisj (approved):
LG

PR #739, opened 2026-07-14
feat(metrics): add posthog.metrics count/gauge/histogram API (alpha)

DanielVisca:
## What this is

The Python leg of the `posthog.metrics` SDK surface ([Metrics docs](https://posthog.com/docs/metrics)) — a faithful port of the posthog-js core client ([posthog-js #4115](https://github.com/PostHog/posthog-js/pull/4115), node wiring in [#4117](https://github.com/PostHog/posthog-js/pull/4117)), so backend Python services can record metrics with zero OpenTelemetry setup:

```python
client = Posthog("<ph_project_api_key>", metrics={"service_name": "billing-worker"})
client.metrics.count("invoices.processed", 1, attributes={"plan": "pro"})
client.metrics.gauge("queue.depth", 42)
client.metrics.histogram("job.duration", 187, unit="ms")
```

## Design (mirrors posthog-js exactly, adapted to Python)

- **Statsd-style pre-aggregation**: samples fold into per-series aggregates in memory; a burst of 10k `count()` calls is one data point on the wire. Delta temporality throughout, so restarts need no cross-window state.
- **Wire shape pinned to the ingest**: OTLP/JSON `ExportMetricsServiceRequest` to `{host}/i/v1/metrics?token=...` — nano timestamps as decimal strings but histogram `count`/`bucketCounts` as plain JSON numbers (string-encoded u64s are silently dropped upstream, opentelemetry-rust#3328). Same `DEFAULT_HISTOGRAM_BOUNDS` as JS.
- **Thread-safe** (unlike the JS single-threaded original): a lock around the window, a separate flush serializer, and a daemon `threading.Timer` armed on first capture (10s default). `shutdown()` flushes and resets.
- **Guardrails**: `max_series_per_flush` cardinality cap (default 1000, warn once per window, existing series keep folding), type-collision warning, monotonic-counter validation, finite-value validation, `before_send` hook.
- **Failure handling**: 5xx/429/network → merge the unsent window back and re-arm (no data loss on transient failures); 413 → drop with a warning; other 4xx → drop with an error log.
- New file is `posthog/metrics_capture.py` (not `metrics.py` — that name would shadow a future module-level `posthog.metrics` accessor and matches the `exception_capture.py` house naming).

TDD — the test file was written first and observed failing at import, then all green:

- **New:** `posthog/test/test_metrics.py` — 14 tests, each pinning a distinct contract: count burst → one delta+monotonic data point; gauge keeps last (and has no `startTimeUnixNano`, matching JS); the full histogram wire shape (the opentelemetry-rust#3328 encoding pin); series identity split by attributes but not by attribute order; OTLP value encoding (including the Python-specific bool-before-int trap — `bool` is an `int` subclass and would silently encode `True` as `intValue: 1`); cardinality cap drops new series but keeps folding existing ones; parameterized invalid-sample drops; disabled client no-ops; resource-attribute layering (user keys can't spoof `telemetry.sdk.*`); transient-failure merge-back (a 503'd window's counts sum into the next flush); shutdown flush.
- **Regression:** `posthog/test/test_client.py` — 145 passed (I touched `Client.__init__` and `shutdown`).
- `ruff check` / `ruff format` clean.
- Not done: `check_public_api.py --write` (griffe isn't installable in my env — PEP 668); CI should regenerate/check the snapshot, and the new `Client.metrics` property will need it. Also no live end-to-end send from this environment — the wire shape is byte-identical to posthog-js's, which was validated end-to-end against ingestion during the metrics alpha.

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) — Daniel directed this; assigned as DRI.

I (Claude) wrote this by porting `packages/core/src/metrics/` from posthog-js (read in full: index.ts, metrics-utils.ts, config.ts, the value encoder in logs-utils.ts, and the `_sendMetricsBatch` transport for the URL/auth/outcome contract). Deliberate deviations from the JS original: `flush_interval` in seconds (Python client idiom, vs `flushIntervalMs`), thread-safety via locks (JS is single-threaded), no `drainWindow()` (browser-unload concern), and plain-`requests` transport with the same ok/too-large/retry-later/fatal outcome classification instead of the fetch retry stack. Part of the metrics-SDK track: js ✅ merged, node in review, python (this PR).


posthog/metrics_capture.py
```diff
@@ -0,0 +1,639 @@
...
+        return {"arrayValue": {"values": [_to_otlp_any_value(v) for v in value]}}
+    try:
+        return {"stringValue": json.dumps(value)}
+    except (TypeError, ValueError):
+        return {"stringValue": str(value)}
+
+
+def _to_otlp_key_value_list(attributes: dict) -> list:
+    return [
+        {"key": key, "value": _to_otlp_any_value(value)}
```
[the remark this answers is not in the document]
DanielVisca: Fixed in be5e767 — keys are now `str()`d at the wire boundary too, matching the series identity. Regression test added (`test_non_string_attribute_keys_stringify_on_the_wire`) asserting `{"a": 1, 2: "x"}` emits keys `{"a", "2"}` as strings.

posthog/metrics_capture.py
```diff
@@ -0,0 +1,643 @@
...
+
+    def _capture(
+        self,
+        metric_type: str,
+        name: str,
+        value: float,
+        unit: Optional[str],
+        attributes: Optional[dict],
+    ) -> None:
+        if getattr(self._client, "disabled", False):
```
dustinbyrne, line 284: [P1] `send=False` still transmits metrics

`Client.__init__` documents `send=False` as queueing captures without sending them (`posthog/client.py:287`), and the event path returns before transport when `self.send` is false. The metrics gate here checks only `disabled`, so the new capture surface bypasses that established transport switch.

For example:

```python
client = Client("phc_test", send=False, thread=0)
client.metrics.count("jobs.processed")
client.metrics.flush()
```

The current code produces:

```text
HTTP POST calls: 1
```

Expected:

```text
HTTP POST calls: 0
```

This causes unexpected telemetry egress in environments and tests that deliberately construct a no-send client.

Please make metrics capture/flush honor `client.send`, or explicitly document why metrics intentionally differ from event capture.

A regression test should cover both manual flush and timer flush with `send=False` and assert that `_get_session().post` is never called.

posthog/client.py
```diff
@@ -423,6 +425,9 @@ def __init__(
         self._flag_definition_cache_provider_async_runner_lock = threading.Lock()
         self.disabled = disabled or not self.api_key
         self.disable_geoip = disable_geoip
+        self._metrics_config = metrics
+        self._metrics: Optional[PostHogMetrics] = None
+        self._metrics_lock = threading.Lock()
```
dustinbyrne, line 430: [P1] Reinitialize the new metrics locks after `fork()` to avoid a child deadlock

`Client._reinit_after_fork()` already replaces thread-owned queue/session/cache state because locks held by vanished parent threads remain locked in the child. The new `_metrics_lock` here, plus `PostHogMetrics._lock` and `_flush_lock`, are not replaced. The metrics PID guard cannot recover because `_capture()` acquires `_lock` before calling `_reset_after_fork_locked()`.

A concrete timeline is:

1. A parent thread holds `client.metrics._lock`.
2. The process forks.
3. The child calls `client.metrics.count("after.fork")`.
4. The child blocks forever on the inherited lock.

The current reproduction exited via a two-second alarm:

```text
fork_child_exit=42
```

Expected: the child records the metric without blocking and uses fresh metrics state.

Please replace `_metrics_lock` and any instantiated metrics locks/timer/window in the existing at-fork callback, without acquiring the inherited locks.

A regression test should hold each metrics lock from another thread, call `os.fork()`, invoke the public `metrics.count()` path in the child, and assert that the child exits normally within a short deadline.

posthog/client.py
```diff
@@ -1789,6 +1818,12 @@ def shutdown(self) -> None:
             ```
         """
         self.flush(timeout_seconds=None)
+        if self._metrics is not None:
+            try:
+                self._metrics.flush()
```
dustinbyrne, line 1829: [P1] The public module-level `posthog.shutdown()` leaves metrics buffered

This adds metrics flushing to direct `Client.shutdown()`, but the documented module helper in `posthog/__init__.py:1122` still calls only `_proxy("flush")` and `_proxy("join")`. A default client obtained through `posthog.setup()` therefore bypasses this new lifecycle path.

For example:

```python
client = posthog.setup()
client.metrics.count("lost.on.global.shutdown")
posthog.shutdown()
```

The current code produces:

```text
metric HTTP calls: 0
buffered series after shutdown: 1
```

Expected:

```text
metric HTTP calls: 1
buffered series after shutdown: 0
```

This loses the final metrics window for applications using the global SDK lifecycle.

Please make module-level `shutdown()` delegate to the client's `shutdown()` implementation rather than separately calling `flush()` and `join()`.

A regression test should record through `posthog.setup().metrics`, call the public `posthog.shutdown()`, and assert that the metrics transport ran and the window was cleared.

posthog/metrics_capture.py
DanielVisca: changed flush to drop the window when send is false, added tests for manual and timer flush, thanks

posthog/client.py
DanielVisca: added the metrics locks to the fork reinit, fresh locks without acquiring the old ones, plus a fork test under an alarm, thanks

posthog/client.py
DanielVisca: changed module shutdown to proxy to client shutdown so the last window flushes, thanks

posthog/metrics_capture.py
```diff
@@ -0,0 +1,661 @@
...
+        metric_type: str,
+        unit: Optional[str],
+        attributes: Optional[dict],
+    ):
+        self.name = name
+        self.type = metric_type
+        self.unit = unit
+        # Snapshot: the series key was computed from these values, so a caller
+        # mutating the dict after capture must not change the stored series.
+        self.attributes = dict(attributes) if attributes else None
```
dustinbyrne, line 162: [P1] Snapshot nested attribute values or enforce the scalar contract

`_SeriesState` only shallow-copies the attributes dictionary. Python currently treats nested values as supported: `_series_key()` explicitly handles lists/dicts, `_to_otlp_any_value()` encodes arrays, and the tests cover list attributes.

This lets the stored attributes diverge from the series key. For example:

1. Record with `attributes={"tags": ["a"]}`.
2. Mutate the list to `["a", "b"]`.
3. Record again and flush.

The current payload has two points, both with `tags=["a", "b"]` and value `1`, although the first series key was created for `["a"]`.

Browser and Node restrict metric attributes to scalar values, making their shallow copy sufficient. Please either deeply snapshot/normalize Python's supported nested values, or align Python with the other two SDKs by validating attributes as scalars and dropping nested values. Add a regression that mutates a nested value after the first capture and verifies stable wire attributes.

posthog/metrics_capture.py
```diff
@@ -0,0 +1,661 @@
...
+        # nothing is transmitted — the flushed window is discarded.
+        if not getattr(self._client, "send", True):
+            return
+
+        payload = self._build_payload(window)
+        outcome = self._send(payload)
+        if outcome == "retry-later":
+            with self._lock:
+                self._consecutive_send_failures += 1
+                if self._consecutive_send_failures > _MAX_CONSECUTIVE_SEND_FAILURES:
```
dustinbyrne, line 473: [P2 parity] Define a consistent bounded retry policy across metrics SDKs

Python retries at the fixed flush interval and drops the window on the fourth failure. Browser and Node also use a fixed metrics flush interval, but retain and retry the window forever; neither applies exponential backoff across failed metrics flushes. Node only adds fixed-delay transport retries within each attempt.

Retrying forever at a fixed cadence does not seem like the right behavior to copy. Can we choose an explicit cross-SDK policy—ideally capped exponential backoff plus a documented attempt or age limit—and pin it with tests? The shared JS logs implementation already uses exponential backoff capped at 64× the base interval and may be a useful precedent.

At minimum, Python's release note and implementation need reconciling: it says the window is dropped after three failed flushes, while this `> 3` condition drops on the fourth.

posthog/client.py
```diff
@@ -1715,6 +1726,30 @@ def _enqueue(self, msg, disable_geoip):
...
+            client = Posthog("<ph_project_api_key>", metrics={"service_name": "billing-worker"})
+            client.metrics.count("invoices.processed", 1, attributes={"plan": "pro"})
+            client.metrics.gauge("queue.depth", 42)
+            client.metrics.histogram("job.duration", 187, unit="ms")
+            ```
+        """
+        if self._metrics is None:
+            with self._metrics_lock:
+                if self._metrics is None:
+                    self._metrics = PostHogMetrics(self, self._metrics_config)
```
dustinbyrne, line 1750: [P1] Keep lazy metrics initialization inside the client's no-throw contract

Most public `Client` methods use `@no_throw`, but `client.metrics` constructs `PostHogMetrics` before the guarded recording path is reached. Because the nested config is typed only as `dict`, a type checker accepts:

```python
Client("phc_...", metrics={"resource_attributes": "bad"})
```

and `client.metrics.count(...)` then raises `AttributeError` into application code.

Please give this path the same no-throw semantics as the rest of the client: validate nested configuration and fall back to defaults or a valid no-op metrics instance when initialization fails. Decorating the property with `@no_throw()` while returning `None` would not be sufficient, because the chained `.count(...)` would still throw. Add a regression asserting the complete public call does not raise.

dustinbyrne (commented):
Human-driven, agent-assisted follow-up review.

dustinbyrne (approved):
approving to unblock but there's some useful considerations in here

PR #740, opened 2026-07-14
fix(flags): malformed flag-dependency conditions no-match locally

patricio-posthog:
## Problem

PR #734 (released as 7.22.1) added upfront validation to `evaluate_flag_dependency` that raises `InconclusiveMatchError` when a flag-dependency condition is malformed (wrong operator, missing `key`, or `value: null`).

`InconclusiveMatchError` is the signal that local evaluation cannot decide, so the client falls back to a network call to the billable `/flags` endpoint — on **every evaluation** of the affected flag. Customers whose flag definitions contain a single malformed dependency condition (typically legacy `value: null` data) went from near-zero `/flags` traffic to millions of requests per day.

## Fix

Mirror the server-side Rust evaluator (`match_flag_value_to_flag_filter`), which never errors on malformed conditions — it treats them as **no match**:

- `operator` is not `flag_evaluates_to` → `return False`
- `key` is missing/empty → `return False`
- `value` is `None`/missing → `return False`

These are now **local, definitive results**: no `InconclusiveMatchError`, no `/flags` fallback. A non-matching condition just means the flag falls through to its other conditions (or evaluates to false if none match) — exactly what the server computes.

Genuinely inconclusive cases still raise and fall back to the server: referenced flag not present locally, an inconclusive dependency in the chain, circular dependencies, and missing evaluation context.

## Tests

- Updated the PR #734 null-value test: it now asserts the flag resolves locally (no-match → `False`) with **zero** `/flags` calls.
- Added coverage for `value: None`, empty/missing `key`, and wrong operator, each asserting zero `/flags` calls.
- Added a control test: a flag whose only condition is the malformed dependency resolves to `False` locally, while a flag with a second matching condition group still resolves via that group.
- Preserved the PR #734 `value: false` tests.

Full feature-flags suite passes (158 passed).

## Changeset

Added `.sampo/changesets/malformed-flag-dependency-no-match.md` (`pypi/posthog: patch`).

### DIFF
--- a/posthog/feature_flags.py
+++ b/posthog/feature_flags.py
@@ -15,7 +15,32 @@ __LONG_SCALE__ = float(0xFFFFFFFFFFFFFFF)
 
 log = logging.getLogger("posthog")
 
+# Tracks (flag_key, reason) pairs already warned about for malformed flag
+# dependency conditions, so the warning fires at most once per process instead
+# of on every get_feature_flag call. A plain set is safe under the GIL; a rare
+# duplicate warning from a race is acceptable.
+_warned_malformed_flag_dependencies: set = set()
+
 NONE_VALUES_ALLOWED_OPERATORS = ["is_not"]
+
+
+def _warn_malformed_flag_dependency_once(flag_key, reason):
+    """Emit a throttled warning for a malformed flag dependency condition.
+
+    Mirrors the server-side Rust evaluator, which logs before treating the
+    condition as not matching, so customers can discover broken flag
+    definitions locally. Deduplicated on (flag_key, reason) to avoid spamming
+    hot evaluation paths.
+    """
+    key = (flag_key, reason)
+    if key in _warned_malformed_flag_dependencies:
+        return
+    _warned_malformed_flag_dependencies.add(key)
+    log.warning(
+        f"Flag dependency condition on '{flag_key or 'unknown'}' is malformed "
+        f"({reason}); treating as not matching during local evaluation. "
+        f"Fix this condition in the PostHog UI."
+    )
 
 
 class ConditionMatch(Enum):
@@ -160,8 +185,12 @@ def evaluate_flag_dependency(
     operator = property.get("operator", "exact")
 
     if operator != "flag_evaluates_to":
+        _warn_malformed_flag_dependency_once(
+            flag_key, f"invalid operator '{operator}'"
+        )
         return False
     if not flag_key or expected_value is None:
+        _warn_malformed_flag_dependency_once(flag_key, "missing key or value")
         return False
 
     # Check if dependency_chain is present - it should always be provided for flag dependencies
--- a/posthog/feature_flags.py
+++ b/posthog/feature_flags.py
@@ -134,17 +134,35 @@ def evaluate_flag_dependency(
 
     Returns:
         bool: Whether the referenced flag's evaluated value matches the
-            condition's expected value.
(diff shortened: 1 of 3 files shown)


matheus-vb (approved):

turnipdabeets:
LGMT, but two questions before merging:

1. Should we log when we skip a malformed condition? The Rust side doesn't return false silently — it does tracing::error!("Flag filter operator for property type Flag must be flag_evaluates_to, skipping...") before returning. With the raise removed and no log added, a customer with a malformed condition now gets False forever with zero breadcrumbs. A log.warning (maybe once per flag key to avoid hot-path spam) would let them discover their broken flag definitions locally.

2. Why does missing dependency_chain stay inconclusive? It's equally "malformed server data", and it would cause the exact same billable-fallback-on-every-evaluation storm this PR fixes — just triggered by a different shape. Is there a case where the server genuinely resolves it better, or should it also be a definitive no-match? Fine as a follow-up, just want the reasoning recorded.

posthog/feature_flags.py
ioannisj, line 815: if the server's cohort evaluator treats a malformed flag-dep condition as "never contributes to a match, negated or not," then the SDK now diverges?

ioannisj (approved):
LG left a minor comment spotted by my agent

patricio-posthog:
> LGMT, but two questions before merging:
> 
> 1. Should we log when we skip a malformed condition? The Rust side doesn't return false silently — it does tracing::error!("Flag filter operator for property type Flag must be flag_evaluates_to, skipping...") before returning. With the raise removed and no log added, a customer with a malformed condition now gets False forever with zero breadcrumbs. A log.warning (maybe once per flag key to avoid hot-path spam) would let them discover their broken flag definitions locally.
> 2. Why does missing dependency_chain stay inconclusive? It's equally "malformed server data", and it would cause the exact same billable-fallback-on-every-evaluation storm this PR fixes — just triggered by a different shape. Is there a case where the server genuinely resolves it better, or should it also be a definitive no-match? Fine as a follow-up, just want the reasoning recorded.

Good questions @turnipdabeets 

1. About logging: I agree, I pushed a commit adding a `log.warning` at the early return, mirroring the Rust side's `tracing::error!`. 

2. About the missing `dependency_chain` and why it stays inconclusive: the distinction is server parity. The three malformed shapes return `False` because the server's evaluator (`match_flag_value_to_flag_filter` in `rust/feature-flags/src/flags/flag_matching_utils.rs`) computes the same `false` for them, and answering locally can't diverge. But a missing `dependency_chain` is different, bcause the condition itself may be perfectly well-formed, and returning `False` there could silently flip a flag the server would say is `true`.

I agreed it's worth a follow-up though.

posthog/feature_flags.py
patricio-posthog: I checked the Rust side, and the server's cohort evaluator doesn't skip these, so no new
divergence. `evaluate_cohort_filter` routes any non-cohort filter (including flag-type)
through `evaluate_property_with_negation`
(`rust/feature-flags/src/cohorts/cohort_operations.rs`), where `match_property` returns an
error for `FlagEvaluatesTo` (`rust/feature-flags/src/properties/property_matching.rs:532`)
and `.unwrap_or(false)` turns that into `false`, and then negation is applied. So the
server's behavior for a malformed flag-dep inside a cohort is exactly false-then-negate,
which is what this code now does too.

turnipdabeets (approved):

PR #743, opened 2026-07-16
feat(error-tracking): link exceptions to active spans

hpouillot:

Exceptions captured inside instrumented code currently lose their active OpenTelemetry context, which prevents correlation with the trace and span where the exception occurred.

`capture_exception` now reads a valid active OpenTelemetry span when the optional dependency is available and adds zero-padded `$trace_id` and `$span_id` properties. Explicitly supplied properties retain precedence, and exception capture remains unchanged when OpenTelemetry is absent or no valid span is active.

- `uv run --extra test pytest posthog/test/test_client.py -k 'capture_exception' --verbose --timeout=30`
- `uv run ruff format --check . && uv run ruff check .`
- `uv run --extra dev --extra test mypy --no-site-packages --config-file mypy.ini . | uv run --extra dev mypy-baseline filter`
- `uv run make public_api_check`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with the Pi coding agent. The active span is read through a lazy OpenTelemetry import so the SDK keeps its optional dependency behavior. Only valid contexts are emitted, identifiers use standard fixed-width lowercase hexadecimal formatting, and caller-supplied values take precedence. A focused branch was created from current `main` to avoid including unrelated local error-tracking work.

### DIFF
--- a/.sampo/changesets/capture-exception-otel-context.md
+++ b/.sampo/changesets/capture-exception-otel-context.md
@@ -1,0 +1,5 @@
+---
+pypi/posthog: minor
+---
+
+Add the active OpenTelemetry span's `$trace_id` and `$span_id` to events captured with `capture_exception`.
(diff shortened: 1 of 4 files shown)


cat-ph (approved):
let's go! 🚀

PR #746, opened 2026-07-16
fix(ai): preserve Anthropic cache-write TTL breakdowns

### LINKED ISSUE #745: Some Python AI integrations drop Anthropic cache-write TTL details
### Version

7.25.0

### Steps to Reproduce

PostHog/posthog#71323 adds separate token fields for Anthropic's 5-minute and 1-hour cache writes. I checked each Python integration to see whether that breakdown reaches ingestion.

Two paths currently lose it:

- The LangChain callback reads the aggregate cache-write count but ignores `ephemeral_5m_input_tokens` and `ephemeral_1h_input_tokens` from `usage_metadata.input_token_details`.
- The Claude Agent SDK processor copies aggregate usage fields but discards the nested `cache_creation` object.

For example, pass this synthetic usage payload through the LangChain callback:

```python
{
    "input_tokens": 800,
    "output_tokens": 50,
    "input_token_details": {
        "cache_read": 200,
        "cache_creation": 0,
        "ephemeral_5m_input_tokens": 300,
        "ephemeral_1h_input_tokens": 200,
    },
}
```

### Expected Result

The resulting `$ai_generation` event keeps the aggregate and the complete TTL pair:

```text
$ai_cache_creation_input_tokens = 500
$ai_cache_creation_5m_input_tokens = 300
$ai_cache_creation_1h_input_tokens = 200
```

For this LangChain payload, normalized uncached input is `100`: `800 - 200 cache reads - 500 cache writes`.

Native PostHog Anthropic clients, including the Bedrock and Vertex wrappers, already preserve the official nested breakdown in `$ai_usage`; the server PR can extract it. LangGraph uses the same PostHog LangChain callback, so it is covered by the callback fix.

### Actual Result

The LangChain path emits an aggregate value of zero, omits both TTL fields, and leaves the 500 cache-write tokens in normalized input. The Claude Agent SDK path emits only its aggregate cache-write value.

The SDK follow-up should fix those loss points, retain aggregate-only behavior, and keep malformed or unavailable TTL data on the existing fallback. Direct `boto3`/OTel traffic is not included: PostHog Python does not wrap that client, and current instrumentation does not expose the TTL split.

This issue is the Python SDK follow-up to PostHog/posthog#71306 and PostHog/posthog#71323. The companion PR should remain draft until the server-side interface is merged.

gouveags:
> [!NOTE]
> This PR depends on PostHog/posthog#71323 for server-side pricing. It can merge independently, but TTL-specific cost calculation takes effect only after the server-side change ships.

PostHog/posthog#71323 adds separate costing for Anthropic's 5-minute and 1-hour cache writes. The Python SDK already preserves that breakdown in its native Anthropic wrappers, but two integrations drop it before ingestion.

Fixes #745. Related to PostHog/posthog#71306. Depends on PostHog/posthog#71323.

This change:

- Makes the LangChain callback emit the aggregate cache-write count and a complete 5-minute/1-hour property pair. LangGraph uses the same callback path.
- Preserves raw usage in Claude Agent stream events and result-message fallback, matching the native Anthropic integration.
- Treats manually supplied TTL token properties as passthrough token data.
- Adds regression coverage for native Anthropic sync, async, streaming, and the shared Anthropic Bedrock wrapper path.

The PostHog Bedrock and Vertex clients reuse the native Anthropic message adapters. Aggregate-only events remain unchanged; invalid or empty TTL details fall back to the aggregate, and a valid breakdown replaces rather than adds to it.

Direct `boto3` calls are outside this PR because `posthog-python` has no boto3 Bedrock wrapper. The OTel integration forwards spans through a separate endpoint and is not changed.

- Full suite on the PR head: `1826 passed, 67 skipped`; one unrelated, order-dependent logging assertion failed in `posthog/test/test_consumer.py::TestConsumer::test_message_only_error_logs_include_posthog_prefix`, and it passes in isolation.
- Focused AI/token suite: `143 passed, 9 skipped`.
- Ruff format and lint.
- Mypy across 186 source files.
- Public API snapshot and import-warning checks.
- Package build and Twine validation.

## :pencil: Checklist

- [x] I updated the docs if needed. The server properties are documented in PostHog/posthog#71323.

### If releasing new changes

- [x] Added a Sampo minor changeset.

## 🤖 Agent context

I used Codex to inspect the server contract, audit the Python SDK integration paths, implement the focused changes, and run independent test, code, and writing reviews. No public session link is available.

The first pass focused on the LangChain callback. The broader audit showed that native Anthropic wrappers already preserve raw usage, while Claude Agent needed the same raw-usage path. The final diff reuses those existing mechanisms instead of adding provider pricing or a new integration layer.

### DIFF
--- a/.sampo/changesets/anthropic-cache-ttl-breakdown.md
+++ b/.sampo/changesets/anthropic-cache-ttl-breakdown.md
@@ -1,0 +1,5 @@
+---
+pypi/posthog: patch
+---
+
+Preserve Anthropic cache-write TTL breakdowns across Python SDK AI integrations.
--- a/.sampo/changesets/anthropic-cache-ttl-breakdown.md
+++ b/.sampo/changesets/anthropic-cache-ttl-breakdown.md
@@ -1,5 +1,5 @@
 ---
-pypi/posthog: patch
+pypi/posthog: minor
 ---
 
 Preserve Anthropic cache-write TTL breakdowns across Python SDK AI integrations.
--- a/posthog/ai/claude_agent_sdk/processor.py
+++ b/posthog/ai/claude_agent_sdk/processor.py
@@ -45,6 +45,7 @@ class _GenerationData:
     output_tokens: int = 0
     cache_read_input_tokens: int = 0
     cache_creation_input_tokens: int = 0
+    raw_usage: Optional[Dict[str, Any]] = None
     start_time: float = 0.0
     end_time: float = 0.0
     span_id: str = field(default_factory=lambda: str(uuid.uuid4()))
@@ -82,9 +83,11 @@ class _GenerationTracker:
             self._current.cache_creation_input_tokens = usage.get(
                 "cache_creation_input_tokens", 0
             )
+            self._current.raw_usage = dict(usage)
 
         elif event_type == "message_delta" and self._current is not None:
             usage = raw.get("usage", {})
+            self._current.raw_usage = {**(self._current.raw_usage or {}), **usage}
             # message_delta usage reports cumulative output tokens
             if usage.get("output_tokens"):
                 self._current.output_tokens = usage["output_tokens"]
@@ -453,6 +456,8 @@ class PostHogClaudeAgentProcessor:
             properties["$ai_cache_creation_input_tokens"] = (
                 gen.cache_creation_input_tokens
             )
+        if gen.raw_usage:
+            properties["$ai_usage"] = gen.raw_usage
 
         if gen.stop_reason is not None:
             properties["$ai_stop_reason"] = gen.stop_reason
@@ -520,6 +525,8 @@ class PostHogClaudeAgentProcessor:
             properties["$ai_cache_read_input_tokens"] = cache_read
         if cache_creation:
             properties["$ai_cache_creation_input_tokens"] = cache_creation
+        if usage:
+            properties["$ai_usage"] = usage
 
         if result.total_cost_usd is not None:
             properties["$ai_total_cost_usd"] = result.total_cost_usd
--- a/posthog/ai/claude_agent_sdk/processor.py
+++ b/posthog/ai/claude_agent_sdk/processor.py
@@ -526,7 +526,7 @@ class PostHogClaudeAgentProcessor:
         if cache_creation:
             properties["$ai_cache_creation_input_tokens"] = cache_creation
         if usage:
-            properties["$ai_usage"] = usage
+            properties["$ai_usage"] = dict(usage)
 
         if result.total_cost_usd is not None:
             properties["$ai_total_cost_usd"] = result.total_cost_usd
--- a/posthog/ai/utils.py
+++ b/posthog/ai/utils.py
@@ -1,6 +1,6 @@
 import time
 import uuid
-from typing import Any, Callable, Dict, List, Optional, cast
+from typing import Any, Callable, Dict, List, Optional, Tuple, cast
 
 from posthog import get_tags, identify_context, new_context, tag, contexts
 from posthog.ai.gateway import warn_if_posthog_ai_gateway
@@ -16,10 +16,34 @@ _TOKEN_PROPERTY_KEYS = frozenset(
         "$ai_output_tokens",
         "$ai_cache_read_input_tokens",
         "$ai_cache_creation_input_tokens",
+        "$ai_cache_creation_5m_input_tokens",
+        "$ai_cache_creation_1h_input_tokens",
         "$ai_total_tokens",
         "$ai_reasoning_tokens",
     }
 )
+
+
+def _extract_cache_creation_ttl_breakdown(
+    cache_creation: Any,
+) -> Optional[Tuple[int, int]]:
+    """Return a complete 5-minute/1-hour cache-write pair when usable."""
+    if not isinstance(cache_creation, dict):
+        return None
+
+    values = (
+        cache_creation.get("ephemeral_5m_input_tokens"),
+        cache_creation.get("ephemeral_1h_input_tokens"),
+    )
+    if not any(value is not None for value in values) or not all(
+        isinstance(value, int) and not isinstance(value, bool) and value >= 0
+        for value in values
+        if value is not None
+    ):
+        return None
+
+    breakdown = (values[0] or 0, values[1] or 0)
+    return breakdown if sum(breakdown) > 0 else None
 
 
 def _get_tokens_source(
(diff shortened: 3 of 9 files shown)


gouveags:
@greptile review

posthog/test/ai/anthropic/test_anthropic.py
```diff
@@ -78,6 +78,14 @@ def create_mock_response(**kwargs):
     return MockResponse(**kwargs)
 
 
+def assert_cache_creation_ttl_breakdown_preserved(props):
+    assert props["$ai_cache_creation_input_tokens"] == 800
```
carlos-marchal-ph, line 82: Worth also asserting `"$ai_cache_creation_5m_input_tokens" not in props` (and the 1h twin) here. The native wrappers deliberately leave the top-level pair to server-side `$ai_usage` extraction, and that absence is part of the contract in the same way `test_cached_tokens` now locks it in for the aggregate-only case.

richardsolomou (approved):
This looks good, thank you! Could you also open a targeted posthog-js follow-up for the equivalent gaps there? Vercel already preserves the raw usage, but Anthropic streaming and the LangChain integration appear to drop the TTL breakdown.

gouveags:
@richardsolomou I opened the targeted JS follow-up in PostHog/posthog-js#4236. Thanks for pointing me in that direction!

marandaneto:
> Commits must have verified signatures.

@gouveags can you fix your commits and force push?

.sampo/changesets/anthropic-cache-ttl-breakdown.md
```diff
@@ -0,0 +1,5 @@
+---
+pypi/posthog: patch
+---
+
+Preserve Anthropic cache-write TTL breakdowns across Python SDK AI integrations.
```
carlos-marchal-ph: This adds two public attributes to `ModelUsage` plus new event properties, which reads as minor rather than patch to me. Deferring to Client Libraries on the convention here.

posthog/ai/claude_agent_sdk/processor.py
```diff
@@ -511,6 +516,8 @@ def _emit_generation_from_result(
             properties["$ai_cache_read_input_tokens"] = cache_read
         if cache_creation:
             properties["$ai_cache_creation_input_tokens"] = cache_creation
+        if usage:
+            properties["$ai_usage"] = usage
```
carlos-marchal-ph: This stores `result.usage` by reference, while the stream path copies with `dict(usage)` at message_start. Capture only enqueues the properties dict, so a caller mutating `result.usage` afterwards could alter the payload before it is serialised. Wouldn't `dict(usage)` keep parity?

posthog/ai/utils.py
```diff
@@ -21,12 +21,36 @@
...
+        "$ai_cache_creation_5m_input_tokens",
+        "$ai_cache_creation_1h_input_tokens",
         "$ai_total_tokens",
         "$ai_reasoning_tokens",
     }
 )
 
 
+def _extract_cache_creation_ttl_breakdown(
+    cache_creation: Any,
```
carlos-marchal-ph: The parameter is named `cache_creation`, but the only call site passes LangChain's flattened `input_token_details`, where `cache_creation` is a sibling aggregate int rather than the nested Anthropic object. It works because only the two ephemeral keys are read, but the name invites passing the wrong dict.

Maybe rename to `token_details`, or note both accepted shapes in the docstring?

carlos-marchal-ph (approved):

gouveags:
@marandaneto Sorry about the unsigned commit. I’ve rebuilt the branch with a verified signature. Could you check it again? Thanks for the heads-up. 

Does this guideline apply across all PostHog repositories, or is it specific to this one? (asking because in other repos nobody noticed this).

posthog/test/ai/anthropic/test_anthropic.py
gouveags: Added the negative assertions to the shared native-wrapper test helper. The sync, async, streaming, and Bedrock cases now verify that the TTL fields stay under `$ai_usage.cache_creation` and are not duplicated at the top level.

.sampo/changesets/anthropic-cache-ttl-breakdown.md
gouveags: Agreed. The public `ModelUsage` additions make this a backward-compatible API expansion, so I changed the changeset from `patch` to `minor`.

posthog/ai/claude_agent_sdk/processor.py
gouveags: Copied `result.usage` before attaching it to `$ai_usage`, matching the streaming path. The regression test verifies that a later top-level mutation to the returned usage dict does not alter the captured properties.

posthog/ai/utils.py
gouveags: Renamed the parameter to `token_details` and updated the docstring to match the flattened LangChain `input_token_details` passed by the call site.

gouveags:
@carlos-marchal-ph all fixed, let me know if you notice something else that needs polishing! Thanks for the feedback!

.sampo/changesets/anthropic-cache-ttl-breakdown.md
```diff
@@ -0,0 +1,5 @@
+---
+pypi/posthog: minor
```
marandaneto, line 2: is that a fix or new feat? title is fix, but the changeset is minor

gouveags: @marandaneto I’m not entirely sure how the PR type and changeset level are expected to line up here, so I’d appreciate some guidance. I originally used `fix` because this came from a bug I ran into while using the SDK at work ([#745](https://github.com/PostHog/posthog-python/issues/745)): the TTL breakdown was being dropped before ingestion (Anthropic internals)

I changed the Sampo changeset from `patch` to `minor` after @carlos-marchal-ph pointed out that the two new public `ModelUsage` attributes and event properties make this look like a backward-compatible API addition.

Would you keep the `fix` title with a minor changeset in this case, or classify the PR differently?

gouveags: Maybe it's a question for both, not just for Manoel

gouveags:
@richardsolomou As the [original PR ](https://github.com/PostHog/posthog/pull/71323) was merged, I think that we're free to merge this one too, unless @marandaneto and @carlos-marchal-ph have any further requests.

.sampo/changesets/anthropic-cache-ttl-breakdown.md
marandaneto: yeah i missed the new model usage attributes, lets keep as is

PR #748, opened 2026-07-16
feat(flags): minimize $feature_flag_called events for non-experiment flags

haacked:

`$feature_flag_called` events carry the full enriched properties dict (super properties, context tags, custom event properties, system context) on every flag call, even for the vast majority of flags that aren't linked to an experiment. This PR minimizes those events for non-experiment flags when the server enables it, cutting event size and ingestion cost without losing anything needed for exposure analysis.

The event is trimmed to a strict allowlist iff the server-controlled gate is on (`minimalFlagCalledEvents` in the v2 `/flags` response, or `minimal_flag_called_events` in the local-evaluation payload) and the evaluated flag's `has_experiment` is exactly `false`. Any missing signal, gate absent, `has_experiment` unknown, or the flag linked to an experiment, sends the full legacy shape unchanged. The gate is stored alongside cached flag state so it survives polls, 304 responses, and external `flag_cache` round-trips.

The stale-cache fallback path (`_get_stale_flag_fallback`) never populates `flag_details`, so `has_experiment` stays unknown there and those events correctly fall back to the full shape instead of risking an incorrect minimization.

The reason for the server-gate is we can't just roll this out to everyone. It could break existing insights, etc. We're going to have to make an announcement first and give people time to adjust.

However, we can use the gate to roll this new behavior out for new teams (for example) after we figure out comms for that. The main goal is to establish `$feature_flag_called` as a special system event and not a general event given we don't charge for these events.

`uv run pytest posthog/test/test_feature_flag_called_minimization.py posthog/test/test_flag_definition_cache.py` (49 tests). The new test file covers the gate on/off, `has_experiment` true/false/missing, local vs. remote evaluation, group context preservation, and the async cache provider with the gate enabled.

## :pencil: Checklist

- [ ] I reviewed the submitted code.

### If releasing new changes


posthog/client.py
```diff
@@ -2743,17 +2815,30 @@ def _capture_feature_flag_called_if_needed(
...
-        # Omitted when unknown (older servers that don't report has_experiment).
+        # Record the server's experiment signal when known, so minimization's impact
+        # can be measured by segmenting on it.
         if has_experiment is not None:
             properties["$feature_flag_has_experiment"] = has_experiment
 
+        # Minimize iff the server-controlled gate is on AND the flag is known to have
+        # no linked experiment. Any missing signal (gate absent, has_experiment
+        # missing) fails safe to the full legacy shape.
+        should_minimize = self._minimal_flag_called_events and has_experiment is False
```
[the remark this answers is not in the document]
haacked: Good catch. Fixed in b6b03f8: the gate is now pinned on the `FeatureFlagEvaluations` snapshot at construction and threaded through to the capture funnel, so a deferred flag access fires an event shaped by the evaluation that produced it — not the client-wide gate at send time. The single-flag path is unchanged since it fires immediately after evaluation. Regression tests cover both directions of the timing window.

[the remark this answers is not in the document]
haacked: Fixed by the same change as the thread above (b6b03f8): the snapshot carries the gate it was created with, so a later response disabling the gate no longer reshapes this snapshot's events. Test: `test_snapshot_keeps_pinned_gate_when_later_response_disables_it`.

posthog/client.py
```diff
@@ -1980,6 +2049,7 @@ def _fetch_feature_flags_from_api(self):
                                 "flags": self.feature_flags or [],
                                 "group_type_mapping": self.group_type_mapping or {},
                                 "cohorts": self.cohorts or {},
+                                "minimal_flag_called_events": self._minimal_flag_called_events,
                             }
```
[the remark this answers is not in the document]
haacked: This is theoretical rather than a practical break: `FlagDefinitionCacheData` is a TypedDict, so there's no runtime validation on our side, and a provider strict enough to reject unknown keys would break on any future field addition — the payload has never carried a stability guarantee of that kind. The new key also round-trips through the equivalent cache providers in the other SDKs' implementations of this contract. The changeset notes the payload now includes `minimal_flag_called_events`, which covers the realistic migration concern.

posthog/client.py
```diff
@@ -967,7 +1013,14 @@ def _get_flags_decision(
             **request_data,
         )
 
-        return normalize_flags_response(resp_data)
+        response = normalize_flags_response(resp_data)
+        # Server-controlled gate for minimal $feature_flag_called events. Only the
+        # v2 response shape carries it; absent (legacy shape, older server, team
+        # not gated) means False.
+        self._minimal_flag_called_events = (
```
dustinbyrne, line 1030: **[P2/design] Cross-SDK gate ownership differs during mixed local/remote evaluation**

Python stores `minimalFlagCalledEvents` in client-wide last-writer-wins state, matching the open Node and PHP implementations. Go instead carries the gate with the particular local or remote evaluation result.

These behave identically after the team-level setting has propagated everywhere, but differ briefly during rollout, stale definition-cache periods, or overlapping responses. Python/Node/PHP use the latest observed gate for subsequent events, while Go uses the gate bundled with each evaluation source.

This is not necessarily a bug, but could we confirm the intended cross-SDK contract? Consider documenting the chosen policy or adding equivalent transition tests so the divergence is explicit.

posthog/client.py
```diff
@@ -199,6 +199,48 @@ def wrapper(self, *args, **kwargs):
     return decorator
 
 
+# Strict allowlist for minimal ``$feature_flag_called`` events, per the cross-SDK
```
dustinbyrne, line 202: **[P2/design] Confirm that static runtime properties are intentionally outside the minimal contract**

posthog-python adds `$os`, `$os_version`, `$os_distro`, `$python_runtime`, and `$python_version` to every event through `system_context()`, but none are in this allowlist. They are therefore absent from the submitted event and are not reconstructed during ingestion.

This matches the changeset’s strict-system-event model, but these are small, relatively static dimensions compared with super properties and flag-state blobs. Could we confirm that platform/runtime breakdowns are intentionally unsupported for minimized `$feature_flag_called` events? If they should remain available, consider adding the relevant dimensions to the canonical cross-SDK allowlist rather than handling Python specially.

dustinbyrne (approved):
Human-driven, agent-assisted review.

posthog/client.py
haacked: Good catch. I checked every PostHog SDK before answering (js, ios, android, flutter, php, ruby, go, dotnet, elixir). What each one auto-attaches splits pretty cleanly into three groups.

There's SDK identity: `$lib`, `$lib_version`, `$is_server`. Already allowlisted, same everywhere, not really in question.

Then there's cheap runtime/OS identity: `$os`, `$os_version`, `$os_distro`, plus whatever language-runtime version the SDK tracks (`$python_runtime`/`$python_version` here, `$go_version` in Go). Low cardinality, static for the life of the process, and actually useful when you're trying to figure out why a flag behaves differently on one platform. This is the bucket you're asking about.

And then there's the rich stuff: mobile device/app fingerprints on iOS/Android, full browser UA/viewport/screen data on JS. That one I'd leave out of a minimal event without much debate, nobody's reconstructing platform breakdowns from screen dimensions.

So: added `$os`, `$os_version`, `$os_distro`, `$python_runtime`, and `$python_version` to the allowlist here. They're static scalars, not device/browser blobs, so keeping them doesn't undercut the point of minimizing the event.

You're also right that this shouldn't be a Python-only call. None of the other SDKs have shipped minimization yet, so there's no existing cross-SDK allowlist to defer to, this PR is effectively the first draft of one. I'm making the same addition on the Go and Node branches. Go's system context is exactly `$os`/`$os_version`/`$os_distro`/`$go_version`, so it gets identical treatment. Node has nothing to add since it never reads `process.version` or `os.platform()` to begin with.

.NET turned out to be its own thing. Its `Capture()`/batch path, the one `$feature_flag_called` actually uses, doesn't attach `$os`/`$framework`/`$arch` at all right now, even on full events. That only happens on the `Identify`/`GetFeatureFlags` paths. So there's nothing to preserve through minimization there yet. Adding it just to the minimal shape would leave minimized events carrying more platform info than full ones, which is a separate bug worth its own fix, not something I want this allowlist to paper over.

posthog/client.py
haacked: Fair point. I fixed it instead of just writing it up. Both the single-flag path and evaluate_flags() snapshots now pull the gate from whichever evaluation actually produced the value, either the poller's local state or that specific `/flags` response, not a second read of the client-wide attribute afterward. Local eval still falls back to poller state since there's nothing more specific to tie it to there, same as Go's own local-eval path.

PR #753, opened 2026-07-20
feat(metrics): support metrics config via module-level settings

DanielVisca:

Apps configured through module attributes (`posthog.api_key = ...` + implicit `setup()`) — including PostHog's own Django monorepo via `posthoganalytics` — have no way to pass the `metrics` config dict. Every series they record through `client.metrics` ships `service.name='unknown_service'`, which defeats the Metrics UI's service filtering.

This adds the module-level `metrics` attr (mirroring how every other `Client` kwarg is exposed as a module setting) and passes it through `setup()`. Because the config is consumed lazily on first `.metrics` access, a repeat `setup()` call also refreshes it until the metrics API is first used — Django `ready()` hooks that run after something already forced `setup()` still apply.

Groundwork for the monorepo's bump-ready `client.metrics` usage from web/celery (companion PR in PostHog/posthog).

- Two new tests in `posthog/test/test_metrics.py`, red before the change (`service.name` came back `unknown_service`): module attr flows into the `setup()` client's OTLP resource attributes, and late assignment applies on repeat `setup()` until first use.
- `pytest --timeout=30`: 1640 passed; the 9 failures are the pre-existing network-dependent `posthog/test/ai/*` integration tests (verified they fail identically on a clean checkout in the same sandbox).
- `ruff format --check`, `ruff check`, `mypy | mypy-baseline filter` (no issues), `python -W error -c "import posthog"` all clean.
- End to end: with this patch, a module-configured client in a forked child (celery-prefork shape) recorded and flushed real metrics that arrived in the Metrics product with the configured service name.

## :pencil: Checklist

### If releasing new changes

- [x] Ran `sampo add` to generate a changeset file (authored directly in `.sampo/changesets/module-level-metrics-config.md` — same format)

## 🤖 Agent context

Written by Claude (PostHog Code session) while making the PostHog monorepo bump-ready for `client.metrics`. TDD: the two tests were written and observed failing before the implementation. The late-assignment refresh in `setup()` follows the existing "always set in case user changes it" post-construction block, and is limited to before first `.metrics` use so an active aggregation window is never reconfigured mid-flight.

---

### DIFF
--- a/.sampo/changesets/module-level-metrics-config.md
+++ b/.sampo/changesets/module-level-metrics-config.md
@@ -1,0 +1,5 @@
+---
+'pypi/posthog': minor
+---
+
+The `client.metrics` config can now be set through module-level settings: assign `posthog.metrics = {"service_name": ..., ...}` alongside `posthog.api_key` and the dict is applied when `setup()` builds the global client. Previously module-configured apps had no way to pass the metrics config, so every series recorded through the global client shipped `service.name='unknown_service'`. Late assignment (e.g. a Django `ready()` hook running after an early `setup()`) still applies on the next `setup()` call, as long as the metrics API hasn't been used yet.
--- a/posthog/__init__.py
+++ b/posthog/__init__.py
@@ -329,6 +329,10 @@ Attributes:
         requests after network, transport, or timeout failures. Defaults to 1.
         Set to 0 to disable retries.
     super_properties: Properties merged into every captured event.
+    metrics: Config dict for the ``client.metrics`` API (``service_name``,
+        ``service_version``, ``environment``, ``flush_interval``, ...). Applied
+        when ``setup()`` builds the global client, or on a later ``setup()``
+        call if the metrics API hasn't been used yet.
     enable_exception_autocapture: Automatically capture uncaught exceptions.
     log_captured_exceptions: Also log exceptions captured by error tracking.
     before_send: Optional callback that can modify or drop events before upload.
@@ -374,6 +378,7 @@ is_server = True  # type: bool
 feature_flags_request_timeout_seconds = 3  # type: int
 feature_flags_request_max_retries = 1  # type: int
 super_properties = None  # type: Optional[Dict]
+metrics = None  # type: Optional[Dict]
 enable_exception_autocapture = False  # type: bool
 log_captured_exceptions = False  # type: bool
 # Used to determine in app paths for exception autocapture. Defaults to the current working directory
@@ -1169,6 +1174,7 @@ def setup() -> Client:
             feature_flags_request_timeout_seconds=feature_flags_request_timeout_seconds,
             feature_flags_request_max_retries=feature_flags_request_max_retries,
             super_properties=super_properties,
+            metrics=metrics,
             # TODO: Currently this monitoring begins only when the Client is initialised (which happens when you do something with the SDK)
             # This kind of initialisation is very annoying for exception capture. We need to figure out a way around this,
             # or deprecate this proxy option fully (it's already in the process of deprecation, no new clients should be using this method since like 5-6 months)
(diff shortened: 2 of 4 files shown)


DanielVisca:
## sdk-dustin review — single pass

*Fully automated AI review; comments were not individually reviewed by a human before posting.* Applied `sdk-dustin-code-review` + `sdk-dustin-sdk-review` (single pass + calibration, per author request — no fix loop needed).

### Verdict
**APPROVE** (agent pre-review; Client Libraries review still required) — no BLOCKING/IMPORTANT findings. One design question flagged below for a deliberate call before this ships in a minor release.

### Existing supported workaround / API necessity
- **Desired outcome:** module-configured apps (`posthog.api_key = ...` + implicit `setup()`) pass the `metrics` config so series don't ship `service.name='unknown_service'`.
- **Mechanisms searched:** all module attrs mirrored into `setup()`; direct client construction; post-hoc mutation of `default_client`.
- **Strongest workaround:** `posthog.default_client = Client(api_key, metrics={...})` — validated against the checkout, but it bypasses `setup()`'s wiring (`disabled`/`debug`/`before_send` sync with module attrs), so **workaround exists but has meaningful gaps**.
- **Conclusion:** the addition follows the SDK's uniform convention (every `Client` kwarg is exposed as a same-named module setting); justified. But see the naming flag.

### Review comment drafts
**[P2/design] `posthog.metrics` as a config attr spends the natural name of a future module-level metrics API.**
`posthog/__init__.py:381`. The module exposes client APIs as proxy functions (`posthog.capture(...)`, `posthog.flush(...)`); there is no metrics accessor yet. If one is ever added, the ergonomic, posthog-js-mirroring name is `posthog.metrics.count(...)` — which this attr forecloses: after this ships (minor release), `posthog.metrics` is publicly a `Optional[dict]`, and `posthog.metrics.count` is an `AttributeError` on `None`. The kwarg-mirroring convention argues for the current name; reserving the accessor name argues for e.g. a `metrics` key consumed from a differently-named attr, or accepting that module-level users will always go through `setup().metrics`. Consider making this an explicit decision (a one-line note in the changeset would do) rather than an accident of convention.

### Criteria summary
- **Backwards compatibility** ✅ — pure addition; default `None` preserves existing behavior; snapshot updated.
- **API surface** ⚠️ one new public module attr (see design flag); consistent with the established kwarg-mirroring pattern.
- **Entry-point tracing** ✅ — `setup()` fresh-construct path passes the kwarg; repeat-`setup()` path applies via `_metrics_config` only while `_metrics is None`, matching the documented "until first use" contract. (The guard is belt-and-braces: config is only read at first `.metrics` access anyway.)
- **Performance** ✅ — config plumbing only.
- **Test coverage** ✅ — both paths executed locally (fresh setup + late assignment), module state restored; 36/36 suite green on this branch.
- **No throwing** ✅ — invalid dicts degrade inside `PostHogMetrics` (hardened further by #742).
- **Cross-SDK parity** ✅/note — posthog-js configures metrics at init too; the naming flag above is where parity pressure will eventually appear.

### Non-blocking observations
- This branch predates #742's metrics hardening; both PRs edit `test_metrics.py` in disjoint regions, so whichever merges second should rebase clean (merge #742 first — it's further along).
- Concurrent first-`.metrics` access racing a repeat `setup()` can silently drop the late config — same outcome as the documented "already used" case; startup-time only, not material.

dustinbyrne (approved):
recommend adding a client constructor argument for this as well

otherwise, users who don't use the default global `posthog` cannot set this (unless it already exists, i didn't check). if you do add it, make sure it's positioned as the last argument in the constructor for backwards compatibility

DanielVisca:
> recommend adding a client constructor argument for this as well
> 
> otherwise, users who don't use the default global `posthog` cannot set this (unless it already exists, i didn't check). if you do add it, make sure it's positioned as the last argument in the constructor for backwards compatibility

Already there 👍🏽

PR #760, opened 2026-07-22
feat(aio): dedicated AI capture lane and multimodal passthrough

carlos-marchal-ph:

AI Observability needs multimodal (base64 media) capture, which doesn't fit the analytics `/batch/` pipeline or its size limits. This PR adds a dedicated AI capture lane and a private multimodal switch, all underscore-private until beta:

- `Client._capture_ai()` — a dedicated private method that is the only client-level entry to the AI lane (posts to `/i/v0/ai/batch/`, 8MiB per-event ceiling vs 900KiB for analytics, lazy-started consumers, fork-safe, lane-aware flush/join/shutdown, pinned to the V0 wire protocol). `capture()` is never rerouted and no name-based `$ai_*` routing remains anywhere.
- A private client-level switch `Posthog(..., _use_ai_lane=True)` routes all AI SDK wrapper events through the AI lane. Wrapper emissions funnel through `_capture_ai_event`, which uses `_capture_ai` when the client opted in and falls back to `capture()` for duck-typed clients (so downstream test suites passing `Mock` clients keep working).
- Both switches are also settable as module attributes (`posthog._use_ai_lane`, `posthog._enable_multimodal_capture`), synced onto the default client by `setup()` on every call — deployments that configure PostHog via module attributes (like our monorepo's lazily auto-instantiated `posthoganalytics` client) set one attribute instead of mutating a constructed client.
- A private `Posthog(..., _enable_multimodal_capture=True)` switch additionally skips base64 media redaction and implies `_use_ai_lane`, so unredacted multi-MB events can never land on the analytics lane's size cap. It replaces the `_INTERNAL_LLMA_MULTIMODAL` env var, which is removed.
- Routing and redaction-skip are hardened against unspecced `Mock` clients (`is True` guards); `Consumer` is generalized to `endpoint=` / `max_msg_size=` and no longer knows AI events exist; dead `is_ai_event` removed; oversized-event drops log name and byte size only, never event contents.

- New `posthog/test/test_ai_capture_lane.py`: lane routing, sync mode, size caps, V0 pinning, lazy start, concurrent first-capture race, fork rebuild, disabled-client behavior, module-attribute config, Mock hardening.
- Per-wrapper routing tests assert `_capture_ai` is used (and `capture` is not) when the client opts in; wrapper-level multimodal passthrough tests for OpenAI and Anthropic.
- Locally: 622 passed / 60 optional-SDK skips across client, consumer, lane, fork, and `posthog/test/ai` suites; `ruff format --check` and mypy clean; public API snapshot check passes.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

**Autonomy:** Human-driven (agent-assisted) — assign @carlos-marchal-ph as DRI.

Built in a Claude Code session directed by Carlos, who made all design decisions (client-level switch over per-wrapper flags, module-attribute config for the monorepo's auto-instantiated default client, removing the `_INTERNAL_LLMA_MULTIMODAL` env var, keeping the 8MiB client-side cap). The final branch was verified by two independent whitebox review subagents running the pr-review skill; both confirmed the intended architecture with no blocking findings, and their remaining suggestions (log hygiene on oversized drops, dead code, changeset notes, extra tests) are incorporated.


posthog/client.py
```diff
@@ -631,6 +796,18 @@ def __init__(
...
 
+    @property
+    def queue(self) -> Queue:
+        """The analytics lane's queue (kept for backwards compatibility)."""
+        return self._analytics_lane.queue
+
+    @property
+    def consumers(self) -> Optional[List[Consumer]]:
+        """Flat list of the lanes' consumers, analytics first (kept for backwards compatibility)."""
+        if self.sync_mode:
```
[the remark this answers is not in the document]
carlos-marchal-ph: Reads behave exactly as before; assignment was never a supported configuration path and now fails loudly instead of silently desyncing the client from its consumers.

posthog/client.py
```diff
@@ -251,6 +250,157 @@ def _metadata_has_experiment(metadata: Any) -> Optional[bool]:
...
+
+    def join(self) -> None:
+        """Pause this lane's consumers and wait for them to exit; a never-started lane is a no-op."""
+        for consumer in self.consumers:
+            consumer.pause()
+            try:
+                consumer.join()
+            except RuntimeError:
+                # consumer thread has not started
+                pass
```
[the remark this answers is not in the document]
carlos-marchal-ph: This matches the pre-existing contract on main: `capture()` after `join()` strands events on the analytics queue the same way, since consumers are paused. Capture-after-shutdown isn't supported; resetting `_started` in `join()` would make a post-shutdown capture silently respawn threads, which is worse.

posthog/client.py
```diff
@@ -323,8 +473,9 @@ def __init__(
         capture_mode: Optional[Union[CaptureMode, str]] = None,
         capture_compression: Optional[Union[CaptureCompression, str]] = None,
         secret_key=None,
-        _dedicated_ai_endpoint=False,
         metrics: Optional[dict] = None,
+        _use_ai_lane=False,
+        _enable_multimodal_capture=False,
```
dustinbyrne, line 486: what will it take to stabilize these? what's the reasoning for keeping these configurable or not defaulting them to `True`?

posthog/client.py
```diff
@@ -1089,6 +1266,33 @@ def capture(
         Category:
             Capture
         """
+        return self._capture(event, self._analytics_lane, **kwargs)
+
+    @no_throw()
+    def _capture_ai(
```
dustinbyrne, line 1280: similar question - do we need to release this as an internal method? if it's already working as intended, consider public release

posthog/ai/sanitization.py
```diff
@@ -1,18 +1,15 @@
...
 REDACTED_IMAGE_PLACEHOLDER = "[base64 image redacted]"
 
 
-def _is_multimodal_enabled() -> bool:
-    """Check if multimodal capture is enabled via environment variable."""
-    return os.environ.get("_INTERNAL_LLMA_MULTIMODAL", "").lower() in (
-        "true",
-        "1",
-        "yes",
-    )
```
dustinbyrne, line 15: anything we need to update downstream in response to this?

dustinbyrne (commented):
overall I think this direction looks great

posthog/client.py
```diff
@@ -251,6 +250,157 @@ def _metadata_has_experiment(metadata: Any) -> Optional[bool]:
...
+                self.consumers.append(consumer)
+
+                if self.send:
+                    consumer.start()
+            self._started = True
+
+    def enqueue(self, msg) -> bool:
+        """Queue `msg` for upload, starting the lane on its first event."""
+        if not self._started:
+            self.start()
```
dustinbyrne, line 349: [P2] Prevent a never-used AI lane from starting after shutdown

This is distinct from the withdrawn `_started=True` case raised by greptile. If the AI lane has never been used, `shutdown()` leaves it with `_started=False`. A late wrapper capture then reaches this branch, calls `start()`, creates a new consumer, and performs an HTTP submission.

I reproduced this with `Client(..., _use_ai_lane=True)`: after shutdown there were no AI consumers, but `_capture_ai_event()` created one live consumer and sent the event. This can happen when an in-flight LLM wrapper finishes during teardown.

I’m not suggesting resetting `_started` in `join()`. Could we instead add a terminal closed state, synchronized with `_start_lock`, so neither lane can accept or start work after shutdown? A regression test should assert that a late AI capture creates no consumer and makes no request.

— Agent-led review, human-reviewed before posting.

posthog/client.py
```diff
@@ -323,8 +473,9 @@ def __init__(
         capture_mode: Optional[Union[CaptureMode, str]] = None,
         capture_compression: Optional[Union[CaptureCompression, str]] = None,
         secret_key=None,
-        _dedicated_ai_endpoint=False,
         metrics: Optional[dict] = None,
+        _use_ai_lane=False,
```
dustinbyrne, line 485: [P1] Coordinate this removal with the current `posthoganalytics` consumer

The generated `posthoganalytics` package inherits this constructor, but the current PostHog app still unconditionally calls `Posthog(..., _dedicated_ai_endpoint=False)`. On this head that raises:

```text
TypeError: Client.__init__() got an unexpected keyword argument '_dedicated_ai_endpoint'
```

Its default-client helper also mutates the old client/consumer attributes, which Python still accepts but the new routing code ignores, so fixing construction alone would still leave wrapper events on `/batch/`. The relevant callers are in [`posthog/ph_client.py`](https://github.com/PostHog/posthog/blob/a41757be988989441abd0691e757c1ed503f42fd/posthog/ph_client.py#L34-L52) and [`get_client()`](https://github.com/PostHog/posthog/blob/a41757be988989441abd0691e757c1ed503f42fd/posthog/ph_client.py#L120-L144).

The option is private, but this is a verified first-party integration. Could we land a coordinated, feature-detected app migration before this mirror is adopted, or retain a temporary compatibility bridge? Generating `posthoganalytics` and running `posthog/test/test_ph_client.py` against it would cover both construction and routing.

— Agent-led review, human-reviewed before posting.

posthog/client.py
carlos-marchal-ph: We have not tested the capture + ingestion + CH side of this yet, in fact part of it is still in PRs awaiting merge. We'll roll this out to team 2 first, try it for a bit, and if the shape looks right and we don't have to change anything, we'll make them public as an opt-in beta and let people self onboard while we monitor load metrics. Eventually we'll default them to yes.

There's basically no way for us to roll this out slowly from the backend since it changes which endpoint requests hit, so it happens before we can "feature flag" it or roll it out.

posthog/client.py
carlos-marchal-ph: Ditto

posthog/ai/sanitization.py
carlos-marchal-ph: Nope we were using it just for team 2. This is what I meant by keeping stuff private for now, allows us to quickly iterate on the exact shape / config while we figure out what works for us.

posthog/client.py
carlos-marchal-ph: I'm gonna merge this in tandem with a followup on the monorepo side that updates the version and the instantiation sites too. It's part of a larger coordinated migration to start using the internal multimodal pipeline so I'll be on top of it coordinating it, I think the risk is quite small, but I guess I could add more resilient code on the monorepo side just in case. Thoughts?

posthog/client.py
carlos-marchal-ph: Fixed!

dustinbyrne (approved):

PR #761, opened 2026-07-24
feat(mcp): stateless and multi-pod server support

gesh:
## Problem

Stateless / multi-pod MCP servers issue no session id. So `$session_id` **fragments** (each pod generates its own) and the client identity — the "harness" (Claude Code, Cursor, …), sent only at `initialize` — is **lost** on any pod that never processed the handshake. Fine in single-pod staging, silently wrong in multi-pod production.

## Fix

A self-encoded session token, wire-compatible with the `@posthog/mcp` TypeScript SDK. At `initialize` we mint the `Mcp-Session-Id` response header as `base64url(JSON{sid,cn,cv,pv})`; the client replays it on every request, so any pod recovers the same session id + client name/version with **no shared state**.

Mint and decode are split because Python's MCP SDK owns `initialize` in its runner layer and forbids overriding it via `request_handlers` (unlike the TS SDK) — so there's no in-SDK seam to mint from:

- **Mint (write):** on the `instrument()` FastMCP path this is **zero-config** — `instrument()` auto-wires the mint into the server's ASGI-app factories (`streamable_http_app` / `sse_app`, which `mcp.run()` also uses). For a custom `PostHogMCP` dispatcher (you own the ASGI app), add `PostHogMcpStatelessSessionMiddleware` once. No manual header handling either way.
- **Decode (read):** in the SDK on every request — the `instrument()` adapters decode the replayed header and backfill client identity; `PostHogMCP` users read `get_mcp_session(request)`. The token session is resolved **per request** (never stored on the shared per-server state), so concurrent clients can't merge under one `$session_id`.

New files: `posthog/mcp/session_token.py` (codec) and `posthog/mcp/asgi.py` (middleware, `get_mcp_session`, `instrument()` auto-wire). Also touched: `session.py` / `_internal.py` (a `"token"` session source) and the two `instrument()` adapters. Example: `examples/mcp_stateless.py`.

The middleware is fail-safe by contract — any error while sniffing/minting is logged and the request passes through untouched (memory-bounded body buffering; broad guards against adversarial input).

## Validation

Unit + real stateless-FastMCP integration tests (`posthog/test/mcp/test_session_token.py`), including the zero-config auto-wire, a frozen cross-SDK wire-format fixture, and adversarial-input pass-through. Also verified against a **2-pod stateless cluster behind round-robin nginx**: with `instrument()` alone, all events across both pods share one `$session_id` and keep the harness — while the published package fragments both.

---


posthog/mcp/asgi.py
```diff
@@ -0,0 +1,248 @@
...
+        if message.get("type") != "http.request":
+            # A non-body message (e.g. http.disconnect); stop and replay what we have.
+            break
+        chunks.append(message.get("body", b"") or b"")
+        more = message.get("more_body", False)
+
+    # Always buffer the whole body so replay is byte-faithful (Starlette's own
+    # Request.body() buffers fully too). This holds the request in memory, which
+    # is fine for tiny JSON-RPC POSTs on an MCP endpoint.
+    buffered = b"".join(chunks)
```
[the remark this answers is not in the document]
gesh: Fixed in c7ab522: `_buffer_body` now stops at `_MAX_SNIFF_BODY`. Once a session-less POST exceeds the cap it can't be an `initialize` handshake, so we stop buffering and stream the remaining chunks straight through instead of accumulating them — memory is bounded to the cap.

posthog/mcp/session.py
```diff
@@ -29,13 +30,42 @@ def derive_session_id_from_mcp_session(mcp_session_id: str) -> str:
...
+            data.token_protocol_version = token.protocol_version
+            data.last_activity = now
+            return data.session_id
+
+        # A token session, like an MCP-derived one, lives as long as the client
+        # replays it -- keep the id even on a request that arrives without it, so
+        # the session doesn't fragment.
+        if data.session_source == "token":
+            data.last_activity = now
+            return data.session_id
```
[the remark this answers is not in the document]
gesh: Fixed in c7ab522: removed the sticky-token fallback. `data` is shared by every client on a server instance, so a request that doesn't replay the token no longer inherits the previous client's token session — token sessions are resolved per request, and the memory fallback only persists a genuine `generated` session (a leftover token/mcp session can't leak into it).

posthog/mcp/_internal.py
```diff
@@ -60,8 +60,14 @@ class MCPAnalyticsData:
...
     session_id: str = ""
-    session_source: str = "generated"  # "generated" | "mcp"
+    session_source: str = "generated"  # "generated" | "mcp" | "token"
     last_mcp_session_id: Optional[str] = None
+    # Client identity recovered from a self-encoded session token (see
+    # session_token.py). On a stateless pod that never processed `initialize`,
+    # the live `client_params` is empty, so these are the only harness source.
+    token_client_name: Optional[str] = None
+    token_client_version: Optional[str] = None
+    token_protocol_version: Optional[str] = None
```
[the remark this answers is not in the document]
gesh: Fixed in c7ab522: removed the `token_client_name/version/protocol_version` fields. Client identity is recovered per request in the adapters (`resolve_session_and_client`), not from shared state — so these were dead. (Storing them on the shared `data` would have the same cross-client leak as the session finding, so removing is the right call.)

pauldambra (approved):
approval to unblock
a bunch of robot comments to address though

thanks for sorting this so quickly!

pauldambra:
i'm also intrigued to see what review hog manages here @gesh

posthog/mcp/session_token.py
```diff
@@ -0,0 +1,136 @@
+"""Self-encoded session tokens for stateless / multi-pod MCP servers.
+
+A stateless server keeps nothing between requests, so every request starts a
+new session and the client name/version (only sent at ``initialize``) is lost.
+The one value clients replay on every request is the ``Mcp-Session-Id`` header.
+So at ``initialize`` we mint that header as a token carrying the session id and
+client identity -- any pod can read them back from the header alone.
+
+The token is unsigned: it holds only what the client already self-reports. It is
+wire-compatible with the TypeScript SDK (same short keys ``sid``/``cn``/``cv``/``pv``),
```
DanielVisca, line 10: The whole point of this token is that it decodes in both this SDK and the TypeScript one (same sid/cn/cv/pv keys, base64url). But nothing in the suite actually protects that. If either side renames a key or changes the encoding, every test still passes and the two SDKs quietly stop reading each other's tokens. Consider one fixture test: a token string minted by the TS SDK, hardcoded here, asserted to decode into the expected fields, and ideally the reverse with a known Python token the TS side must read. That turns the compatibility claim into something CI defends.

posthog/mcp/asgi.py
```diff
@@ -0,0 +1,258 @@
...
+    value = scope.get(_SCOPE_KEY)
+    return value if isinstance(value, SessionTokenPayload) else None
+
+
+class PostHogMcpStatelessSessionMiddleware:
+    """ASGI middleware that mints a session token onto the ``Mcp-Session-Id``
+    response header at ``initialize`` (when the client sent none) and decodes the
+    replayed token on every request, exposing it via :func:`get_mcp_session`.
+
+    Fail-safe: any error while sniffing/minting is logged and the request passes
```
DanielVisca, line 76: The docstring says any error while sniffing or minting is logged and the request passes through untouched, but only the header read is wrapped in try/except. If `_buffer_body`'s receive() raises, or anything in the mint block throws, it propagates to the host instead of passing through. In practice a failing receive() is already a dead request so the blast radius is small, but the promise reads broader than what the code guarantees. Either tighten the wording or wrap the POST mint block in the same try/except so the failsafe matches what it claims.

posthog/mcp/__init__.py
```diff
@@ -44,6 +44,18 @@
 from .logger import log, set_logger
 from .posthog_mcp import PostHogMCP
 from .session import derive_session_id_from_mcp_session, new_session_id
+from .session_token import (
```
DanielVisca, line 47: Seven new public names land here: the middleware, get_mcp_session, encode_session_id, decode_session_id, read_mcp_session_header, SessionTokenPayload, MCP_SESSION_HEADER. The middleware and get_mcp_session clearly need to be public for the custom dispatcher story. I'm less sure encode_session_id and read_mcp_session_header need to be public on day one versus staying internal until a real custom HTTP layer needs them. Not a blocker, just worth a second look since public surface is easy to add and hard to walk back.

posthog/mcp/asgi.py
```diff
@@ -0,0 +1,258 @@
...
+    """Make stateless minting zero-config on the ``instrument()`` path.
+
+    Wraps a FastMCP server's ASGI-app factories so the app they build already has
+    :class:`PostHogMcpStatelessSessionMiddleware` applied. Covers both
+    ``server.streamable_http_app()`` / ``sse_app()`` and ``mcp.run(transport=...)``
+    (which calls those factories internally), so the user adds nothing.
+
+    No-op for servers without app factories (stdio / low-level ``Server``), and safe
+    if the middleware is also added manually (it only mints when none is present)."""
+    for attr in ("streamable_http_app", "sse_app", "http_app"):
```
DanielVisca, line 248: On the bundled MCP SDK FastMCP this is fine: http_app doesn't exist and the other two build distinct apps. On fastmcp 2.x though, streamable_http_app and sse_app are often thin aliases over http_app, so wrapping all three could add the middleware twice to one app. It stays correct thanks to the no clobber check on the response header and decode being idempotent, but you'd mint two session ids and buffer the body twice per initialize. Worth confirming against fastmcp 2.x, or wrapping only the factory that actually builds the app.

DanielVisca (approved):
Pulled this down and ran the full mcp suite, 133 passing, including the stateless FastMCP end to end test and the zero config autowire path, and the public API snapshot check is clean. 
Traced instrument() through the autowire into both adapters and the resolved session id is threaded as a local all the way to record_tool_call, so two requests sharing one pod can't stamp each other's session. 
A few non blocking thoughts below, none of them need to hold the merge.

DanielVisca:
The description points at examples/mcp_stateless_fastapi.py but the file that landed is examples/mcp_stateless.py. Worth fixing so nobody hunts for the wrong path.

posthog/mcp/asgi.py
```diff
@@ -0,0 +1,248 @@
...
+    if not isinstance(message, dict) or message.get("method") != "initialize":
+        return None
+    params = message.get("params")
+    params = params if isinstance(params, dict) else {}
+    client_info = params.get("clientInfo")
+    client_info = client_info if isinstance(client_info, dict) else {}
+    try:
+        return encode_session_id(
+            SessionTokenPayload(
+                session_id=new_session_id(),
```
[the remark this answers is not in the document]
gesh: Fixed in cce83af. `RecursionError` subclasses `RuntimeError`, so the narrow `(ValueError, TypeError)` missed it and it escaped the un-guarded call site. Broadened the `json.loads` guard and wrapped the whole buffer+mint block in `__call__` so *any* failure (hostile JSON or a raising `receive()`) logs and passes the request through untouched. Added a test that a ~40KB deeply-nested body passes through with no mint and no raise.

posthog/mcp/asgi.py
gesh: Fixed in cce83af — same change as the RecursionError finding: the buffer+mint block is now inside a try/except that falls through to the untouched app, so a raising `receive()` or anything in the mint path no longer propagates. The docstring promise now matches the code.

posthog/mcp/session_token.py
gesh: Done in cce83af: added `test_cross_sdk_wire_format_is_frozen` — it pins the exact base64url string for a known payload and asserts both directions (`encode == frozen` and `decode(frozen) == fields`). A key rename or encoding change on either side now fails CI instead of silently breaking cross-SDK reads. The string is the canonical compact-JSON wire the TS SDK also produces (same sid/cn/cv/pv order).

posthog/mcp/__init__.py
gesh: Trimmed in cce83af: dropped `read_mcp_session_header` from the top-level exports (it's an internal header-parsing helper — the middleware and `get_mcp_session` cover the documented story). Kept `encode_session_id`/`decode_session_id` as a symmetric codec pair (mirrors the TS SDK's exported `encodeSessionId`, for a custom HTTP layer that mints/decodes without our middleware).

posthog/mcp/asgi.py
gesh: Good catch — fixed in cce83af. Auto-wire now skips adding the middleware if the built app already carries it (`_app_already_wrapped` checks `user_middleware` / the raw-ASGI wrapper). So on fastmcp 2.x where `streamable_http_app`/`sse_app` delegate to `http_app`, the app is wrapped exactly once — no double buffer/mint. Added `test_autowire_does_not_double_add_middleware`.

PR #804, opened 2026-07-30
fix(ai): preserve Anthropic messages stream API

marandaneto:

PostHog's Anthropic wrapper returned tracked generators from `messages.stream()` instead of Anthropic's native stream managers. That broke drop-in compatibility for sync and async callers relying on the provider's `with`/`async with` contract, `text_stream`, response helpers, and close behavior. The async wrapper also incorrectly required awaiting `stream()`.

This change preserves Anthropic's native manager API while instrumenting the raw stream underneath it. It uses base Anthropic resources to avoid re-entering the tracked `create(stream=True)` override on older supported Anthropic versions, preventing double capture and preserving privacy options.

- `python -m pytest -q posthog/test/ai/anthropic/test_anthropic.py` (37 passed, 1 skipped)
- `ruff check posthog/ai/stream.py posthog/ai/anthropic/anthropic.py posthog/ai/anthropic/anthropic_async.py posthog/test/ai/anthropic/test_anthropic.py`
- `ruff format --check posthog/ai/stream.py posthog/ai/anthropic/anthropic.py posthog/ai/anthropic/anthropic_async.py posthog/test/ai/anthropic/test_anthropic.py`
- Focused mypy validation of the three runtime modules
- `git diff --check`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

A human directed a Pi worker to implement and prepare this compatibility fix. Pi autoreview identified an older-version double-tracking/privacy risk; the implementation was adjusted to construct base Anthropic resources, and the final autoreview rerun had no actionable findings. Human review is still required before merge.

### DIFF
--- a/posthog/ai/anthropic/anthropic.py
+++ b/posthog/ai/anthropic/anthropic.py
@@ -126,7 +126,9 @@ class WrappedMessages(Messages):
         # cannot re-enter our tracked ``create`` override.
         manager = Messages(self._client).stream(**kwargs)
         request_attribute = "_MessageStreamManager__api_request"
-        request = getattr(manager, request_attribute)
+        request = getattr(manager, request_attribute, None)
+        if request is None:
+            return manager
 
         def tracked_request():
             start_time = time.time()
--- a/posthog/ai/anthropic/anthropic.py
+++ b/posthog/ai/anthropic/anthropic.py
@@ -10,7 +10,7 @@ import time
 import uuid
 from typing import Any, Dict, List, Optional
 
-from posthog.ai.stream import StreamWrapper
+from ..stream import _StreamWrapper
 from posthog.ai.types import StreamingContentBlock, TokenUsage, ToolInProgress
 from posthog.ai.utils import (
     call_llm_and_track_usage,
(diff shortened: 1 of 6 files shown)


marandaneto:
@PostHog/team-ai-observability not sure if intentional so pls take a look at it

posthog/ai/anthropic/anthropic.py
```diff
@@ -115,37 +116,74 @@ def stream(
...
         if posthog_trace_id is None:
             posthog_trace_id = str(uuid.uuid4())
 
-        return self._create_streaming(
+        # Construct the provider resource directly so older Anthropic versions,
+        # whose stream manager delegates through ``self.create(stream=True)``,
+        # cannot re-enter our tracked ``create`` override.
+        manager = Messages(self._client).stream(**kwargs)
+        request_attribute = "_MessageStreamManager__api_request"
+        request = getattr(manager, request_attribute)
```
ioannisj: Should we guard the getattr() against a future anthropic rename? same for the async twin in anthropic_async.py

ioannisj (approved):
Looks reasonable

posthog/ai/anthropic/anthropic.py
marandaneto: Good catch! Fixed in a86e144 — both sync and async wrappers now return the native manager unchanged if Anthropic renames the private request attribute, with regression coverage for both paths.

Radu-Raicea (approved):

PR #808, opened 2026-07-30
fix(flags): prevent stale definition publication

marandaneto:
Overlapping feature flag definition refreshes can complete out of order, allowing an older response to overwrite newer definitions and their ETag. This change orders publication so stale responses, conditional responses, and reset errors cannot detach the published ETag from the matching in-memory state.

External cache writes are serialized separately from in-memory publication, preserving write order without holding the publication lock during provider I/O. Fork reinitialization replaces the new synchronization locks.

- `python -m pytest posthog/test/test_feature_flags.py posthog/test/test_client_fork.py posthog/test/test_flag_definition_cache.py --timeout=30 -q` (213 passed)
- `ruff format --check posthog/client.py posthog/test/test_feature_flags.py posthog/test/test_client_fork.py`
- `ruff check posthog/client.py posthog/test/test_feature_flags.py posthog/test/test_client_fork.py`
- `python -m mypy --no-site-packages --config-file mypy.ini posthog/client.py | python -m mypy_baseline filter`
- `python -W error -c 'import posthog'`
- `git diff --check`

## :pencil: Checklist

- [x] I updated the docs if needed. (No documentation changes were needed.)

### If releasing new changes

## 🤖 Agent context

The human DRI directed the scope, changeset text, commit message, and PR metadata. Pi coding agents implemented and validated the concurrency fix, with the autoreview tool used for iterative review. Review feedback led to keeping provider I/O outside the publication lock, preserving ordered cache writes, clearing invalid ETags after definition resets, and correctly fencing stale 304 responses.

The implementation intentionally stays within flag-definition fetch publication and fork-safe synchronization. A human review is required before merge.

### DIFF
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -2259,6 +2259,8 @@ class Client(object):
                         self._last_feature_flag_poll = datetime.now(tz=timezone.utc)
                         return
 
+                    self._flag_definition_published_generation = fetch_generation
+                    self._flag_definition_cache_generation = fetch_generation
                     self.log.debug(
                         "[FEATURE FLAGS] Flags not modified (304), using cached data"
                     )
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -627,6 +627,11 @@ class Client(object):
         self.flag_cache = self._initialize_flag_cache(flag_fallback_cache_url)
         self.flag_definition_version = 0
         self._flags_etag: Optional[str] = None
+        self._flag_definition_fetch_generation = 0
+        self._flag_definition_published_generation = 0
+        self._flag_definition_cache_generation = 0
+        self._flag_definition_publication_lock = threading.Lock()
+        self._flag_definition_cache_write_lock = threading.RLock()
         self._flag_definition_cache_provider = flag_definition_cache_provider
         self._flag_definition_cache_provider_async_runner: Optional[
             _BackgroundEventLoopRunner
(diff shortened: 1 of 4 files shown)


posthog/client.py
```diff
@@ -2216,91 +2226,137 @@ def _fetch_feature_flags_from_api(self):
...
-                )
-                self.log.error("[FEATURE FLAGS] %s", detail)
-                self.feature_flags = []
-                self.group_type_mapping = {}
-                self.cohorts = {}
+                    self.log.debug(
+                        "[FEATURE FLAGS] Flags not modified (304), using cached data"
+                    )
+                    self._last_feature_flag_poll = datetime.now(tz=timezone.utc)
+                    return
```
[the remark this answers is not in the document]
marandaneto: Good catch! Fixed in 5e6963e — accepted 304 responses now advance the publication and cache generations, with a regression test covering the older in-flight response.

posthog/client.py
```diff
@@ -2216,91 +2226,139 @@ def _fetch_feature_flags_from_api(self):
...
-                detail = (
-                    f"Error loading feature flags: {e.message}. "
-                    "Please verify both your project_api_key and secret_key. "
-                    "More information: https://posthog.com/docs/api/overview"
-                )
-                self.log.error("[FEATURE FLAGS] %s", detail)
-                self.feature_flags = []
-                self.group_type_mapping = {}
-                self.cohorts = {}
+                    self._flag_definition_published_generation = fetch_generation
```
ioannisj, line 2263: pre-PR code did `self._flags_etag = response.etag` ahead of the 304 check. Was that change intentional?

ioannisj (approved):
Small comment/question from me

posthog/client.py
marandaneto: It was not intentional. Restored the pre-PR behavior in 789041a: once a 304 is validated against the request ETag, its response ETag is published while the cached flag state remains unchanged. I also updated the 304 regression test to cover the ETag change.

gustavohstrassburger (approved):

PR #814, opened 2026-07-30
fix(capture): cap v0 retry delays

marandaneto:

Capture v0 treated a positive `Retry-After` response header as an unbounded replacement for its configured exponential backoff. A server-provided numeric value or far-future HTTP date could therefore block the consumer thread for an unexpectedly long time, while a small value could reduce the existing retry delay.

This change aligns v0 with the bounded retry behavior already used by capture v1: both the exponential delay and `Retry-After` use the shared 30-second ceiling, and the actual wait is their maximum.

- `/Users/marandaneto/Github/posthog-python/.venv/bin/python -m pytest posthog/test/test_consumer.py -q` (30 passed)
- `/Users/marandaneto/Github/posthog-python/.venv/bin/ruff check posthog/consumer.py posthog/test/test_consumer.py`
- `/Users/marandaneto/Github/posthog-python/.venv/bin/ruff format --check posthog/consumer.py posthog/test/test_consumer.py`
- `git diff --check`

Regression coverage verifies huge and small numeric and HTTP-date `Retry-After` values and confirms the configured number of attempts is preserved.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

A human directed the scope and PR requirements. Pi coding agent using GPT-5.6-sol (session `019fb2fd-ad74-70b3-a210-34f28b8ed762`) implemented the bounded delay and regression tests; Pi's autoreview tool independently reported no actionable findings. The implementation reuses capture v1's existing retry ceiling rather than introducing a second constant, and intentionally preserves v0's retry count and retryable-status behavior.

This PR requires human review and is not intended for autonomous merge.

### DIFF
--- a/posthog/consumer.py
+++ b/posthog/consumer.py
@@ -7,7 +7,7 @@ from threading import Thread
 from posthog._logging import _configure_posthog_logging
 from posthog.capture_compression import CaptureCompression
 from posthog.capture_mode import CaptureMode
-from posthog.capture_v1 import _send_v1_batch
+from posthog.capture_v1 import _MAX_BACKOFF_SECONDS, _send_v1_batch
 from posthog.request import (
     EVENTS_ENDPOINT,
     APIError,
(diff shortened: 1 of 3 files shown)


posthog/consumer.py
```diff
@@ -209,12 +209,14 @@ def is_retryable(exc):
                 if not is_retryable(e):
                     raise
                 if attempt < self.retries:
-                    # Respect Retry-After header if present, otherwise use exponential backoff
+                    configured_backoff = min(2**attempt, _MAX_BACKOFF_SECONDS)
```
ioannisj: `capture_v1._backoff` already implements this exact clamp-and-max and its docstring says the ceiling is deliberately shared between v0/v1 
https://github.com/PostHog/posthog-python/blob/2449a065df39a584b3c96102111b86bdf7c64d24/posthog/capture_v1.py#L436

Since `_MAX_BACKOFF_SECONDS` is already imported from there, worth replacing the inline block with `_backoff(attempt, retry_after)` so the formula lives in one place?

ioannisj (approved):

posthog/consumer.py
marandaneto: Good catch! Fixed in 8928bb7 — v0 now calls the shared `_backoff()` helper instead of duplicating its clamp-and-max logic.

PR #815, opened 2026-07-30
fix(mcp): isolate pending capture tasks

marandaneto:

MCP capture tasks were tracked globally, so flushing one analytics instance could wait for another instance's captures. Async drains could also encounter tasks created on another event loop.

Track each capture task by its lifecycle owner and event loop. Async analytics now drain only their own tasks on the current loop, while `PostHogMCP` consistently schedules and drains its captures on the shared background loop.

- `uv run --extra test pytest -q posthog/test/mcp` (145 passed)
- `uv run ruff check posthog/mcp/__init__.py posthog/mcp/_instrumentation.py posthog/mcp/posthog_mcp.py posthog/test/mcp/_helpers.py posthog/test/mcp/test_review_fixes.py posthog/test/mcp/test_pending_tasks.py`
- `uv run ruff format --check posthog/mcp/__init__.py posthog/mcp/_instrumentation.py posthog/mcp/posthog_mcp.py posthog/test/mcp/_helpers.py posthog/test/mcp/test_review_fixes.py posthog/test/mcp/test_pending_tasks.py`
- `git diff --check`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Pi implemented the human-directed fix from autoreview report 38. The selected approach scopes pending capture work by owner and event loop, and keeps the synchronous `PostHogMCP` lifecycle on the shared background loop so `flush()` and `shutdown()` can drain it safely. The generated patch changeset describes the user-visible fix.

This agent-assisted change requires human review before merge.


marandaneto:
@PostHog/team-mcp-analytics im not sure if this was intended or not, but i found out while reviewing some other stuff

gesh (approved):
Looks good to me! 
cc @lucasheriques as he might be more familiar with this implementation and whether it was intentional

posthog/mcp/__init__.py
```diff
@@ -125,10 +125,14 @@ async def capture(self, event: str, properties: Optional[dict] = None) -> None:
...
-        """Await in-flight auto-captured events scheduled on the current event loop.
+        """Await this server's in-flight auto-captures on the current event loop.
         Call this before ``posthog.shutdown()`` on exit so trailing tool-call events
         aren't dropped. (Then call ``posthog.flush()``/``shutdown()`` to send them.)"""
-        await drain_pending()
+        if self._key is None:
+            return
+        data = get_server_tracking_data(self._key)
+        if data is not None:
+            await drain_pending(data)
```
lucasheriques: if this method silently stops draining, a host that does tool call, `flush()`, exit loses every event. i checked: `['$mcp_initialize', '$mcp_tool_call']` becomes `[]`, because `asyncio.run()` cancels the pending capture tasks when main returns.

nothing exercises it. every `_flush()` in the suite is the internal `flush_background()` helper, which drains cross owner and wraps bg futures, so it's more forgiving than what ships. body replaced with `return`, drain pointed at the wrong owner, `self._key is None` guard deleted: suite stays green for all three.

`test_pending_tasks.py` drives the private `drain_pending()` with synthetic `object()` owners. that pins the mechanism, not the wiring, and the wiring is what changed. correctness depends on `get_server_tracking_data(self._key)` returning the same object identity the capture sites pass as `owner`. true today, unasserted, and `dataclasses.replace(data)` is enough to break it silently.

```python
async def test_analytics_flush_drains_its_own_captures():
    async def slow_before_send(event):
        await asyncio.sleep(0.05)
        return event

    server, client = make_server(), FakeClient()
    analytics = instrument(
        server, client, MCPAnalyticsOptions(before_send=slow_before_send)
    )

    await server._tool_manager.call_tool("add", {"a": 2, "b": 3, "context": "why"})
    assert _events(client, "$mcp_tool_call") == []  # still in flight
    await analytics.flush()
    assert len(_events(client, "$mcp_tool_call")) == 1
```

the slow `before_send` is load bearing. with default options the capture finishes during `call_tool` teardown, and the test passes even when `flush()` does nothing.

separately, the `self._key is None` guard two lines down isn't boilerplate. `_server_tracking` is a `WeakKeyDictionary` and `.get(None)` raises `TypeError: cannot create weak reference to 'NoneType'`, so `_NoopAnalytics` would raise from `flush()`. moving it to an override next to the existing `capture()` one says that out loud:

```python
class _NoopAnalytics(McpAnalytics):
    async def flush(self) -> None:
        # No tracking key, so there is nothing to drain -- and `_server_tracking` is a
        # WeakKeyDictionary, which raises TypeError when looked up with None.
        return None
```

posthog/mcp/_instrumentation.py
```diff
@@ -25,10 +25,13 @@
...
 
-# Keep strong refs to in-flight capture tasks/futures so they aren't GC'd mid-flight,
-# and so the asyncio ones can be awaited via drain_pending() before shutdown. Holds
-# asyncio.Task (running-loop path) or concurrent.futures.Future (sync background-loop path).
+# Keep strong refs to in-flight capture tasks/futures so they aren't GC'd mid-flight.
+# The metadata lets lifecycle drains select only work belonging to their analytics
+# handle/client and, for asyncio tasks, only work bound to the current event loop.
 _BACKGROUND_TASKS: Set[Any] = set()
+_TASK_OWNERS: Dict[Any, Any] = {}
+_TASK_LOOPS: Dict[Any, asyncio.AbstractEventLoop] = {}
```
lucasheriques: three structures, one key set, mutated together under one lock. both derived maps are unnecessary. `_BACKGROUND_TASKS` is `_TASK_OWNERS.keys()`. `_TASK_LOOPS` duplicates what the objects already know, and `_helpers.py:39` already uses `task.get_loop() is loop` instead of the map.

`_TASK_LOOPS.get(task) is _bg_loop` in `drain_pending_sync` is dead. every future there came from `run_coroutine_threadsafe(coro, _get_background_loop())`, so it can't be false. it also reads the mutable `_bg_loop` global outside `_bg_loop_lock`.

collapsed to one `task -> owner` dict, the filters become:

```python
# drain_pending
if task_owner is owner and isinstance(task, asyncio.Task) and task.get_loop() is loop and not task.done()

# drain_pending_sync
if task_owner is owner and isinstance(task, concurrent.futures.Future) and not task.done()
```

147/147 pass, ruff clean, +17/-21, two fewer globals, `Set` drops from the typing import, `_helpers.py` unchanged. can push the patch if useful.

lucasheriques (approved):
approving. mechanism is sound.

verified: `asyncio.Task` isn't a `concurrent.futures.Future` subclass, so the two drain filters are disjoint. locking is deadlock free, `_tasks_lock` is released before both `add_done_callback` and `concurrent.futures.wait`. owner and loop identity match end to end through `instrument()`. `background=True` for `PostHogMCP` fixes a real gap, covered by `test_posthog_mcp_sync_flush_drains_capture_from_async_host`.

two non blocking comments below, neither changes behaviour.

the one that matters: `flush()` is what gets trailing captures over the line before the loop closes. if it silently stops draining, a host that does tool call, `flush()`, exit loses every event, and nothing in the suite would catch that.

posthog/mcp/_instrumentation.py
marandaneto: Agreed, simplified in c719044 — _BACKGROUND_TASKS is now the single task-to-owner dictionary, asyncio loop identity comes from task.get_loop(), and the redundant sync background-loop check was removed.

posthog/mcp/__init__.py
marandaneto: Good catch! Fixed in c719044 — added an end-to-end slow-before_send regression test proving the returned analytics handle drains its own captures, and moved the no-key flush behavior into an explicit _NoopAnalytics.flush() override.

PR #816, opened 2026-07-30
fix(mcp): reset background capture after fork

marandaneto:

The MCP synchronous capture path can inherit background event-loop state across `fork()`, but the daemon thread that drives that loop does not survive in the child. An inherited locked mutex or pending parent work can then deadlock or prevent later child captures from completing.

Reset the MCP background capture loop, lock, and pending-work references in forked children without touching the parent's state or trying to close the defunct inherited loop.

- `/Users/marandaneto/Github/posthog-python/.venv/bin/python -m pytest posthog/test/mcp/test_instrumentation_fork.py -q`
- `/Users/marandaneto/Github/posthog-python/.venv/bin/ruff check posthog/mcp/_instrumentation.py posthog/test/mcp/test_instrumentation_fork.py`
- `/Users/marandaneto/Github/posthog-python/.venv/bin/ruff format --check posthog/mcp/_instrumentation.py posthog/test/mcp/test_instrumentation_fork.py`
- `/Users/marandaneto/Github/posthog-python/.venv/bin/python -W error -c "import posthog.mcp._instrumentation"`
- `git diff --check`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Pi implemented the targeted fix under human direction. Pi's autoreview report guided the final submission and found no actionable issues. The chosen approach uses `os.register_at_fork` to replace only child-process background capture state, including the potentially locked mutex, while preserving parent-process behavior. A real-fork regression verifies that inherited work is cleared and a fresh child loop completes capture.

### DIFF
--- a/.sampo/changesets/ardent-queen-vainamoinen.md
+++ b/.sampo/changesets/ardent-queen-vainamoinen.md
@@ -1,0 +1,5 @@
+---
+pypi/posthog: patch
+---
+
+Reset MCP background capture state after fork
--- a/posthog/mcp/_instrumentation.py
+++ b/posthog/mcp/_instrumentation.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 
 import asyncio
 import concurrent.futures
+import os
 import threading
 from datetime import datetime, timezone
 from typing import Any, Dict, List, Optional, Set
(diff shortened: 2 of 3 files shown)


gesh (approved):
Thanks! 🙌

PR #832, opened 2026-08-05
fix: prevent client lifecycle deadlocks

marandaneto:

Client lifecycle methods could deadlock when invoked reentrantly from consumer error callbacks, poller/cache-provider cleanup, executor hops, or concurrent threads. Forked clients could also inherit queue and synchronization locks held by threads that no longer exist.

This merges the drain signaling from #797 and serializes lifecycle ownership without holding state locks across user callbacks. It defers and coalesces callback lifecycle work, preserves queued delivery for normal `join()`/`shutdown()`, safely accounts for undrainable or interrupted queue work, hardens async cache-provider runner startup/close/executor behavior, and rebuilds fork-unsafe state before restarting child threads. The `on_error` and `shutdown()` documentation now explains direct deferred calls and the safe application-thread handoff pattern. Review follow-up also makes `join()`/`shutdown()` no-throw public APIs and removes redundant lifecycle status/error fields. Deferred callback requests are coalesced with a lock-protected dirty flag rather than a generation counter. Automated review follow-up adds per-attempt async-runner startup state, closes rejected/cancelled coroutines, daemonizes deferred helpers, and clarifies best-effort delivery semantics. Interpreter exit now shares one process-wide one-second best-effort flush budget across clients, then stops daemon consumers without joining in-flight transports. Read-only event loops supplied by custom policies are preserved rather than replaced with a platform-default loop.

- `uv run --extra test pytest -q --timeout=30` — 2008 passed, 15 skipped
- `uv run --extra test pytest -q --timeout=30 posthog/test/test_async_utils.py posthog/test/test_client.py posthog/test/test_consumer.py posthog/test/test_client_fork.py posthog/test/test_ai_capture_lane.py`
- Async runner startup/close stress loop — 50/50 passed
- `uv run --extra dev ruff format --check .`
- `uv run --extra dev ruff check .`
- `uv run --extra dev mypy --no-site-packages --config-file mypy.ini . | uv run --extra dev mypy-baseline filter`
- `uv run --extra dev python .github/scripts/check_public_api.py`
- Independent parallel reviews: clean

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with Pi, independent reviewer agents, and autoreview. The final review found no concrete lifecycle, queue, fork, or async-runner defects. Autoreview's remaining read-only-loop concern was rejected after verification: uvloop permits per-instance wrapping, configured extensible policy loops are preserved, and the fallback is intentionally limited to loop implementations that prohibit the wrapping required for safe custom-executor reentry.


marandaneto:
@dustinbyrne some eyes here would be good since its a complex code path, but i managed to reproduce the bug before and it seems fixed now + extensible tests run locally

posthog/client.py
```diff
@@ -2181,53 +2258,245 @@ def flush(self, timeout_seconds: Optional[float] = 10) -> None:
...
+                        self._lifecycle_in_progress = False
+                        self._lifecycle_owner = None
+                        self._lifecycle_condition.notify_all()
+                        return
+            except BaseException:
+                with self._lifecycle_condition:
+                    self._lifecycle_in_progress = False
+                    self._lifecycle_owner = None
+                    self._lifecycle_condition.notify_all()
+                raise
```
dustinbyrne, line 2577: this will leak via `join` / `shutdown`

posthog/test/test_client.py
```diff
@@ -2248,13 +2249,520 @@ def test_shutdown_clears_feature_flag_called_dedupe_cache(self):
...
+        exception_capture = mock.Mock()
+        client._metrics = metrics
+        client.exception_capture = exception_capture
+
+        client.join()
+        client.shutdown()
+
+        metrics.flush.assert_called_once()
+        metrics.reset.assert_called_once()
+        exception_capture.close.assert_called_once()
```
dustinbyrne, line 2701: expected to assert here?

posthog/client.py
```diff
@@ -2181,53 +2258,245 @@ def flush(self, timeout_seconds: Optional[float] = 10) -> None:
...
-            self.poller.stop()
-
-        # Shutdown the cache provider (release locks, cleanup)
-        self._shutdown_flag_definition_cache_provider()
-        self._unregister_duplicate_client()
+        if self._defer_lifecycle_from_callback("join"):
+            return
+        self._run_lifecycle()
 
     def shutdown(self) -> None:
```
dustinbyrne, line 2629: we could potentially decorate these with `@no_throw`

posthog/client.py
```diff
@@ -654,6 +689,26 @@ def __init__(
...
+        self._lifecycle_lock = threading.Lock()
+        self._lifecycle_condition = threading.Condition(self._lifecycle_lock)
+        self._lifecycle_in_progress = False
+        self._lifecycle_owner: Optional[threading.Thread] = None
+        self._workers_joined = False
+        self._join_cleanup_complete = False
+        self._shutdown_requested = False
+        self._shutdown_complete = False
+        self._deferred_lifecycle_error: Optional[BaseException] = None
+        self._deferred_lifecycle_failure = threading.Event()
```
dustinbyrne: i see these written to but not read anywhere

posthog/client.py
```diff
@@ -654,6 +689,26 @@ def __init__(
...
+        self._deferred_lifecycle_thread_pending = False
+        self._deferred_lifecycle_generation = 0
+        self._lifecycle_callback_context: ContextVar[bool] = ContextVar(
+            "posthog_lifecycle_callback", default=False
+        )
+        self._deferred_flush_lock = threading.Lock()
+        self._deferred_flush_pending = False
+        self._deferred_flush_followup = False
+        self._deferred_flush_followup_timeout: Optional[float] = None
+        self._shutdown_complete_event = threading.Event()
```
dustinbyrne: this is only used in tests

posthog/client.py
```diff
@@ -654,6 +689,26 @@ def __init__(
         self.debug = debug
         self.send = send
         self.sync_mode = sync_mode
+        self._lifecycle_lock = threading.Lock()
+        self._lifecycle_condition = threading.Condition(self._lifecycle_lock)
+        self._lifecycle_in_progress = False
+        self._lifecycle_owner: Optional[threading.Thread] = None
```
dustinbyrne: it looks like `_lifecycle_in_progress` is always set with `_lifecycle_owner` and indicates whether or not `_lifecycle_owner` is not None

we can probably drop `_lifecycle_in_progress`?

posthog/client.py
```diff
@@ -2181,53 +2258,245 @@ def flush(self, timeout_seconds: Optional[float] = 10) -> None:
...
+                with self._lifecycle_lock:
+                    generation = self._deferred_lifecycle_generation
+                    require_shutdown = self._shutdown_requested
+                try:
+                    self._run_lifecycle(require_shutdown=require_shutdown)
+                except BaseException as error:
+                    attempt += 1
+                    self.log.exception("Deferred %s attempt %d failed", name, attempt)
+                    with self._lifecycle_lock:
+                        if generation != self._deferred_lifecycle_generation:
```
dustinbyrne: i've noticed that gpt-5.6 seems to have a real inclination to introduce generation counters to resolve concurrent operations, and sometimes it seems like it's a lazy fix - the real issue lies elsewhere

not saying this is incorrect, but something to keep an eye out for

posthog/client.py
marandaneto: Agreed. Removed `_lifecycle_in_progress` in 6ea9973; `_lifecycle_owner is not None` is now the single source of truth.

posthog/client.py
marandaneto: Agreed. Removed the test-only shutdown completion event in 6ea9973 and updated the concurrency tests to assert the actual lifecycle state instead.

posthog/client.py
marandaneto: Done in 6ea9973 — both `join()` and `shutdown()` now use `@no_throw()`.

posthog/client.py
marandaneto: Good catch. Added `@no_throw()` to both public lifecycle entry points in 6ea9973, preserving debug-mode propagation while preventing normal `join()`/`shutdown()` calls from leaking cleanup exceptions.

posthog/client.py
marandaneto: Removed the unused deferred lifecycle error/event state and its reset/publication paths in 6ea9973.

posthog/client.py
marandaneto: Agreed that a generation counter was more machinery than this coalescing case needs. Replaced it in a19299e with a lock-protected boolean dirty flag: callback requests set it, the worker clears it before each pass, and atomically either loops when it was dirtied or clears the pending-worker flag. This preserves the completion-window guarantee without assigning identities to requests. Added deterministic success and failure-path escalation coverage and ran repeated race/deadlock stress tests.

posthog/_async_utils.py
```diff
@@ -49,26 +122,62 @@ def _ensure_loop(self) -> asyncio.AbstractEventLoop:
...
+                    name="PostHogBackgroundEventLoopRunner",
+                    daemon=True,
+                )
+                self._thread.start()
 
         self._started.wait()
         with self._lock:
+            if self._startup_error is not None:
+                raise self._startup_error
             assert self._loop is not None
```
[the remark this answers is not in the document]
marandaneto: Addressed in 04b3019 with per-startup state rather than shared generation state. Each waiter retains its own completion event, loop, and error even if a later startup begins. Added a deterministic test that pauses a failed generation’s waiter, successfully starts a second generation, then verifies the first waiter still receives its own startup error.

posthog/client.py
```diff
@@ -2181,53 +2258,245 @@ def flush(self, timeout_seconds: Optional[float] = 10) -> None:
...
+        return runner is not None and runner.owns_thread(current)
+
+    def _start_lifecycle_thread(self, target, name: str, *args) -> None:
+        threading.Thread(
+            target=target,
+            args=args,
+            name=f"posthog-{name}",
+            daemon=False,
+        ).start()
+
```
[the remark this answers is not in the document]
marandaneto: Addressed in 04b3019. Deferred lifecycle/flush helpers are now daemon threads, while lifecycle calls from application-owned threads remain blocking. Added coverage for daemon status and the operation-neutral lifecycle thread name.

posthog/client.py
```diff
@@ -2181,53 +2258,245 @@ def flush(self, timeout_seconds: Optional[float] = 10) -> None:
...
+                with self._lifecycle_condition:
+                    self._lifecycle_in_progress = False
+                    self._lifecycle_owner = None
+                    self._lifecycle_condition.notify_all()
+                raise
+
     def join(self) -> None:
         """
-        End the consumer thread once the queue is empty. Do not use directly, call `shutdown()` instead.
+        Flush queued events and end the consumer threads. Do not use directly, call `shutdown()` instead.
```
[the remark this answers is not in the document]
marandaneto: Addressed in 04b3019. The client and module-level lifecycle docs now describe best-effort processing attempts, explicitly state that failed or undrainable events may be dropped and reported through logging/`on_error`, and clarify that returning does not guarantee server receipt.

posthog/_async_utils.py
```diff
@@ -49,26 +122,62 @@ def _ensure_loop(self) -> asyncio.AbstractEventLoop:
...
             return self._loop
```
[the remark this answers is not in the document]
marandaneto: Addressed in 04b3019. Loop startup now has per-attempt state and no assertion. If close wins after startup, `run()` rejects the obsolete loop under the runner lock and retries on a fresh loop. Added a deterministic regression that pauses the waiter after startup, closes the runner, and verifies successful retry without `AssertionError`.

posthog/client.py
```diff
@@ -2181,53 +2258,245 @@ def flush(self, timeout_seconds: Optional[float] = 10) -> None:
...
+                    if generation != self._deferred_lifecycle_generation:
+                        attempt = 0
+                        continue
+                    self._deferred_lifecycle_thread_pending = False
+                    return
+
+        self._start_lifecycle_thread(run, name)
+        return True
+
+    def _defer_flush_from_callback(self, timeout_seconds: Optional[float]) -> bool:
```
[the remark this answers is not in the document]
marandaneto: Addressed in 04b3019. Coalesced work now runs on `posthog-lifecycle`, `_defer_lifecycle_from_callback` no longer accepts the first operation name, and failure logs use the operation selected for that pass (`join` or `shutdown`).

posthog/_async_utils.py
```diff
@@ -1,40 +1,113 @@
...
         self._started = threading.Event()
+        self._startup_error: BaseException | None = None
+        self._close_requested = False
         self._lock = threading.Lock()
 
     def run(self, awaitable: Awaitable[Any]) -> Any:
-        loop = self._ensure_loop()
-        future = asyncio.run_coroutine_threadsafe(self._await_result(awaitable), loop)
+        if threading.current_thread() is self._thread:
+            raise RuntimeError("cannot synchronously run from the runner thread")
```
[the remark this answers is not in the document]
marandaneto: Addressed in 04b3019. `run()` now closes coroutine objects on runner-thread rejection, loop-acquisition/scheduling failures, and cancellation after scheduling but before the wrapper takes its first step. Tests run under `-W error` cover startup rejection and the scheduled-before-first-step cancellation race.

posthog/test/test_client.py
marandaneto: Yes — `exception_capture.close.assert_called_once()` is the assertion here. `exception_capture` is a `unittest.mock.Mock`, so `close` is also a mock and `assert_called_once()` verifies that shutdown-only cleanup closes it exactly once after `join()` followed by `shutdown()`. No code change is needed.

marandaneto:
@dustinbyrne tried to simplify as much as i could, also ran a few cycles of reviews with different models

dustinbyrne (approved):
ran a bunch of targeted tests locally to validate, all look good:
- Full suite: 2045 passed, 16 skipped                                                                                                                                                                                                          
- Focused lifecycle suite on Python 3.11, 3.12, and 3.13: 305 passed + 14 subtests each                                                                                                                                                        
- 150 concurrent lifecycle rounds with producer threads, two lanes, upload failures, and callback re-entry                                                                                                                                     
- 200 shutdown/delivery races: every accepted event delivered exactly once                                                                                                                                                                     
- 250 async run()/close() races plus 100 startup/close races                                                                                                                                                                                   
- 100 forks while all relevant client/queue locks were held                                                                                                                                                                                    
- Real HTTP 503/on-error shutdown with 40 accepted batches                                                                                                                                                                                     
- Real in-flight synchronous request versus concurrent shutdown                                                                                                                                                                                
- Five-client atexit test completed in about 1.16s

PR #833, opened 2026-08-05
feat(openai-agents): per-run identity via RunConfig trace_metadata

marco-g-pm:
## Problem

The tracing processor registers once per process, so `distinct_id` / `properties` passed to `instrument()` are process-global. For any server handling multiple users, that stamps every user's traces with the same identity. The only per-run escape hatch was passing a callable `distinct_id` and threading values through trace metadata by hand — awkward, and the docs don't mention it.

Surfaced while testing `wizard ai-observability` against the wizard-workbench test apps: the wizard follows the docs, and on the openai-agents app produced events with no person attribution.

## Change

The processor now reads two well-known keys directly from `RunConfig(trace_metadata=...)`:

```python
instrument()  # process-constant config only

result = Runner.run_sync(
    agent,
    prompt,
    run_config=RunConfig(
        group_id=conversation_id,  # already maps to $ai_session_id
        trace_metadata={
            "posthog_distinct_id": user_id,
            "posthog_properties": {"plan": "scale"},
        },
    ),
)
```

- Key names match the per-call kwargs the OpenAI/Anthropic wrappers already use (`posthog_distinct_id`, `posthog_properties`) — one convention across paths, carried by whatever per-call surface each path has.
- Per-run values win over `instrument()` defaults.
- Applied to the `$ai_trace` event and every span/generation in the trace.
- Control keys are stripped from `$ai_trace_metadata`.
- No behavior change when the keys are absent; static/callable `distinct_id` still works as the default.

## Tests

7 new tests in `TestPerRunMetadata` (override precedence, span inheritance, property merge order, control-key stripping, fallback paths). Full openai_agents suite: 60 passed. Ruff clean.

Docs follow-up: posthog.com openai-agents page will be updated to show this pattern once this lands.

### DIFF
--- a/posthog/ai/openai_agents/__init__.py
+++ b/posthog/ai/openai_agents/__init__.py
@@ -34,11 +34,18 @@ def instrument(
 
     Args:
         client: Optional PostHog client instance. If not provided, uses the default client.
-        distinct_id: Optional distinct ID to associate with all traces.
-            Can also be a callable that takes a trace and returns a distinct ID.
+        distinct_id: Optional default distinct ID for all traces. Only suitable
+            as a static value when one process serves one user (a CLI or
+            worker); servers should pass identity per run via
+            ``RunConfig(trace_metadata={"posthog_distinct_id": ...})``, which
+            takes precedence. Can also be a callable that takes a trace and
+            returns a distinct ID.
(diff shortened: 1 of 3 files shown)


marco-g-pm:
claude got ahead of himself

PR #499, opened 2026-04-10
feat(llma): add $ai_stop_reason extraction for all providers

carlos-marchal-ph:
## Problem

LLM analytics users cannot filter or analyze why model responses ended. The finish/stop reason was never extracted from provider responses in the Python SDK.

## Changes

- New extractor functions for OpenAI (`choices[0].finish_reason` / `status`), Anthropic (`stop_reason`), and Gemini (`candidates[0].finish_reason`)
- Wired into both sync and async `call_llm_and_track_usage` and `capture_streaming_event`
- Streaming support for all three providers (captures from final chunk)
- LangChain, OpenAI Agents SDK, and Claude Agent SDK support
- Added `stop_reason` field to `StreamingEventData` TypedDict
- Tests for Gemini (non-streaming, max_tokens, streaming), OpenAI, and Anthropic
- Integration tests for all three providers (API-key gated)

PR #861, opened 2026-08-09
fix(ai): route async OpenAI streaming through capture_streaming_event

ckarnell:

The two async OpenAI streaming paths (`openai_async.py:242` and `:567`) build the `$ai_generation` property dict inline. Sync OpenAI, both Anthropic twins and both Gemini twins all delegate to `posthog.ai.utils.capture_streaming_event`. Async OpenAI streaming therefore drops `$ai_usage`, which #411 added "for backend cost calculations", along with `$ai_tokens_source`, `$ai_instructions` on the Responses path, and the AI Gateway double-billing warning.

The history suggests drift rather than a decision. #411 and #444 changed only `utils.py` and missed both async paths, while #499 edited `openai_async.py` directly and the property it added is on both twins today. `$ai_tokens_source` has been absent since February.

This routes both async paths through the shared helper instead of adding the missing keys to the inline dicts, since the duplication is what let them drift apart.

`posthog/test/ai/openai/test_async_parity.py` compares the properties captured by sync and async streaming and fails on anything the async side omits. On `main` it fails with `drops ['$ai_tokens_source', '$ai_usage']`, and it passes with this change. Anthropic runs as a control in the same test and its twins already agree, so the OpenAI difference is not an artifact of the harness.

`posthog/test/ai/` is 420 passed, 88 skipped, with `otel/` excluded for a dependency missing locally. `ruff format` is clean.

I have no live PostHog project and made no real OpenAI calls, so neither event has been watched landing in ingestion. `mypy` was not run.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

I direct a small studio and set this work going. The submission is mine. The finding came from a scanner we run across Python SDKs looking for sync and async twins that disagree, and Claude Code (Opus) wrote both the patch and the parity test.

Verification was deliberately not left to the same session that produced the patch. A second session re-ran the test against `main` and against the change, confirmed the failure text names the two dropped properties, and checked the ruff delta separately. The commit history behind the drift claim above was re-read at the commits, not carried over from the first summary.

One alternative was considered and dropped. Adding the four missing keys to the two inline dicts is a smaller diff, and it leaves in place the duplication that caused this.

The template asks for the DRI to be set as assignee. GitHub does not let an outside contributor assign anyone, so it is left blank and I am the DRI.


posthog/test/ai/openai/test_async_parity.py
```diff
@@ -0,0 +1,93 @@
...
+
+from unittest.mock import patch
+
+import pytest
+
+from posthog.ai.openai import OpenAI, AsyncOpenAI
+
+# fixtures mock_client and streaming_tool_call_chunks come from test_openai.py in this dir
+from posthog.test.ai.openai.test_openai import (  # noqa: F401
+    mock_client,
```
marandaneto: blocking: Fixture imports fail Ruff

Importing these fixtures directly and then reusing `mock_client` and `streaming_tool_call_chunks` as function or fixture parameters triggers Ruff `F811`, so `uv run ruff check .` fails. Please expose the shared fixtures through `conftest.py`, a pytest plugin, or another mechanism that avoids binding these names here rather than suppressing the lint.

marandaneto (commented):
Inline review finding.

marandaneto:
moving to draft until comments are resolved

ckarnell:
Done. You were right to push back on the suppression, since it would have left the same import doing the same thing with the warning turned off.

Both fixtures now live in `posthog/test/ai/openai/conftest.py`, so pytest supplies them by discovery and no test module imports or rebinds the names. The noqa comments are gone. `mock_client` and `streaming_tool_call_chunks` moved out of `test_openai.py` unchanged, along with the two chunk-type imports only they used.

`ruff check` passes on those files. `posthog/test/ai` is 420 passed and 88 skipped, unchanged.

The API snapshot was also out of date and is regenerated. Routing the async streaming path through `capture_streaming_event` dropped the `get_model_params` alias from `openai_async`, which is the one line that changed.

posthog/ai/openai/openai_async.py
```diff
@@ -18,7 +18,6 @@
     _capture_ai_event,
     extract_available_tool_calls,
     finalize_ai_content,
-    get_model_params,
```
marandaneto, line 21: blocking: Removing this import also removes the tracked public export `posthog.ai.openai.openai_async.get_model_params`, so existing imports now raise `ImportError`. Since this is a patch release and the PR does not declare a breaking change, please keep an explicit re-export rather than deleting the snapshot entry.

posthog/ai/openai/openai_async.py
```diff
@@ -254,71 +253,33 @@ async def _capture_streaming_event(
...
-                posthog_privacy_mode,
-                finalize_ai_content(
-                    format_openai_streaming_output(output, "responses"),
-                    self._client._ph_client,
-                ),
-            ),
-            "$ai_http_status": 200,
-            "$ai_input_tokens": usage_stats.get("input_tokens", 0),
-            "$ai_output_tokens": usage_stats.get("output_tokens", 0),
-            "$ai_cache_read_input_tokens": usage_stats.get(
```
marandaneto, line 286: blocking: The shared helper only emits `$ai_cache_read_input_tokens` and `$ai_reasoning_tokens` when positive, whereas this async path previously always emitted them with a default of `0`. Normal streams without cache/reasoning usage will therefore lose these properties, which can break queries based on property existence; please preserve the existing async event contract while adding the missing shared properties.

posthog/ai/openai/openai_async.py
```diff
@@ -254,71 +253,33 @@ async def _capture_streaming_event(
...
-        if hasattr(self._client._ph_client, "capture"):
-            _capture_ai_event(
-                self._client._ph_client,
-                "$ai_generation",
-                distinct_id=posthog_distinct_id or posthog_trace_id,
-                properties=event_properties,
-                groups=posthog_groups,
-            )
+        # Use the common capture function
+        capture_streaming_event(self._client._ph_client, event_data)
```
marandaneto, line 281: suggestion: Both changed helpers retain an unused `available_tool_calls` parameter, while their callers still invoke `extract_available_tool_calls`; this common function then extracts the same tools again from `kwargs`. Removing the dead parameter and caller-side extraction would avoid duplicate work and misleading plumbing.

marandaneto (commented):
Inline review findings.

posthog/ai/openai/openai_async.py
marandaneto: Fixed in b35d5b9 by restoring `get_model_params` as an explicit public re-export and restoring its API snapshot entry.

posthog/ai/openai/openai_async.py
marandaneto: Fixed in b35d5b9. Both async streaming paths now preserve the zero-valued cache-read and reasoning properties, with regression assertions covering them.

posthog/test/ai/openai/test_async_parity.py
marandaneto: Already addressed in ade11ac by moving the shared fixtures into `conftest.py`; Ruff passes without suppressions.

posthog/ai/openai/openai_async.py
marandaneto: Addressed in b35d5b9 by removing the unused parameters and caller-side extraction. The explicit import remains only to preserve the existing public API export.

marandaneto (approved):

marandaneto:
pushed some changes

ckarnell:
Thanks for cleaning it up, the re-export restore and the zero value assertions both look right to me.

Radu-Raicea (approved):

PR #866, opened 2026-08-11
fix: deliver events under gevent monkey-patching with an SDK-owned lane queue

### LINKED ISSUE #865: Silent total event loss under gunicorn gevent workers: flush crashes on gevent Queue (all_tasks_done), and 7.38.4 still emits only a boot-time burst with --preload
## Bug description

Recent `posthog-python` releases silently stop delivering **all** events when running inside gunicorn workers with `--worker-class gevent` (and `--preload`). `capture()` succeeds from the application's point of view, but the consumer/flush path is broken, so events buffer in memory and are never sent. For us this produced a multi-day, alert-free blackout of all server-side events from our web tier, while celery workers (no gevent) on the *same installed version* kept emitting normally.

## Environment

- Flask app on gunicorn: `gunicorn --workers 12 --worker-class gevent --preload wsgi:app` (gevent monkey-patching active in workers)
- `posthog` installed unpinned, so deploys pick up the latest release
- Python 3.x on Amazon Linux (Elastic Beanstalk)

## Phase 1 — loud failure (release current as of 2026-08-04)

A deploy on 2026-08-04 picked up the then-latest release. From that moment every flush attempt in the gevent workers crashed:

```
[ERROR] posthog: error flushing queue: 'gevent._gevent_cqueue.Queue' object has no attribute 'all_tasks_done'
```

gevent's monkey-patched `Queue` replacement does not implement `all_tasks_done` / `join()` semantics the SDK's flush path relies on. The consumer effectively dies; events are enqueued forever and 100% dropped in every gevent worker.

## Phase 2 — silent failure (7.38.4)

After upgrading to `7.38.4` (2026-08-10) the crash log disappeared, but emission from gunicorn workers still did not recover. The observable pattern: each worker emits a **brief burst of events right at process boot, then goes permanently silent**. Because gunicorn recycles workers (`--max-requests`), this shows up as a tiny trickle (~0.1% of normal volume) rather than zero — which is arguably worse, since it defeats "did events stop entirely?" alerting.

This smells like an interaction with `--preload` similar to #290: if consumer threads are now started eagerly (pre-fork, in the master), forked workers inherit a dead consumer thread and their queues are never drained. Under the older release (lazy consumer start on first capture in the worker) the same gunicorn config worked fine for months.

## Repro sketch

1. `gunicorn --worker-class gevent --preload app:app`
2. `posthog.capture(...)` inside request handlers
3. Observe: no events delivered (phase 1 versions crash in flush; 7.38.4 delivers only a boot-time burst per worker)

## Impact

- 6+ days of missing server-side analytics across two incident investigations; nothing user-visible failed, so nothing alerted
- The per-provider/per-endpoint gaps looked like real traffic drops, which sent us chasing phantom outages in downstream vendors

## Suggested fixes

- Feature-detect gevent monkey-patching / missing `all_tasks_done` instead of assuming stdlib `queue.Queue` behavior in the consumer/flush path
- Start (or restart) consumer threads post-fork, or document a `post_fork` `posthog.setup()` requirement loudly for `--preload` users (extends #290)
- Emit a loud warning when events are being enqueued while no live consumer thread exists — silent buffering-forever is the worst failure mode for an analytics SDK
- CI regression test: capture + flush under (a) gevent monkey-patching, (b) gunicorn preload-fork

Happy to provide log excerpts or exact version pins from our deploy history if useful.

ThaddKara:

Fixes #865.

gevent's `monkey.patch_all()` replaces `queue.Queue` with an implementation that lacks the private synchronization attributes CPython's pure-Python queue exposes: `mutex`, `not_empty`, `not_full`, `all_tasks_done`, `unfinished_tasks`, `_qsize()`, `_get()`. The consumer loop (`_DrainSignal`) and `Client.flush()` synchronize on exactly those attributes, so in a gunicorn `--worker-class gevent` worker:

- the consumer thread dies immediately with `AttributeError: 'gevent._gevent_cqueue.Queue' object has no attribute 'not_empty'`,
- `flush()` raises `error flushing queue: 'gevent._gevent_cqueue.Queue' object has no attribute 'all_tasks_done'`,
- and every captured event buffers forever and is silently dropped — `capture()` succeeds, nothing ships. In our production deployment this produced a multi-day, alert-free server-side telemetry blackout.

**The fix:** lanes now use `LaneQueue` (`posthog/_queue.py`), the pure-Python CPython queue implementation carried by the SDK itself. It builds only on `threading` primitives, which gevent patches compatibly, so behavior is identical on stock CPython and under monkey-patching — and its private surface can't be swapped out from under the SDK. `_DrainSignal`, `flush()`, and `rebuild_after_fork()` work unchanged; no call sites change beyond the two queue constructions.

- New `posthog/test/test_gevent_compat.py`:
  - `LaneQueue` interface + semantics tests (the private-attribute list is asserted as a contract).
  - Client capture→flush round-trip asserting the lane queue is SDK-owned.
  - A subprocess test that mirrors a real gevent worker (`monkey.patch_all()` first, import the SDK after), asserting events are delivered and the queue drains. On the unfixed code this test reproduces the exact production failure, including the `error flushing queue` log line and zero deliveries.
- `gevent` added to the `test` extra (CPython-only marker) so CI exercises the subprocess test; it skips cleanly where gevent is unavailable.
- Full `test_consumer.py` + `test_client.py` suites pass (250 tests), `ruff format --check` / `ruff check` clean on the locked ruff (0.12.2), `python -W error -c 'import posthog'` clean.

## :pencil: Checklist

### If releasing new changes

- [x] Ran `sampo add` to generate a changeset file (hand-written in the repo format: `.sampo/changesets/gevent-queue-compat.md`)

### DIFF
--- a/.sampo/changesets/gevent-queue-compat.md
+++ b/.sampo/changesets/gevent-queue-compat.md
@@ -1,0 +1,5 @@
+---
+pypi/posthog: patch
+---
+
+fix: deliver events under gevent monkey-patching by giving lanes an SDK-owned queue with CPython `queue.Queue` semantics; previously gevent's replacement `queue.Queue` lacked the private synchronization attributes the consumer and `flush()` rely on, so gevent gunicorn workers silently dropped every event.
--- a/.sampo/changesets/gevent-queue-compat.md
+++ b/.sampo/changesets/gevent-queue-compat.md
@@ -2,4 +2,4 @@
 pypi/posthog: patch
 ---
 
-fix: deliver events under gevent monkey-patching by giving lanes an SDK-owned queue with CPython `queue.Queue` semantics; previously gevent's replacement `queue.Queue` lacked the private synchronization attributes the consumer and `flush()` rely on, so gevent gunicorn workers silently dropped every event.
+fix: deliver events under gevent monkey-patching by giving lanes an SDK-owned queue (`LaneQueue`) with CPython `queue.Queue` semantics, including Python 3.13's `shutdown()`/`ShutDown` API; previously gevent's replacement `queue.Queue` lacked the private synchronization attributes the consumer and `flush()` rely on, so gevent gunicorn workers silently dropped every event. Note for integrators reaching into the backwards-compatible `Client.queue` property: the concrete type is now `posthog._queue.LaneQueue`, which matches `queue.Queue`'s full attribute surface but is deliberately not an instance of `queue.Queue` (inheriting would re-import the gevent bug).
--- a/LICENSE-PSF-2.0.txt
+++ b/LICENSE-PSF-2.0.txt
@@ -1,0 +1,54 @@
+This file applies to posthog/_queue.py, which is derived from CPython's
+Lib/queue.py. CPython is distributed under the Python Software Foundation
+License Version 2, reproduced below as required for derivative works.
+
+PYTHON SOFTWARE FOUNDATION LICENSE VERSION 2
+--------------------------------------------
+
+1. This LICENSE AGREEMENT is between the Python Software Foundation
+("PSF"), and the Individual or Organization ("Licensee") accessing and
+otherwise using this software ("Python") in source or binary form and
+its associated documentation.
+
+2. Subject to the terms and conditions of this License Agreement, PSF hereby
+grants Licensee a nonexclusive, royalty-free, world-wide license to reproduce,
+analyze, test, perform and/or display publicly, prepare derivative works,
+distribute, and otherwise use Python alone or in any derivative version,
+provided, however, that PSF's License Agreement and PSF's notice of copyright,
+i.e., "Copyright (c) 2001, 2002, 2003, 2004, 2005, 2006, 2007, 2008, 2009, 2010,
+2011, 2012, 2013, 2014, 2015, 2016, 2017, 2018, 2019, 2020, 2021, 2022, 2023 Python Software Foundation;
+All Rights Reserved" are retained in Python alone or in any derivative version
+prepared by Licensee.
+
+3. In the event Licensee prepares a derivative work that is based on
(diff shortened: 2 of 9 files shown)


ThaddKara:
Follow-up commit after an internal review pass (68c3b5f):

- **public_api_snapshot regenerated** for the `Client.queue` annotation change — `make public_api_check` passes.
- **Python 3.13 `shutdown()`/`ShutDown`** implemented on `LaneQueue` with CPython semantics, so the backwards-compat `Client.queue` surface keeps 3.13 parity. `isinstance(client.queue, queue.Queue)` remains deliberately `False` — inheriting would re-import the bug under monkey-patching (base class would resolve to gevent's Queue), and `queue.Queue` isn't an ABC; documented in the module docstring and flagged in the changeset.
- **PSF-2.0 license text retained** (`LICENSE-PSF-2.0.txt`) for the CPython-derived implementation; ships via the default `LICEN[CS]E*` glob. Happy to fold it into whatever third-party-license convention you prefer.
- **Consumer/_DrainSignal suite now runs against `LaneQueue`** (one-line alias in test_consumer.py) so the production queue class is what the delivery tests exercise.
- **gevent test floor raised to 25.4.1** — the first version whose `monkey.patch_all()` replaces `queue.Queue` — and the subprocess test now asserts that premise, so the regression test can't pass vacuously.
- Test hygiene: clients are `join()`ed so no consumer thread outlives the tests.
- `uv.lock` redone with current uv: purely additive (+123/-0, gevent + deps only), exclude-newer sentinel preserved.
- Relative import per AGENTS.md.

marandaneto:
superseeded by https://github.com/PostHog/posthog-python/pull/867
thanks @ThaddKara

PR #868, opened 2026-08-12
test: Isolate local feature flag polling assertions

marandaneto:

Several local feature flag evaluation tests patched the module-level `posthog.client.get` function and asserted that it was never called. A background poller from another client could call the same mock while the test was running. This caused an intermittent CI failure even though the client under test evaluated its flag locally.

The affected tests now patch `_load_feature_flags` on their own client instance. The assertions still verify that local evaluation does not load flag definitions, but requests from unrelated clients can no longer change the call count.

- Ran the four affected tests 30 times: all passed.
- Ran `posthog/test/test_feature_flags.py`: 173 passed with 13 subtests.
- Ran Ruff formatting and lint checks.
- Ran mypy for the changed test file.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with Pi under human direction after a flaky Python 3.13 CI failure on #867. Pi autoreview reported no actionable findings for commit `428fd8ca52ee0ccf48b361dfb3d85a823e0a7db4`.

### DIFF
--- a/posthog/test/test_feature_flags.py
+++ b/posthog/test/test_feature_flags.py
@@ -1439,9 +1439,11 @@ class TestLocalEvaluation(unittest.TestCase):
         self.assertEqual(patch_capture.call_count, 0)
 
     @mock.patch("posthog.client.flags")
-    @mock.patch("posthog.client.get")
-    def test_feature_flags_local_evaluation_None_values(self, patch_get, patch_flags):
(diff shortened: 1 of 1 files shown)


arnohillen (approved):
[arno's agent] Reviewed at head 428fd8ca: change verified against head code, CI classified, no blocking findings.

PR #870, opened 2026-08-13
chore(ci): automate PostHog Python SDK upgrades

marandaneto:

PostHog uses the Python SDK under two distribution names. The application, LLM gateway, and ingestion tests use `posthoganalytics` to avoid colliding with PostHog's own Python package. The infra CLI installs the same SDK as `posthog`.

New Python SDK releases currently require someone to update these declarations and both uv lockfiles manually. This adds a workflow that keeps every declaration on the released SDK version and opens a signed pull request in `PostHog/posthog`.

The release workflow dispatches the upgrade after publishing `posthoganalytics`.

- Ran `actionlint` against both changed workflows.
- Parsed both changed workflows as YAML.
- Ran the upgrade commands in a clean PostHog worktree with version 7.39.0.
- Confirmed the commands updated the root and LLM gateway manifests and lockfiles, the ingestion test requirement, and the infra CLI requirement.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Pi implemented the workflow from the existing `posthog-rs` upgrade flow. The workflow treats `posthoganalytics` and `posthog` as distribution names for the same SDK release. It updates all six dependency and lock files in the PostHog repository and fails if any expected file remains unchanged.

### DIFF
--- a/.github/workflows/posthog-upgrade.yml
+++ b/.github/workflows/posthog-upgrade.yml
@@ -1,0 +1,142 @@
+name: "PostHog Upgrade"
+
+on:
+  workflow_dispatch:
+    inputs:
+      package_name:
+        type: choice
+        description: "Package name to upgrade"
+        required: true
+        options:
+          - posthoganalytics
+      package_version:
+        description: "Package version to upgrade to"
+        required: true
+        type: string
+
+permissions:
+  contents: read
+
+jobs:
+  posthog-upgrade:
+    name: Upgrade PostHog package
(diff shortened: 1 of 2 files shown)


marandaneto:
this is similar to posthog-js and posthog-rs already so nothing new

arnohillen (approved):
[arno's agent] Reviewed at head db8e0558: change verified against head code, CI classified, no blocking findings.

PR #877, opened 2026-08-18
test: harden snapshot User-Agent normalization

dustinbyrne:

Release-dependent SDK versions can make otherwise-correct server payload snapshots fail after a release. This hardens snapshot normalization so expected `posthog-python/<version>` User-Agent values are stable while malformed or unrelated values remain snapshot-visible.

This prevents false-negative snapshots without release-version churn and does not change runtime behavior.

- `uv run pytest --verbose --timeout=30 posthog/test/test_server_payload_snapshots.py` (10 passed)
- `uv run ruff format --check posthog/test/test_server_payload_snapshots.py`
- `uv run ruff check posthog/test/test_server_payload_snapshots.py`
- `git diff --check`

## :pencil: Checklist

### If releasing new changes

This is test-only, so no Sampo changeset is required.

## 🤖 Agent context

Implemented with the Pi coding agent in a human-directed session; no shareable session URL is available. The final review identified that the initial User-Agent matcher accepted non-token HTTP product versions, so it was tightened to the RFC token character set and covered with malformed-version cases before submission.

### DIFF
--- a/posthog/test/test_server_payload_snapshots.py
+++ b/posthog/test/test_server_payload_snapshots.py
@@ -1,16 +1,18 @@
 import json
+import re
 from pathlib import Path
 from unittest import mock
 
+import pytest
 import requests
 from freezegun import freeze_time
 
 from posthog.client import Client
 
-
 _SNAPSHOT_DIRECTORY = Path(__file__).with_name("snapshots")
 _TEST_FILE_SUFFIX = "posthog/test/test_server_payload_snapshots.py"
 _FIXED_TIME = "2026-01-02T03:04:05+00:00"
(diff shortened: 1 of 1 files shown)


arnohillen (approved):
<small><em>(created by PostHog Desktop)</em></small>

The character class is exactly the RFC 7230 tchar set, `fullmatch` anchors both ends so no partial or multiline over-match is possible, and a non-matching value falls through to the raw string rather than being dropped. Ran the regex against all six negative cases and confirmed none match, and the live header (`"posthog-python/" + VERSION` at `request.py:189`) matches the gated form so the fixtures stay release-bump-proof.

PR #878, opened 2026-08-19
fix(flags): fall back for missing local definitions

marandaneto:

When `evaluate_flags()` receives a non-empty `flag_keys` list, it first checks locally loaded flag definitions. A requested key that was missing from those definitions was silently omitted, even when remote evaluation was allowed. This can happen for a newly created flag before the next definitions poll.

This follows the cross-SDK behavior agreed in [PostHog/posthog-android#678](https://github.com/PostHog/posthog-android/pull/678), including the [SDK comparison](https://github.com/PostHog/posthog-android/pull/678#issuecomment-5341493624), [spec analysis](https://github.com/PostHog/posthog-android/pull/678#issuecomment-5341513993), and [decision to fall back for missing keys](https://github.com/PostHog/posthog-android/pull/678#issuecomment-5341879578).

The local evaluation path now makes one `/flags` fallback per `evaluate_flags()` call when any requested key is absent. It sends the caller's original key list, keeps locally evaluated values when the remote response also contains them, and filters unexpected response keys out of the returned snapshot. `only_evaluate_locally=True` still makes no remote request and leaves missing keys absent.

Python does not cache complete `evaluate_flags()` snapshots across calls. A key missing from both local definitions and `/flags` therefore costs one fallback on each call. An empty `flag_keys` list keeps the existing behavior and is treated like `None`, which evaluates all flags.

The changeset uses a minor bump because this changes observable request behavior. Previous versions omitted a missing local key without making a request.

- `uv run pytest --timeout=30` - 2107 passed, 15 skipped
- `uv run pytest posthog/test/test_evaluate_flags.py --timeout=30` - 43 passed
- `uv run ruff check .`

The regression tests cover scoped fallback, local values winning over conflicting remote values, filtering unrequested remote flags, strict local-only behavior, and a server-missing key making exactly one fallback per evaluation call.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented and reviewed by Pi worker agents under the direction of @marandaneto. The behavior follows the cross-SDK decision linked above while preserving Python's existing no-snapshot-cache and empty-list conventions.

### DIFF
--- a/.sampo/changesets/gallant-princess-kalma.md
+++ b/.sampo/changesets/gallant-princess-kalma.md
@@ -1,0 +1,5 @@
+---
+pypi/posthog: patch
+---
+
+Fall back to remote evaluation when a requested flag is missing from local definitions
--- a/.sampo/changesets/gallant-princess-kalma.md
+++ b/.sampo/changesets/gallant-princess-kalma.md
@@ -1,5 +1,5 @@
 ---
-pypi/posthog: patch
+pypi/posthog: minor
 ---
 
-Fall back to remote evaluation when a requested flag is missing from local definitions
+Fall back to remote evaluation when a requested flag is missing from local definitions. This changes the previous behavior where the key was omitted without a request.
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -4044,8 +4044,11 @@ class Client(object):
             only_evaluate_locally: If True, never fall back to remote evaluation —
                 flags that can't be evaluated locally are simply omitted from the snapshot.
             disable_geoip: Whether to disable GeoIP lookup.
-            flag_keys: Optional list of flag keys to scope the underlying ``/flags``
-                request to a subset.
+            flag_keys: Optional list that scopes local evaluation, the underlying ``/flags``
+                request, and the returned snapshot. A requested key absent from loaded local
+                definitions is included in one remote fallback per ``evaluate_flags`` call unless
+                ``only_evaluate_locally`` is True. If the server also does not know the key, it is
+                omitted from the snapshot.
             device_id: Optional device ID override. If not provided, falls back to the
                 context device_id (which may be set via tracing headers). Used by
                 experience-continuity flags to match users across distinct_id changes.
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -4092,6 +4092,7 @@ class Client(object):
             )
         )
         groups = groups or {}
+        requested_keys = set(flag_keys) if flag_keys else None
 
         records: Dict[str, _EvaluatedFlagRecord] = {}
         request_id: Optional[str] = None
@@ -4121,6 +4122,11 @@ class Client(object):
         feature_flags_by_key: Dict[str, Any] = self.feature_flags_by_key or {}
         local_flags = local_result.get("featureFlags") or {}
         local_payloads = local_result.get("featureFlagPayloads") or {}
+        if requested_keys and not requested_keys.issubset(local_flags):
+            # A requested flag may have been created since the last definitions poll.
+            # Ask /flags for the caller's original scope unless this is a local-only call.
+            fallback_to_server = True
+
         for key, value in local_flags.items():
             flag_def = feature_flags_by_key.get(key) or {}
             records[key] = _EvaluatedFlagRecord(
@@ -4164,6 +4170,8 @@ class Client(object):
                     response.get("minimalFlagCalledEvents") is True
                 )
                 for key, detail in response.get("flags", {}).items():
+                    if requested_keys is not None and key not in requested_keys:
+                        continue
                     if key in locally_evaluated_keys:
                         continue
                     payload = _parse_flag_payload(
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -4044,8 +4044,9 @@ class Client(object):
             only_evaluate_locally: If True, never fall back to remote evaluation —
                 flags that can't be evaluated locally are simply omitted from the snapshot.
             disable_geoip: Whether to disable GeoIP lookup.
-            flag_keys: Optional list that scopes local evaluation, the underlying ``/flags``
-                request, and the returned snapshot. A requested key absent from loaded local
+            flag_keys: Optional non-empty list that scopes local evaluation, the underlying
+                ``/flags`` request, and the returned snapshot. An empty list is treated like
+                ``None`` and evaluates all flags. A requested key absent from loaded local
                 definitions is included in one remote fallback per ``evaluate_flags`` call unless
                 ``only_evaluate_locally`` is True. If the server also does not know the key, it is
                 omitted from the snapshot.
@@ -4095,6 +4096,7 @@ class Client(object):
             )
         )
         groups = groups or {}
+        # Keep the existing API convention that an empty list means no scope.
         requested_keys = set(flag_keys) if flag_keys else None
 
         records: Dict[str, _EvaluatedFlagRecord] = {}
--- a/posthog/test/test_evaluate_flags.py
+++ b/posthog/test/test_evaluate_flags.py
@@ -395,6 +395,65 @@ class TestEvaluateFlagsLocalPayloads(unittest.TestCase):
         self.assertEqual(properties["$feature_flag_payload"], {"copy": "new"})
 
 
+class TestEvaluateFlagsMissingLocalDefinition(unittest.TestCase):
+    def setUp(self):
+        self.client = Client(FAKE_TEST_API_KEY, secret_key="test")
+        self.client.feature_flags = [
+            {
+                "id": 1,
+                "name": "Local flag",
+                "key": "local-flag",
+                "active": True,
+                "filters": {"groups": [{"properties": [], "rollout_percentage": 100}]},
+            }
+        ]
+
+    def tearDown(self):
+        self.client.shutdown()
+
+    @staticmethod
+    def _remote_response():
+        return {
+            "flags": {
+                "local-flag": {"enabled": False, "variant": None},
+                "remote-only": {"enabled": True, "variant": None},
+                "unrequested": {"enabled": True, "variant": None},
+            }
+        }
+
+    @mock.patch("posthog.client.flags")
+    def test_missing_requested_key_triggers_one_scoped_fallback(self, patch_flags):
+        patch_flags.return_value = self._remote_response()
+        requested_keys = ["local-flag", "remote-only"]
+
+        flags = self.client.evaluate_flags("user-1", flag_keys=requested_keys)
+
+        self.assertEqual(set(flags.keys), set(requested_keys))
+        self.assertTrue(flags.get_flag("local-flag"))
+        self.assertTrue(flags.get_flag("remote-only"))
+        patch_flags.assert_called_once()
+        self.assertEqual(
+            patch_flags.call_args.kwargs["flag_keys_to_evaluate"], requested_keys
+        )
+
+    @mock.patch("posthog.client.flags")
+    def test_missing_requested_key_is_omitted_for_local_only_evaluation(
+        self, patch_flags
+    ):
+        patch_flags.return_value = self._remote_response()
+
+        flags = self.client.evaluate_flags(
+            "user-1",
+            flag_keys=["local-flag", "remote-only"],
+            only_evaluate_locally=True,
+        )
+
+        self.assertEqual(flags.keys, ["local-flag"])
+        self.assertTrue(flags.get_flag("local-flag"))
+        self.assertIsNone(flags.get_flag("remote-only"))
+        patch_flags.assert_not_called()
+
+
 class TestEvaluateFlagsLocalDeviceBucketing(unittest.TestCase):
     def setUp(self):
         self.client = Client(FAKE_TEST_API_KEY)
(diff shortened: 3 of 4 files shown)


dustinbyrne (approved):

posthog/test/test_evaluate_flags.py
```diff
@@ -395,6 +395,81 @@ def test_flag_called_event_carries_parsed_local_payload(
...
+        first = self.client.evaluate_flags("user-1", flag_keys=requested_keys)
+        second = self.client.evaluate_flags("user-1", flag_keys=requested_keys)
+
+        for flags in (first, second):
+            self.assertEqual(flags.keys, ["local-flag"])
+            self.assertTrue(flags.get_flag("local-flag"))
+            self.assertIsNone(flags.get_flag("typo-flag"))
+        self.assertEqual(patch_flags.call_count, 2)
+        for call in patch_flags.call_args_list:
+            self.assertEqual(call.kwargs["flag_keys_to_evaluate"], requested_keys)
```
haacked: `blocking`: Nothing in the test suite proves the fallback stays *off* when every requested key resolves locally. That's the whole point of the new check at [`posthog/client.py:4130`](https://github.com/PostHog/posthog-python/blob/9691157f91c9128e97e9949a67fd79eab1e259f3/posthog/client.py#L4130) — it keeps `evaluate_flags(flag_keys=[…])` off the network when local definitions already cover the request.

I verified this through mutation testing: replacing [line 4130](https://github.com/PostHog/posthog-python/blob/9691157f91c9128e97e9949a67fd79eab1e259f3/posthog/client.py#L4130) with `if requested_keys:` leaves 563 tests green across ten flag-related test files. The opposite mutation, `if False:`, fails two of your new tests, so the "fallback fires" direction is pinned and only "fallback stays off" is naked. There's also no safety net: the `Targeted mutation tests` job only runs against `posthog/utils.py`, `posthog/test/test_utils.py`, and `posthog/test/test_size_limited_dict.py`, so it never checks `client.py`.

A regression here wouldn't corrupt flag values (the merge loop still skips `locally_evaluated_keys`), but it would silently add a round trip per request on the hot path. That's the kind of thing that surfaces as a latency degradation months later. This class already loads `local-flag` in setUp, so a test takes five lines:

```python
    @mock.patch("posthog.client.flags")
    def test_no_fallback_when_all_requested_keys_resolve_locally(self, patch_flags):
        flags = self.client.evaluate_flags("user-1", flag_keys=["local-flag"])

        self.assertEqual(flags.keys, ["local-flag"])
        patch_flags.assert_not_called()
```

posthog/client.py
```diff
@@ -4121,6 +4127,11 @@ def evaluate_flags(
         feature_flags_by_key: Dict[str, Any] = self.feature_flags_by_key or {}
         local_flags = local_result.get("featureFlags") or {}
         local_payloads = local_result.get("featureFlagPayloads") or {}
+        if requested_keys and not requested_keys.issubset(local_flags):
```
haacked: `suggestion`: A stale key in a caller's `flag_keys` list now costs a blocking network round trip on every `evaluate_flags` call, with nothing to bound it. This condition is true forever for a deleted flag or a typo — the definitions poll will never bring the key back, and `/flags` will not return it either. Your `test_server_missing_key_falls_back_once_per_evaluation_call` documents exactly that.

The cost scales with request volume, because `evaluate_flags` runs once per incoming request. A service handling 500 requests per second with one dead key in an otherwise fully-local key list goes from 0 to 500 `/flags` requests per second, each a synchronous POST on the calling thread. In the degraded case, with the defaults at [`posthog/client.py:674-681`](https://github.com/PostHog/posthog-python/blob/9691157f91c9128e97e9949a67fd79eab1e259f3/posthog/client.py#L674-L681) (`timeout=3`, `max_retries=1`), a `/flags` endpoint that times out or returns 502 costs 3 seconds, a 0.3 second backoff, then another 3 seconds. That's roughly 6.3 seconds of a blocked worker per request. Before this change, a caller whose keys all resolved locally never touched the network at all.

I'm not calling this a defect. The spec makes negative-knowledge caching `MAY` rather than `SHALL`, and your PR body names the tradeoff, so shipping without it is compliant and deliberate. The question is only whether unbounded is the right first version, since the optional cache collapses the deleted-key case to about one probe per definitions refresh, roughly 0.03 requests per second at the default `poll_interval=30` instead of 500. If you do add it later, `_load_feature_flags` and `on_flag_definitions_received` in [`posthog/flag_definition_cache.py`](https://github.com/PostHog/posthog-python/blob/9691157f91c9128e97e9949a67fd79eab1e259f3/posthog/flag_definition_cache.py) are the clearing points, and the spec's two `SHALL NOT` cases are the traps: never record negative knowledge from a failed request, a quota-limited response, or one with `errorsWhileComputingFlags` set.

posthog/client.py
```diff
@@ -4092,6 +4096,8 @@ def evaluate_flags(
             )
         )
         groups = groups or {}
+        # Keep the existing API convention that an empty list means no scope.
+        requested_keys = set(flag_keys) if flag_keys else None
```
haacked: `suggestion`: Both docstrings now promise that `flag_keys=[]` behaves like `None` and evaluates all flags, and this line is the only thing making that true. Nothing tests it.

The reason it's worth pinning is that the two new call sites disagree about emptiness: [line 4130](https://github.com/PostHog/posthog-python/blob/9691157f91c9128e97e9949a67fd79eab1e259f3/posthog/client.py#L4130) uses truthiness, [line 4178](https://github.com/PostHog/posthog-python/blob/9691157f91c9128e97e9949a67fd79eab1e259f3/posthog/client.py#L4178) uses `is not None`. Today that's safe, because this line can only produce `None` or a non-empty set. If a future refactor let an empty set through, line 4130 would skip the fallback while line 4178 would filter out every remote flag. Callers passing `[]` would get an empty snapshot with no error anywhere.

Two cheap things close it. First, make the guards read the same way, so nobody has to re-derive that invariant:

```python
if requested_keys is not None and not requested_keys.issubset(local_flags.keys()):
```

Second, add a test — `TestEvaluateFlagsRemote` loads no local definitions, so a test there covers both new lines:

```python
    @mock.patch("posthog.client.flags")
    def test_empty_flag_keys_evaluates_all_flags(self, patch_flags):
        patch_flags.return_value = _flags_response_fixture()

        flags = self.client.evaluate_flags("user-1", flag_keys=[])

        self.assertEqual(
            sorted(flags.keys), ["boolean-flag", "disabled-flag", "variant-flag"]
        )
        self.assertNotIn("flag_keys_to_evaluate", patch_flags.call_args.kwargs)
```

posthog/client.py
```diff
@@ -4044,8 +4044,12 @@ def evaluate_flags(
             only_evaluate_locally: If True, never fall back to remote evaluation —
                 flags that can't be evaluated locally are simply omitted from the snapshot.
             disable_geoip: Whether to disable GeoIP lookup.
-            flag_keys: Optional list of flag keys to scope the underlying ``/flags``
-                request to a subset.
+            flag_keys: Optional non-empty list that scopes local evaluation, the underlying
```
haacked: `suggestion`: This entry opens by calling `flag_keys` a "non-empty list" and then two sentences later says an empty list is treated like `None`. A reader who stops at the first sentence will conclude that `[]` is unsupported.

Two other things got lost in the rewrite. "One remote fallback per `evaluate_flags` call" reads like a reassuring bound, when the actual cost is a `/flags` request on *every* call for a key the server never knows about. And the purpose guidance is gone — the four sibling `flag_keys_to_evaluate` entries in this file at lines [1284](https://github.com/PostHog/posthog-python/blob/9691157f91c9128e97e9949a67fd79eab1e259f3/posthog/client.py#L1284), [1321](https://github.com/PostHog/posthog-python/blob/9691157f91c9128e97e9949a67fd79eab1e259f3/posthog/client.py#L1321), [3908](https://github.com/PostHog/posthog-python/blob/9691157f91c9128e97e9949a67fd79eab1e259f3/posthog/client.py#L3908), and [3955](https://github.com/PostHog/posthog-python/blob/9691157f91c9128e97e9949a67fd79eab1e259f3/posthog/client.py#L3955) all tell the reader why they would reach for the parameter, and the old text did too. What replaced it is entirely edge-case semantics.

Something like this, applied to both this entry and its mirror at [`posthog/__init__.py:1080`](https://github.com/PostHog/posthog-python/blob/9691157f91c9128e97e9949a67fd79eab1e259f3/posthog/__init__.py#L1080) so the two stay in sync:

```
            flag_keys: Optional list of flag keys that scopes local evaluation, the
                underlying ``/flags`` request, and the returned snapshot. Use it when you
                need a handful of flags out of many — the smaller request is cheaper.
                ``None`` or an empty list evaluates all flags. When a requested key has no
                local definition, every call makes one ``/flags`` request for the full
                requested scope, unless ``only_evaluate_locally`` is True. Keys the server
                does not know are absent from the snapshot.
```

.sampo/changesets/gallant-princess-kalma.md
```diff
@@ -0,0 +1,5 @@
+---
+pypi/posthog: minor
+---
+
+Fall back to remote evaluation when a requested flag is missing from local definitions. This changes the previous behavior where the key was omitted without a request.
```
haacked: `suggestion`: The changeset doesn't warn about the case that will actually surprise people on upgrade, and it omits the second observable change entirely.

A requested key that's missing from local definitions *and* unknown to the server, a typo or a deleted flag, now costs one blocking `/flags` request on every `evaluate_flags()` call, where it previously cost zero. The PR body says this and `test_server_missing_key_falls_back_once_per_evaluation_call` pins it, but the changeset is what users actually read on upgrade, and those requests are quota-metered. Separately, the filter at [`posthog/client.py:4178`](https://github.com/PostHog/posthog-python/blob/9691157f91c9128e97e9949a67fd79eab1e259f3/posthog/client.py#L4178) now scopes the returned snapshot to the requested keys, not just the request. That's a real behavior change and the changeset doesn't mention it at all.

Something covering both:

```
`evaluate_flags()` now falls back to a remote `/flags` request when a requested `flag_keys` entry has no local flag definition, instead of silently omitting the key. This makes a newly created flag resolve before the next definitions poll picks it up. A key missing from both local definitions and the server, a deleted flag or a typo, costs one `/flags` request per `evaluate_flags()` call, since the SDK does not remember which keys the server rejected. Scope `flag_keys` to keys you expect to exist, or set `only_evaluate_locally=True` to keep the previous no-request behavior. The returned snapshot is now also scoped to `flag_keys`, so flags outside that list no longer appear even if the server returns them.
```

posthog/test/test_evaluate_flags.py
```diff
@@ -395,6 +395,81 @@ def test_flag_called_event_carries_parsed_local_payload(
         self.assertEqual(properties["$feature_flag_payload"], {"copy": "new"})
 
 
+class TestEvaluateFlagsMissingLocalDefinition(unittest.TestCase):
```
haacked: `suggestion`: Worth adding one more case to this class: what a scoped call returns when the newly triggered fallback fails.

A call where one requested key is missing locally now reaches the network on a path that used to stay local, so a `/flags` outage or a quota limit now degrades evaluations that were previously pure local evaluation. The good news is that the locally resolved records are built before the `try` block, so they do survive. That's the behavior worth locking down, because it's easy to break by moving record construction later.

[`test_is_enabled_returns_default_value_when_flags_request_fails`](https://github.com/PostHog/posthog-python/blob/9691157f91c9128e97e9949a67fd79eab1e259f3/posthog/test/test_evaluate_flags.py#L130) has the failure setup to mirror:

```python
    @mock.patch("posthog.client.flags")
    def test_local_values_survive_a_failed_fallback(self, patch_flags):
        patch_flags.side_effect = Exception("flags request failed")

        flags = self.client.evaluate_flags(
            "user-1", flag_keys=["local-flag", "remote-only"]
        )

        self.assertEqual(flags.keys, ["local-flag"])
        self.assertTrue(flags.get_flag("local-flag"))
```

posthog/client.py
```diff
@@ -4164,6 +4175,8 @@ def evaluate_flags(
                     response.get("minimalFlagCalledEvents") is True
                 )
                 for key, detail in response.get("flags", {}).items():
+                    if requested_keys is not None and key not in requested_keys:
```
haacked: `question`: This filter drops flags from the snapshot that users on older self-hosted PostHog instances get today. Is that narrowing something you want called out in the changeset?

Against Cloud it's a no-op, because the server already returns only the requested keys. It's not a no-op against a self-hosted instance older than 2025-08-18: `/flags` only gained `flag_keys_to_evaluate` support in PostHog/posthog#36807, and this SDK posts to `/flags/?v=2`. On those servers the response carries every flag for the user and all of it currently reaches the snapshot, so after this change `flags.is_enabled("some-unrequested-flag")` stops returning the server's value and returns the default instead.

I read the filter itself as correct and intentional. The spec already required values outside the requested set to be dropped before the snapshot is returned, and without it the new fallback would flood exactly those older servers' snapshots with every flag. This is only about whether the upgrade note should say so.

.sampo/changesets/gallant-princess-kalma.md
```diff
@@ -0,0 +1,5 @@
+---
+pypi/posthog: minor
```
haacked: `nit`: The PR title says `fix(flags):` while this asks for a `minor` bump, so the squash commit will read like a patch on a release that bumps the minor version. Nothing breaks, since sampo takes the level from this front matter and the title lint only checks Conventional Commits shape, but `feat(flags):` would keep the two in sync.

PR #880, opened 2026-08-20
fix: Drop events when before_send callbacks raise

marandaneto:

A `before_send` callback that raised an exception caused the SDK to continue with the original event. This could send data that the callback intended to redact or reject.

This change logs the callback error and drops the event instead. Invalid callback output is handled the same way. The event is never enqueued or sent, including when the callback mutates it before raising.

- `uv run --extra test pytest posthog/test/test_before_send.py --verbose --timeout=30`
- `uv run ruff format --check posthog/client.py posthog/test/test_before_send.py`
- `uv run ruff check posthog/client.py posthog/test/test_before_send.py`
- `git diff --check`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Pi coding agent helped audit the existing behavior, update the implementation and tests, run validation, and prepare this PR. The chosen behavior follows the cross-SDK contract: callback failures are contained and the event is dropped rather than sent without the callback's intended processing.

### DIFF
--- a/.sampo/changesets/stalwart-king-erika.md
+++ b/.sampo/changesets/stalwart-king-erika.md
@@ -1,0 +1,5 @@
+---
+pypi/posthog: patch
+---
+
+Drop events when before_send callbacks raise exceptions
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -2303,7 +2303,7 @@ class Client(object):
                 msg = clean(modified_msg)
             except Exception as e:
                 self.log.exception(f"Error in before_send callback: {e}")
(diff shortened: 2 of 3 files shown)


turnipdabeets (approved):
Approving. `return None` lands on an `_enqueue` path callers already handle, so there's no ripple to the return contract, and the tests pin both lanes — the sync-mode case asserts `batch_post` was never called, the queued case asserts `_analytics_lane.enqueue` was never called, and both mutate the event before raising. Matches `before-send-hook` spec §Error handling.

PR #1, opened 2020-02-18
Fix typo with "identify" code example

mariusandra:
Hi, I found a typo :)

PR #881, opened 2026-08-20
feat(mcp): support MCP Python SDK v2 and cross-SDK parity with TS SDK

gesh:
The [2026-07-28 MCP spec revision](https://blog.modelcontextprotocol.io/posts/2026-07-28/) removed the `initialize` handshake and `Mcp-Session-Id` header, and MCP Python SDK v2 (`mcp==2.0.0`, now the default on `pip install mcp`) reshaped every seam `posthog.mcp` hooks. The Python twin of the `@posthog/mcp` v2 work (PostHog/posthog-js#4449, PostHog/posthog-js#4565).

**Before this PR, `instrument()` crashes on mcp≥2**: `_compatibility.py` imported `mcp.server.fastmcp` at module scope — gone in v2 — and the `ImportError` propagated straight into the host app (worse than the JS silent no-op).

Mega issue: https://github.com/PostHog/posthog/issues/64016

## Fixed by this PR — was failing before

| Capability | Before (mcp≥2 / spec 2026-07-28) | Now |
|---|---|---|
| `instrument()` on a v2 server | 💥 `ImportError` crashes the host app at startup | ✅ Instruments `MCPServer` and low-level `Server`; unknown shapes degrade to a logged no-op |
| Tool call capture (`$mcp_tool_call`) | ❌ Nothing captured on v2 | ✅ Captured on both protocol eras, both server levels |
| Error capture (`$mcp_is_error` + `$exception`) | ❌ Nothing | ✅ Raised errors and `is_error` results, re-raised untouched |
| `tools/list` + intent (`context` injection/stripping) | ❌ Nothing | ✅ Same strip-vs-leave policy as the 1.x adapters |
| Client identity + `$mcp_protocol_version` | ❌ Nothing | ✅ From the handshake, the session token, or the per-request 2026-07-28 `_meta` envelope |
| Sessions on 2026-07-28 (spec removed the session header) | ❌ Every request its own `$session_id` — also on SDK 1.x | ✅ Conversation-anchored `$session_id` with `enable_conversation_id`, byte-parity with `@posthog/mcp`, works across stateless pods — anchored only once the handle is confirmed delivered |
| Stateless session token on v2 (legacy era) | ❌ Dead (crash) | ✅ Minted at `initialize`, carries identity + session across pods (wire-verified) |
| Handlers registered after `instrument()` (v2) | ❌ Never wrapped (the #4449 pattern) | ✅ `add_request_handler` patched; late registrations captured |
| Invented `conversation_id` echoes (e.g. `conv-1`) | ⚠️ Accepted verbatim — two unrelated callers could merge into one session once anchored | ✅ uuidv7 minted-shape guard; invented values get a fresh mint (JS parity) |
| Conversation handle on a failing first call | ⚠️ Prompt-back dropped on errored results — retry started a new conversation | ✅ Prompt-back rides errored results (JS parity) |
| Conversation handle on tools with structured output | ❌ Delivered only as a `content` text block — clients that read `structuredContent` never render `content`, so the agent had nothing to echo (0% echo rate, measured against Claude Code) | ✅ Also declared on the tool's `outputSchema` and mirrored into `structuredContent` on every response |
| Reading HTTP headers in `identify` / `event_properties` callbacks | ⚠️ v1 exposed no way to reach them from `extra`; v2 exposed only an undocumented path, so one callback body couldn't serve both majors (silent failure → every event anonymous) | ✅ Uniform `extra["ctx"]` on all adapters + exported `get_request_headers(extra)` |
| Raw request `ctx` in captured events | ⚠️ The whole `extra` dict was embedded in `$identify` parameters and stringified by truncation | ✅ Callbacks still get the full `ctx`; captured events carry only a JSON-safe scalar projection |
| CI coverage | ⚠️ v1 only, buried inside the whole-repo `tests` matrix | ✅ Named checks per SDK major — `MCP SDK v1 (Python 3.10/3.14)` + `MCP SDK v2 (Python 3.10/3.14)` — plus dual-era wire tests on every PR |

### Test gate first
- The MCP suite now runs twice in CI: the existing `tests` matrix stays on `mcp>=1.26,<2`; a new **`tests-mcp`** matrix runs `posthog/test/mcp` per SDK major — named checks `MCP SDK v1/v2 (Python 3.10/3.14)`; the v2 legs swap in `mcp>=2,<3`. A `conftest.py` splits collection by installed major.
- `test_v2_mcpserver.py` / `test_v2_lowlevel.py` — the v2 adapters driven directly, mirroring the v1 files.
- `test_v2_wire_dual_era.py` — raw JSON-RPC over the real streamable-http app (deliberately not the SDK `Client`, which negotiates the legacy era) in **both protocol eras**, stateless topology: envelope client identity, no session header on 2026-07-28, conversation-anchored `$session_id` across two fresh server instances, self-encoded token surviving pods on the legacy era.
- `test_conversation_session.py` — the cross-SDK derivation contract, with byte-for-byte vectors validated against `posthog-js` (`deriveSessionIdFromConversation`).
- `test_no_crash.py` — `instrument()` degrades to a logged no-op on both majors.

Screenshot 2026-08-20 at 16 00 03

### SDK changes
- **New `_instrument_v2.py`**: high-level `MCPServer` wraps `ToolManager.call_tool` (every dispatch routes through it; late-registered tools covered) + tools/list at the low-level registry; low-level `Server` wraps the string-keyed registry via the public `add_request_handler`/`get_request_handler` API — **late registrations included** (the #4449 lesson). Strip-vs-leave policy matches the 1.x pair.
- **Import-tolerant, shape-based detection** (JS ADR-0005): probes answer False instead of raising; unsupported servers → logged no-op handle, never a crash.
- **Conversation-anchored sessions** (JS ADR-0004, previously missing in Python): with `enable_conversation_id`, `$session_id` derives deterministically from the agent's `conversation_id` — the only correlation that survives 2026-07-28's per-request instances. Includes the minted-shape (uuidv7) guard so two callers inventing `conv-1` never merge, lowercased echoes, and per-request (never sticky) resolution. New export: `derive_session_id_from_conversation`. Anchoring applies on 1.x too — it is not era-gated, matching JS.
  - **The session is resolved only after the handle's fate is known.** A handle the SDK mints is not a fact until the agent has it, so anchoring on it up front strands events in a session nobody holds whenever the prompt-back can't be delivered (a tool that raises past our seam, a result with nowhere to put it) — one orphan session per call, plus a duplicate `$identify` each time. Resolving after delivery gives both properties: a delivered handle anchors the call that minted it, so the whole conversation is one session; an undelivered one anchors nothing. The shared per-server session is settled *before* dispatch so an in-tool `analytics.capture()` is attributed to the current caller rather than the previous one.
  - **The handle travels as data, not an instruction.** It is returned as `{"conversation_id": "..."}` rather than a `[SERVER]: Reuse ...` sentence: a tool result is untrusted content, so an imperative from the server is exactly what a client's prompt-injection filter strips — and a stripped block breaks the feature silently. It also kept appearing in users' transcripts. Same payload as `@posthog/mcp`.
- **Prompt-back rides errored results** (JS parity): a tool failing on the first call of a conversation is exactly when the agent needs the handle; previously the minted id was dropped on `isError` results.
- **Dual-shape reads** (`isError`/`is_error`, `inputSchema`/`input_schema`, `clientInfo`/`client_info`); captured payloads dump `by_alias` so both majors emit the camelCase wire shape. Version advisory widened to `mcp>=1.26,<3`.
- No new runtime deps; `mcp` stays a lazily-imported peer dependency. jlowin's `fastmcp` keeps the 1.x seams (it pins `mcp<2`).

### Follow-ups found while manually testing against real clients
- **`structuredContent` delivery channel** (new `_output_instructions.py`, ports posthog-js #4430/#4431). Manual testing against Claude Code showed the conversation feature was inert for any tool declaring an output schema: such clients render `structuredContent` and never see the `content` text block, so the handle was invisible and every call minted a fresh one. Now an optional `_mcp_instructions` key is declared on the tool's advertised `outputSchema` at `tools/list` and mirrored into `structuredContent` on *every* response. The declaration is what makes the write safe — clients validate against the advertised schema, so an undeclared key fails the customer's whole result under `additionalProperties: false`. Only declared tools are written to; an instance that never served a listing fails closed; composed schemas (`oneOf`/`allOf`/`anyOf`/`$ref`) and tools owning the key are skipped and stay content-only.
- **Uniform callback context + `get_request_headers`** (new `request_headers.py`). Adding v2 support had left the callback contract divergent — v2 passed the request context as `extra["ctx"]`, v1 passed only `session_id` — so an `identify()` written for one major silently got nothing on the other, and on v1 headers were unreachable from `extra` entirely. All four adapters now pass the SDK's own per-request context identically, and the exported helper flattens Starlette `Headers`, plain mappings, or any iterable of pairs into a lowercase-keyed dict (`None` on stdio, never raises). Mirrors `@posthog/mcp`'s `getRequestHeaders` and its ADR-0006 stance: hand hosts the raw context rather than synthesising a fake uniform shape.
- **`$identify` no longer captures the raw request context** (review finding). Callbacks still receive the full `extra` — including the v2 `ctx` so hosts can read headers — but captured event parameters now carry only a JSON-safe scalar projection, so an opaque context object can't be stringified by truncation into an event.

### Deliberate 1.x behavior changes (parity with `@posthog/mcp`)
1. An **invented** (non-uuidv7) `conversation_id` echo is replaced with a fresh mint instead of accepted verbatim — required before anchoring `$session_id` to it.
2. Minted prompt-backs are appended to **errored** results too (tests updated accordingly).

## Review loop (`/pr-shepherd`, 4 rounds)

An automated review loop found and fixed **8 further defects — 3 HIGH — after CI was already green**. Each was reproduced before being fixed and each fix has a regression test verified to fail without it; round 4 came back clean. Detail and per-finding status in the [QA Swarm summary comment](https://github.com/PostHog/posthog-python/pull/881#issuecomment-5358333479).

The three HIGHs, all in code added by this PR:
1. **Strict output schemas broke after a tool-cache rebuild** (`5ed2086`) — mcp 1.x validates `structuredContent` against its *cached* tool definition, rebuilt from the internal `req=None` listing pass we skipped injecting on. One call to an unlisted tool name and every later call to a tool with `additionalProperties: false` came back `isError` — analytics destroying a customer's successful result.
2. **Cross-caller conversation-handle leak** (`5ed2086`) — the mirror and prompt-back mutated the caller's result object in place, so a tool returning a shared/cached result served one conversation's handle to every later caller; since the handle outranks the transport session, that collapses unrelated clients into one session and identity merging can then write one user's person properties onto another's. Both channels now copy instead of mutating.
3. **The first fix was incomplete** (`e335876`) — the raw low-level adapter had the same cache trap on the *input* side, rejecting the very arguments the SDK tells agents to send.

Plus: a second `tools/list` silently disabling the mirror, an unguarded `request_context` read crashing the public `FastMCP.call_tool()`, a parameter marked required in the schema we strip it from (`strict_input_validation=True` → every call failed), a narrowed guard in the tool-call hot path, and a docstring that walked hosts toward putting a bearer token in a person property. A simplify pass (`07dd1c8`) reused existing helpers in the v2 adapter.

Base was also 3 commits behind, and `main` had changed `before_send` semantics (#880) and regenerated the public API snapshot this PR edits — merged and re-verified (`d7f2f39`).

## Validation
- v1 lane: `uv sync --extra test` → full `posthog/test/mcp` green (199 passed), plus the whole-repo suite (only pre-existing, network-dependent `posthog/test/ai` failures, reproduced on a clean tree).
- v2 lane: same env with `mcp==2.0.0` (fastmcp removed) → 182 passed, 8 skipped.
- `ruff format --check` / `ruff check`, `mypy` (baseline clean), `make public_api_check` (snapshot regenerated for the new export), `uv build` all green.
- Cross-SDK vectors computed from the TS `deterministicPrefixedId` and asserted byte-for-byte in `test_conversation_session.py`.

Out of scope (matching JS): prompts/resources auto-instrumentation, MRTR/`input_required`, tasks, `subscriptions/listen`, `server/discover` capture; posthog.com docs update (the Python example there still shows the v1 import path) is a follow-up in the docs repo.

---


posthog/mcp/_instrument_v2.py
```diff
@@ -0,0 +1,611 @@
...
+        arguments: Dict[str, Any],
+        context: Any = None,
+        convert_result: bool = False,
+    ) -> Any:
+        ctx = getattr(context, "_request_context", None)
+        token, client_name, client_version, protocol_version, mcp_session_id = (
+            _resolve_ctx(ctx)
+        )
+        request = build_tool_call_request(name, arguments)
+        extra: Dict[str, Any] = {"session_id": mcp_session_id, "ctx": ctx}
```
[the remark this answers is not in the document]
gesh: Fixed in 021823f. `handle_identify` now captures only a JSON-safe scalar projection of `extra` (`_captured_extra()` in `_internal.py`) — callbacks still receive the full dict including the raw `ctx` so hosts can read headers, but the opaque object can no longer reach captured parameters and be stringified by truncation. Regression test: `test_identify_callback_sees_ctx_but_capture_does_not` asserts the callback sees `ctx` while the captured $identify parameters carry only `session_id`. (Lines 384/490 shared the same single capture sink — `handle_identify` was the only place `extra` was embedded into an event — so the projection covers all three call sites.)

posthog/mcp/session.py
```diff
@@ -29,25 +29,54 @@ def derive_session_id_from_mcp_session(mcp_session_id: str) -> str:
...
+    source that survives a stateless / multi-pod deployment.
 
     The token session is resolved *per request*, never sticky: ``data`` is shared
     by every client hitting this server instance, so reusing a stored token session
     for a request that didn't replay the token would merge unrelated clients under
     one ``$session_id``. A compliant client replays the header on every request, so
     a genuine token session never needs the fallback.
     """
+    if conversation_id:
+        return derive_session_id_from_conversation(conversation_id)
```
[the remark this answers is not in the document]
gesh: Working as designed — this is the deliberate cross-SDK contract with `@posthog/mcp` (its ADR-0004), which ships this exact behavior today. The 2026-07-28 revision removed protocol sessions, so two stateless pods that never met have **no shared state** through which to verify mint provenance; deterministic agreement on the echoed handle is the entire point of the feature. The guard against the merge hazard is upstream in `resolve_conversation_id`: only a handle the SDK *could have minted* (uuidv7 shape, 74 random bits) anchors — an invented value like `conv-1` is replaced with a fresh mint, covered by `test_invented_handle_is_not_anchored` and mirroring the JS test 'does not merge two callers that invent the same conversation_id'. Two independent clients presenting the same uuidv7 means the handle was leaked or deliberately copied, which no stateless scheme can distinguish; diverging here would split the same conversation into different sessions depending on which SDK served the call (byte-parity vectors are pinned in `test_conversation_session.py`). Also note the feature is opt-in (`enable_conversation_id=False` by default).

gesh:
> [!NOTE]
>
> Multi-perspective review: router (cheap-first pass) + delegated reviewers (technical + security lenses), run as a 4-round quality loop

## Verdict: ✅ APPROVE <sub>(round 4 @ d7f2f39 — loop converged, round 4 dry)</sub>

Four rounds. **Eight defects found and fixed — three of them HIGH** — on a PR whose CI was already green. Every finding was reproduced before being fixed, and every fix has a regression test **verified to fail without it**. Round 4 came back clean.

### Findings and status

| # | Severity | Finding | Status |
| :-- | :-- | :-- | :-- |
| 1 | 🟠 HIGH | Strict output schemas broke after a tool-cache rebuild: mcp 1.x validates `structuredContent` against its *cached* tool definition, rebuilt from the internal `req=None` listing pass we skipped injecting on. One call to an unlisted name — our own `get_more_tools` will do it — and every later call to a tool with `additionalProperties: false` returned `isError`. Analytics destroying a customer's successful result. | ✅ Fixed `5ed2086` |
| 2 | 🟠 HIGH | Cross-caller handle leak: the mirror and prompt-back mutated the caller's result **in place**, so a tool returning a shared/cached `CallToolResult` served one conversation's handle to every later caller. Because the handle outranks the transport session, that collapses unrelated clients into one session — where identity merging writes one user's person properties onto another's profile. Reproduced on both SDK majors. | ✅ Fixed `5ed2086` |
| 3 | 🟠 HIGH | Fix #1 was incomplete — the raw low-level adapter had the same cache trap on the **input** side, and worse: it advertises `context`/`conversation_id` without stripping them, so a rebuilt cache rejected the very arguments we tell agents to send. | ✅ Fixed `e335876` |
| 4 | 🟡 MEDIUM | A second `tools/list` silently disabled the mirror: `add_instructions_to_output_schema` couldn't tell its own prior declaration from a customer's, so servers returning persistent `Tool` objects flipped ownership `True → False` — switching the feature off for exactly the clients it exists for, and blaming the customer in the log. | ✅ Fixed `5ed2086` |
| 5 | 🟡 MEDIUM | Unguarded `getattr(context, "request_context", None)`: FastMCP's property *raises* outside a request and `getattr` only swallows `AttributeError`, so the public `FastMCP.call_tool()` entry point began raising from analytics. Clean regression against `main`. | ✅ Fixed `5ed2086` |
| 6 | 🟡 MEDIUM | We marked `context` **required** in the very schema we strip it from before the SDK validates — under `FastMCP(strict_input_validation=True)` every call failed with `'context' is a required property`. Latent on `main`, made permanent by fix #3's cache injection. | ✅ Fixed `45f5630` |
| 7 | 🟢 LOW | The simplify pass narrowed an exception guard, so a malformed `properties` could raise (or answer by substring) in the tool-call hot path. | ✅ Fixed `e335876` |
| 8 | ⚪ NIT | `get_request_headers` docstring example read `headers.get("authorization")` inside `identify` — one step from a bearer token landing in a person property, the one payload area redaction never touches. | ✅ Fixed `5ed2086` |

### Convergence

The in-place-mutation defect (#2) was flagged independently by both delegated reviewers — the technical lens rated it LOW (contingent on result caching), the security lens traced it through to cross-user PII contamination and rated it HIGH. Resolved at the root: the SDK no longer mutates caller-owned objects at all.

Three of the eight (#3, #6, #7) were defects *in the previous round's fixes* — which is the argument for the loop running to convergence rather than stopping at one pass.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🧭 router (sonnet) | Read every hunk with context, cross-checked the v2 SDK source, ran both lanes green. Danger MEDIUM / confidence HIGH, **delegated nothing** — the null result was wrong. Escalated anyway per the caller model policy (public API + broad abstraction ⇒ second-model validation); that override is why anything was found. |
| 🔬 qa-team/correctness (fable) | Rounds 1–4. Confirmed the wrapper plumbing sound (idempotency, `delivered` accounting, exception isolation), then found #1, #3, #4, #6, #7 — each with a runnable probe. |
| 🔐 security-audit (opus) | Confirmed the capture side clean (`by_alias` is a pure rename, v1/v2 field parity, the earlier `_captured_extra` fix holds), then found the leak running the *other* way: #2, #5, #8. |
| 🧹 simplify (sonnet) | Reused existing helpers in the v2 adapter (header/session-id normalisation, `schema_has_param`, `params_to_request_dict`); declined to collapse the three adapters, which diverge deliberately. `07dd1c8` |

### Also in this iteration

- **Base updated.** The branch was 3 commits behind, and `main` had changed `before_send` semantics (#880) *and* regenerated `references/public_api_snapshot.txt`, which this PR also edits. Merged and re-verified rather than letting it fail after merge.
- **CI:** 40 passing, 0 failing, no repair needed.
- Local at `d7f2f39`: MCP SDK v1 197 passed · MCP SDK v2 182 passed/6 skipped · ruff, mypy, public-API snapshot clean.

<details>
<summary>Previous rounds (3)</summary>

- round 1 @ `e9c490d` — ⚠️ REQUEST CHANGES: 2 HIGH, 2 MEDIUM, 1 NIT (#1, #2, #4, #5, #8) → fixed in `5ed2086`; simplify pass `07dd1c8`
- round 2 @ `07dd1c8` — ⚠️ REQUEST CHANGES: 1 HIGH, 1 LOW, 1 NIT (#3, #7, dedup) → fixed in `e335876`
- round 3 @ `e335876` — 💬 1 MEDIUM (#6) → fixed in `45f5630`

</details>

---

posthog/mcp/_instrumentation.py
```diff
@@ -263,7 +277,9 @@ async def prepare_request(
     ``$mcp_initialize`` is anonymous even when identify resolves on the same request.
     (Still not byte-parity with the TS SDK, which wraps the real initialize handler;
     the Python SDK handles initialize in the session layer, not ``request_handlers``.)"""
-    session_id = await resolve_session_id(data, mcp_session_id, token=token)
+    session_id = await resolve_session_id(
+        data, mcp_session_id, token=token, conversation_id=conversation_id
```
marandaneto, line 308: blocking: Session is anchored before a minted conversation handle is delivered

`prepare_request()` derives the session from a newly minted `conversation_id` and immediately emits identify/initialize before either adapter knows whether the result can carry the prompt-back. On an MCPServer/FastMCP exception converted outside `ToolManager.call_tool`, or a non-injectable result, the client never receives this handle, so a retry mints another session while these events remain in an unreachable conversation-derived session. Please delay anchoring newly minted handles until delivery is confirmed, or wrap the outer error conversion so the handle is always returned; cover official FastMCP/MCPServer error and non-injectable-result paths.

posthog/mcp/request_headers.py
```diff
@@ -0,0 +1,75 @@
+# Portions of this package are derived from MCPCat/mcpcat-typescript-sdk
+# Copyright (c) 2025 MCPcat
+# Licensed under the MIT License: https://github.com/MCPCat/mcpcat-typescript-sdk/blob/main/LICENSE
+
+"""Read HTTP request headers inside a host callback, on either MCP SDK major.
+
+``identify``, ``intent_fallback``, ``event_properties`` and ``before_send``
```
marandaneto: blocking: `before_send` documentation promises an unsupported `extra["ctx"]` argument

`before_send` is typed and invoked as `before_send(capture)` in `_sink._apply_before_send`; a callback following this documented `(capture, extra)` contract raises `TypeError`, which is caught and silently drops each event. Please either remove the `before_send` claims here, in `posthog/mcp/__init__.py`, and the changeset, or add backward-compatible request-context plumbing while preserving one-argument callbacks and surfacing callback-signature failures.

posthog/mcp/_instrumentation.py
gesh: Confirmed and fixed in ff81024. Reproduced first: two failing calls produced **two orphan sessions**, each with its own duplicate `$identify` and `$mcp_initialize` — a conversation-derived session nobody holds, and the next call mints another, which is strictly worse than not anchoring at all. Took the first option you offered: an echo is the only proof of delivery, so a freshly minted handle no longer anchors — the minting call stays in the transport/memory session it would have used anyway, and every call after the agent returns the handle joins the conversation's session, so undeliverable prompt-backs degrade to pre-feature behaviour instead of fragmenting. Covers the MCPServer/FastMCP outer-conversion path and non-injectable results alike, since neither can produce an echo. The cross-pod contract is intact where it matters: a pod receiving an echoed handle still derives the same session as any other pod with no shared state — it just starts from the echo rather than the mint. Regression test added (two failed calls must stay in one session), and the three tests that asserted the old mint-anchors behaviour now document the new contract. Worth noting `@posthog/mcp` anchors mints the same way today, so I'll raise the equivalent fix there.

posthog/mcp/request_headers.py
gesh: Correct — `_apply_before_send` calls `before_send(capture)` with one argument, so a callback following the documented `(capture, extra)` shape would raise `TypeError` and have every event silently dropped. Removed the claim in ff81024 from the `request_headers` docstring, `posthog/mcp/__init__.py` and the changeset rather than plumbing a request context in: `before_send` is handed the finished capture payload, not the request, so `extra` doesn't belong there — `identify`, `intent_fallback` and `event_properties` are the three that genuinely receive it.

posthog/mcp/_instrument_fastmcp.py
```diff
@@ -173,17 +195,28 @@ async def wrapped(
...
+            if minted:
+                injected = _inject_prompt_back(result, conversation_id)
+                if injected is not result:
+                    delivered = True
+                result = injected
+                # Only a minted handle can be lost — one the agent supplied, it has.
+                if not delivered:
+                    delivered_conversation_id = None
+
+        session_id = await _session(delivered_conversation_id)
```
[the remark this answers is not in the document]
gesh: Fixed in 80c9c35 — a real regression from the previous commit. Moving `prepare_request()` after the call left `data.session_id` holding the prior request's value while the tool body ran, so an in-tool `analytics.capture()` was attributed to the previous caller (and through the identity cache, their person). `prime_session()` now settles the transport/memory session before dispatch, with the conversation anchor still finalised after delivery — an in-tool event predates that decision, so the transport session is the correct answer available at that moment. Regression test drives two callers with distinct `Mcp-Session-Id` headers and asserts B's body sees B's session; verified red without the fix. Applied at all four sites you listed.

marandaneto (approved):

PR #884, opened 2026-08-21
refactor(ai): share processor privacy and capture policy

marandaneto:

The Claude Agent SDK and OpenAI Agents processors duplicated privacy redaction and event capture policy. Privacy handling is security-sensitive, and changes could drift between processors.

This change shares privacy handling for duck-typed clients. It also centralizes common property merging, capture capability checks, identity fallback, event submission, and error handling while preserving existing behavior. In particular, a per-call privacy setting can enable redaction but cannot disable processor-level or client-level privacy.

- `uv run pytest --verbose --timeout=30 posthog/test/ai/claude_agent_sdk posthog/test/ai/openai_agents`: 99 passed
- `make public_api_check`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Pi and Slopo identified the shared privacy and capture policy. The implementation keeps Claude per-call privacy and groups handling separate where they differ. Review feedback identified an unintended privacy-precedence change, which was corrected with regression coverage before the branch was updated.


marandaneto:

posthog/ai/claude_agent_sdk/processor.py
```diff
@@ -434,20 +419,16 @@ def _emit_generation(
         }
 
         if input_messages is not None:
-            properties["$ai_input"] = (
-                None
-                if privacy
-                else self._with_privacy_mode(
-                    finalize_ai_content(input_messages, self._client)
-                )
+            properties["$ai_input"] = _with_privacy_mode(
```
turnipdabeets, line 420: [question] On main the old `self._with_privacy_mode` also consulted `self._privacy_mode`, so a processor built with `privacy_mode=True` stayed redacted even when a call passed `posthog_privacy_mode=False`. Here that same call captures the content instead (checked against main: `$ai_input` comes back `None` there). Looks deliberate — it matches the "Override privacy mode for this query" docstring and what `_emit_tool_span` already did — but nothing fails at build time for anyone relying on the old precedence. Worth calling out in the description rather than leaving it to the release notes?

turnipdabeets (approved):

posthog/ai/claude_agent_sdk/processor.py
marandaneto: Good catch. This is a behavior-preserving refactor, so I restored the previous precedence in 3e12bc4: per-call privacy can enable redaction but cannot disable processor-level privacy. I also updated the regression test and PR description.

PR #888, opened 2026-08-22
refactor(mcp): share instrumentation lifecycles

marandaneto:

The MCP v1 and v2 adapters duplicated setup, teardown, error capture, and missing-capability reporting for tools, prompts, and resources. Keeping those lifecycle paths separate made it easy for telemetry behavior to drift.

This refactor introduces shared instrumentation lifecycle helpers while leaving adapter-specific SDK calls in each adapter. It preserves support for missing-capability requests that omit tool arguments and avoids reprocessing customer-owned `conversation_id` schema fields.

- MCP SDK v1 suite before the latest `main` merge: 221 passed
- MCP SDK v2 suite before the latest `main` merge: 205 passed, 9 skipped
- `uv run pytest --verbose --timeout=30 posthog/test/mcp` after merging `main` and addressing feedback: 231 passed
- `make public_api_check`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Pi and Slopo identified the repeated MCP instrumentation lifecycle. The implementation keeps version-specific SDK interactions in their adapters and shares telemetry lifecycle state and capture behavior. Autoreview found that missing-capability requests may omit tool arguments, so the shared helper remains null-safe. Reviewer feedback also restored the previous no-warning behavior for tools that own `conversation_id`.


marandaneto:

posthog/mcp/_instrumentation.py
```diff
@@ -459,6 +602,57 @@ def read_tool_category(tool: Any) -> Optional[str]:
...
+    original_schema = schema
+    if is_context_enabled(data.options.context) and not owns_context:
+        schema = add_context_parameter_to_schema(
+            schema,
+            tool.name,
+            get_context_description(data.options.context),
+            required=context_required,
+        )
+    if data.options.enable_conversation_id:
+        schema = add_conversation_id_to_schema(schema, tool.name)
```
turnipdabeets, line 647: [nit] The low-level and v2 adapters used to pre-check `_schema_has_param(schema, "conversation_id")` before calling this. `add_conversation_id_to_schema` still no-ops correctly, but it logs `WARN: Tool "x" already has 'conversation_id'. Skipping injection.` — so a server that declares its own `conversation_id` now gets that line on every tools/list. Same behaviour, just noisier.

turnipdabeets (approved):

posthog/mcp/_instrumentation.py
marandaneto: Good catch. Fixed in 87f20ab by checking for a customer-owned `conversation_id` before running the injection helper. I added a regression test confirming the shared schema pipeline no longer emits that warning.

PR #889, opened 2026-08-22
refactor(openai): share resource and embedding policy

marandaneto:

The sync, async, and Azure OpenAI clients repeated the same resource discovery and replacement logic. Sync and async embedding wrappers also built the same telemetry properties separately. These copies could drift when OpenAI adds resources or embedding telemetry changes.

This refactor shares resource installation through explicit wrapper maps and uses a private helper for common embedding event construction. Provider dispatch and sync or async API calls remain in their existing adapters.

- `uv run pytest --verbose --timeout=30 posthog/test/ai/openai`: 79 passed, 1 skipped
- `make public_api_check`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Pi and Slopo identified the duplicated OpenAI resource and embedding paths. The resource-wrapping test now consumes the production wrapper maps so it detects mapping drift. Static type-only assignments remain because the enforced public API snapshot records those dynamically installed class attributes.


marandaneto:

posthog/test/ai/openai/test_resource_wrapping.py
```diff
@@ -0,0 +1,81 @@
...
+    WrappedChat as AsyncWrappedChat,
+)
+from posthog.ai.openai.openai_async import (
+    WrappedEmbeddings as AsyncWrappedEmbeddings,
+)
+from posthog.ai.openai.openai_async import (
+    WrappedResponses as AsyncWrappedResponses,
+)
+
+_SYNC_WRAPPERS = {
```
turnipdabeets: [suggestion] These re-declare the mapping the test is meant to guard, so adding a resource to `_SYNC_RESOURCE_WRAPPERS` without touching the test still passes — which is the drift this PR is fixing. Could we import `_SYNC_RESOURCE_WRAPPERS` / `_ASYNC_RESOURCE_WRAPPERS` here instead? (The `Wrapped*` imports above go away with them.)

posthog/ai/openai/openai.py
```diff
@@ -59,23 +60,13 @@ def __init__(self, posthog_client: Optional[PostHogClient] = None, **kwargs):
...
-        self._original_chat = getattr(self, "chat", None)
-        self._original_embeddings = getattr(self, "embeddings", None)
-        self._original_beta = getattr(self, "beta", None)
-        self._original_responses = getattr(self, "responses", None)
+        _wrap_openai_resources(self, _SYNC_RESOURCE_WRAPPERS)
 
-        # Replace with wrapped versions (only if originals exist)
-        if self._original_chat is not None:
+        # Keep dynamically installed resources visible to API and type inspection.
+        if _TYPE_CHECKING:
```
turnipdabeets: [suggestion] These never run, so this block and `_SYNC_RESOURCE_WRAPPERS` have to be kept in sync by hand — and there are four copies across `openai.py`, `openai_async.py`, and `openai_providers.py`. Class-scope annotations under `if _TYPE_CHECKING:` (`chat: WrappedChat`, ...) give mypy the same information without the fake constructor calls. (Worth noting the comment overstates it slightly: `check_public_api.py` imports modules and only sees module-level names, so it never looks at these.)

turnipdabeets (approved):

posthog/ai/openai/openai.py
marandaneto: I checked this against `make public_api_check`. The checker uses Griffe to inspect class attributes: removing these assignments removes the existing `OpenAI.chat`, `OpenAI.embeddings`, `OpenAI.beta`, and `OpenAI.responses` entries, while class annotations change their recorded API shape. Since this PR must preserve the existing public API snapshot, I am keeping the type-only assignments. Runtime resource installation remains driven solely by the wrapper mappings.

posthog/test/ai/openai/test_resource_wrapping.py
marandaneto: Good suggestion. Fixed in b910864: the resource-wrapping test now imports `_SYNC_RESOURCE_WRAPPERS` and `_ASYNC_RESOURCE_WRAPPERS` directly, so additions to the production mappings are covered without duplicating them in the test.

posthog/ai/openai/openai.py
marandaneto: I reconsidered this, and you’re right that class-scope annotations are the cleaner separation between the static resource API and the runtime wrapper mappings. Addressed in 425fec9: the dead constructor calls are replaced with class-scope `TYPE_CHECKING` annotations, using quoted forward references because the wrapper classes are defined later. I also regenerated the public API snapshot, which now records the same resource names as typed attributes rather than constructor expressions.

PR #898, opened 2026-08-26
feat(error-tracking): standardize exception event metadata

hpouillot:

Standardize Python exception events with the cross-SDK metadata contract so severity, capture source, mechanism semantics, nested linkage, and reserved property ownership are predictable.

Canonical contract: https://github.com/PostHog/sdk-specs/blob/main/openspec/specs/exception-event-metadata/spec.md

- Focused exception capture, Django, and Celery tests: 78 passed
- Ruff format and lint checks
- Focused mypy checks
- Public API snapshot check

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with pi from the shared OpenSpec contract. Integration-only capture metadata is kept out of the public API while manual and automatic capture paths preserve their existing behavior.

### DIFF
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -2014,7 +2014,9 @@ class Client(object):
         Args:
             exception: The exception to capture.
             distinct_id: The distinct ID of the user.
-            properties: A dictionary of additional properties.
+            properties: A dictionary of additional properties. Overriding reserved
+                exception properties is deprecated and will stop working in the next
+                major version.
             flags: A ``FeatureFlagEvaluations`` snapshot from ``evaluate_flags()``.
                 Attaches those exact flag values to the captured `$exception` event.
             send_feature_flags: Deprecated. Pass ``flags`` from ``evaluate_flags()`` instead.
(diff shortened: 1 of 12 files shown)


posthog/client.py
```diff
@@ -2070,11 +2080,38 @@ def capture_exception(
             )
             all_exceptions_with_trace_and_in_app = event["exception"]["values"]
 
+            reserved_properties = {
```
ablaszkiewicz, line 2085: should we add `$release_id`?

ablaszkiewicz (approved):

posthog/client.py
ablaszkiewicz: Actually it doesn't matter at this point! My work will cover that in the future

posthog/exception_utils.py
```diff
@@ -647,95 +711,90 @@ def exceptions_from_error(
...
 def exceptions_from_error_tuple(
     exc_info,  # type: ExcInfo
     mechanism=None,  # type: Optional[Dict[str, Any]]
 ):
     # type: (...) -> List[Dict[str, Any]]
     exc_type, exc_value, tb = exc_info
 
-    is_exception_group = BaseExceptionGroup is not None and isinstance(
-        exc_value, BaseExceptionGroup
+    (_, exceptions) = _exceptions_from_error(
```
marandaneto, line 791: blocking: Restrict group traversal to actual exception groups — Routing every exception through `_exceptions_from_error` now exposes ordinary exceptions to its `hasattr(exc_value, "exceptions")` group heuristic. A custom exception with an unrelated non-iterable `exceptions` attribute raises `TypeError`, which `Client.capture_exception` swallows and consequently drops the event; an iterable value instead creates phantom child exceptions. Previously, non-`BaseExceptionGroup` exceptions used `walk_exception_chain` and avoided this branch, so group-member traversal must remain restricted to actual `BaseExceptionGroup` instances.

marandaneto (commented):

marandaneto:
This change introduces backwards-incompatible public schema/API/contract changes: Reserved exception properties supplied through capture_exception(..., properties=...) can no longer override SDK-generated or processor-owned values.

If thats intentional, thats ok, but i think its not as designed, user provided props should always win

posthog/exception_utils.py
```diff
@@ -618,21 +674,29 @@ def walk_exception_chain(exc_info):
...
     Creates the list of exceptions.
     This can include chained exceptions and exceptions from an ExceptionGroup.
     """
 
+    if seen_exception_ids is None:
+        seen_exception_ids = set()
+    if exc_value is not None:
+        if id(exc_value) in seen_exception_ids or exception_id >= 50:
+            return (exception_id, [])
+        seen_exception_ids.add(id(exc_value))
```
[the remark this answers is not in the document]
hpouillot: This is intentional for the canonical exception graph contract: the same exception object must not be serialized more than once, including when it is referenced by multiple `ExceptionGroup` positions. The shared identity set also provides cycle safety. I added a regression test covering a repeated group member in 0fbbfc4.

cat-ph (approved):

Issue #103, opened 2023-10-18
Create an async version of posthog-python


neilkakkar:
For some users where local evaluation doesn't work, request blocking in python makes things much slower. Consider an async version of the library in this case

more context: https://posthoghelp.zendesk.com/agent/tickets/6553 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/9491)

Paillat-dev:
Yes please

benrapport:
+1!

sumersao:
Please this would be a life saving!

benrapport:
made one with only capture support for now: https://pypi.org/project/posthog-async/ just a fork of the main python package.

skulltech:
This would be much appreciated!!

michaelfeil:
+1

preardon:
+1

claudia-ciuciu:
+1

r-sniper:
> For some users where local evaluation doesn't work, request blocking in python makes things much slower. Consider an async version of the library in this case
> 
> more context: https://posthoghelp.zendesk.com/agent/tickets/6553 (moved to PostHog: https://us.posthog.com/project/2/support/tickets/9491)

Is the Zendesk request public, cause I see a not found page here

<img width="924" height="469" alt="Image" src="https://github.com/user-attachments/assets/e694932f-56d5-4dcf-b8d6-6477d3617fc4" />

marandaneto:
https://github.com/PostHog/posthog-python/pull/899
exmaples here https://github.com/PostHog/posthog-python/pull/899#issuecomment-5426395067

PR #900, opened 2026-08-26
feat: add async feature flag evaluation

marandaneto:

Continues the async SDK work for https://github.com/PostHog/posthog-python/issues/103. This PR is stacked on #899, supersedes #720, and incorporates its open review feedback.

The new `await AsyncPosthog.evaluate_flags(...)` method sends a non-blocking `/flags` request and returns the existing `FeatureFlagEvaluations` snapshot. Its `is_enabled()`, `get_flag()`, and `get_flag_payload()` accessors remain synchronous because they only read the returned snapshot. Passing that snapshot to `capture(flags=...)` attaches the exact values used for application branching.

This PR intentionally does not add the deprecated single-flag client methods or deprecated `capture(send_feature_flags=...)` behavior. It also does not add local flag-definition polling or fallback caches. The initial async API focuses on the reported problem: remote evaluation when local evaluation is unavailable. This avoids synchronous Redis access, poller lifecycle failures, and the deferred flag-called-event races reported on #720.

`$feature_flag_called` events use the synchronous buffered `capture()` method from #899. Deduplication therefore happens before control returns without scheduling extra tasks. The client also tracks flag and remote-config requests as in-flight operations so shutdown waits before closing its instance-owned HTTP transport.

`await get_remote_config_payload(...)` provides the corresponding non-blocking remote-config API with bearer authentication and encoded request paths.

All changes remain additive. Existing synchronous clients and module-level APIs are unchanged. The new constructor options are appended to the new async client's signature.

`playgrounds/fastapi-async-client` demonstrates the complete stacked API in FastAPI, including buffered and immediate capture, identify operations, exception capture, flag evaluation, remote config, flush, and lifespan-managed shutdown.

- `uv run ruff check .`
- `uv run --extra test pytest --timeout=30 -q` - 2306 passed, 15 skipped, 36 subtests passed
- `uv run --extra dev make public_api_check`
- `uv lock --check`
- Generated and imported the `posthoganalytics` mirror with the async flag APIs
- Autoreview against `origin/feat/async-client-capture-v2` reported no actionable findings at `598b7f8`

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with pi. We narrowed the old #720 API to the recommended snapshot-based evaluation method and remote config. Prior review findings guided the retry contract, request sanitization, shutdown tracking, flag-called deduplication, and the decision to omit deprecated and local-evaluation surfaces from this stack.


posthog/test/test_async_request.py
```diff
@@ -117,6 +128,96 @@ async def test_async_batch_post_rejects_absolute_request_path():
...
+    )
+
+    with mock.patch("posthog._async_request.asyncio.sleep") as sleep:
+        await async_flags(
+            "project-key",
+            "https://example.com",
+            timeout=3,
+            max_retries=1,
+            client=client,
+            distinct_id="user-1",
```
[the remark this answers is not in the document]
marandaneto: Good catch! Fixed in fb22cd6 — the retry success test now covers both one and multiple transient failures and asserts the full 0.3s/0.6s exponential backoff progression before recovery.

posthog/_async_request.py
```diff
@@ -84,6 +91,86 @@ def _process_response(response: Any) -> None:
...
+        for failed_attempt in range(retries + 1):
+            try:
+                data, headers = await asyncio.to_thread(
+                    _serialize_v0_body, api_key, False, request_data
+                )
+                response = await http_client.post(
+                    "/flags/?v=2", content=data, headers=headers, timeout=timeout
+                )
+                return await asyncio.to_thread(_process_flags_response, response)
+            except (httpx_module.TimeoutException, httpx_module.NetworkError):
```
dustinbyrne: `httpx.RemoteProtocolError` is a `TransportError`, but it is not a `NetworkError`. A server disconnect can therefore skip the configured retry. `evaluate_flags()` then catches the error and returns an empty snapshot. I reproduced this with `feature_flags_request_max_retries=1`: the first request raised `RemoteProtocolError`, the second would have succeeded, but the client made only one request. Could we include this transient error in the bounded retry policy and add a disconnect-then-success test?

dustinbyrne (commented):
Agent-led review, human-reviewed before posting.

dustinbyrne (approved):

posthog/_async_request.py
marandaneto: Good catch! Fixed in c36c79c — the bounded flag retry policy now catches httpx.TransportError, which includes RemoteProtocolError, NetworkError, and timeouts. Added a server-disconnect-then-success regression test that verifies the second attempt and backoff.

PR #902, opened 2026-08-27
fix(flags): align local case folding with the flags service

### LINKED ISSUE #78019: Case folding and numeric stringification are inconsistent between SDK local evaluation and the flags service
While shipping `starts_with`/`ends_with` local evaluation across the SDKs (#72992), a pattern emerged: our case-insensitive property matching is inconsistent between the SDKs and the flags service, and within the flags service itself. Filing this so we can settle it once instead of per code review.

## How the flags service folds case

In `rust/feature-flags/src/properties/property_matching.rs`:

- `icontains` / `starts_with` / `ends_with` (and negations): `to_ascii_lowercase()`, ASCII-only. The code comment says it's a deliberate perf choice.
- `exact` / `is_not`: full Unicode `to_lowercase()`.

So the server already applies two different folding rules depending on operator.

## What the SDKs use for the icontains/starts_with/ends_with family

| SDK | Folding | Matches server? |
|---|---|---|
| PHP | `strtolower` (ASCII-only) | ✅ |
| Ruby | `downcase(:ascii)` as of PostHog/posthog-ruby#229 | ✅ |
| Elixir | ASCII-only in PostHog/posthog-elixir#192 (pending) | ✅ |
| Python | `casefold()` (most aggressive Unicode; `"ss"` matches `"ß"`) | ❌ |
| Go | `strings.ToLower` (Unicode) | ❌ |
| Node | `toLowerCase()` (Unicode) | ❌ |
| Rust SDK | `to_lowercase()` (Unicode) | ❌ |
| .NET | `OrdinalIgnoreCase` (Unicode simple folding) | ❌ |

Concretely: person property `"Äbc"` with `starts_with: "ä"` matches locally in Python/Go/Node/Rust/.NET and does **not** match on `/flags`. Same flag, same person, different answer depending on evaluation path — exactly the divergence local evaluation is supposed to avoid.

## Adjacent inconsistencies in the same code

- The Rust SDK's `exact` operator uses `eq_ignore_ascii_case` while the server's `exact` is Unicode — the mirror image of the divergence above.
- Numeric stringification: the server renders a float-typed `323.0` as `"323.0"` (serde `Display`), as do Ruby and Python, but Go and Node render `"323"`, so a JSON `323.0` property matches `ends_with: "3"` locally but not remotely. .NET additionally used the host locale for decimal separators (`3.14` → `"3,14"` on de-DE), fixed in PostHog/posthog-dotnet#270.

## Options

1. **Standardize the SDKs on ASCII-only folding** for the icontains/starts/ends family to match the server. PHP and Ruby are already there; the rest are one-line changes plus tests. The numeric stringification fixes for Go/Node belong in the same sweep.
2. **Make the server Unicode-aware** and leave the SDKs alone. Friendlier for international text, but changes existing server behavior, and the ASCII choice was deliberate.
3. **Leave it and document** the edge case.

I lean toward 1, plus an explicit decision on whether the server's Unicode `exact`/`is_not` is intentional. Non-ASCII values differing only by case are rare, but today the answer depends on which SDK and which operator you hit, and that's the worst version.

marandaneto:

Local feature flag evaluation returned different `exact` and string-operator results from the released `/flags` service.

This fixes PostHog/posthog#78019 against the service's current Rust matcher. A separate draft proposes simplifying the backend's unusual boolean behavior: https://github.com/PostHog/posthog/pull/90694. This PR intentionally mirrors the released behavior until that proposal merges.

The evaluator now:

- applies the backend boolean-like filter gate before normal array membership
- preserves the backend's current empty-array and aggregate boolean-array behavior
- serializes JSON values with compact arrays and recursively sorted object keys
- formats representable Python floats like `serde_json`, including exponent thresholds and negative zero
- treats non-finite numbers and non-JSON values as inconclusive so evaluation can fall back to `/flags`
- uses full Unicode lowercase for `exact` and `is_not`
- uses ASCII-only lowercase for contains, prefix, and suffix operators

Python preserves integer and float types after JSON parsing, but it does not preserve a number's original source spelling. Matching uses the canonical spelling produced from the parsed value, as the backend does with `serde_json::Value`.

Red-green regression coverage reproduces the released backend's boolean precedence, empty arrays, composite JSON, float spellings, final sigma, and dotted capital I.

- `uv run pytest -q` - 2,285 passed, 15 skipped
- `uv run ruff check .`
- `uv run python .github/scripts/check_public_api.py`
- `uv run python -W error -c 'import posthog'`
- 100,000 deterministic finite float bit patterns matched `serde_json` formatting in a temporary differential check
- `autoreview --mode local` - no actionable findings after one accepted blocker was fixed

## :pencil: Checklist

### If releasing new changes

The patch changeset is `.sampo/changesets/standardize-flag-case-folding.md`.

## 🤖 Agent context

Pi used the address-pr-comments, check-pr, autoreview, and karpathy-guidelines skills. The final review caught a non-finite value bypass in the boolean path, which was fixed before publication. A human must review before merge.

### DIFF
--- a/.sampo/changesets/standardize-flag-case-folding.md
+++ b/.sampo/changesets/standardize-flag-case-folding.md
@@ -1,0 +1,5 @@
+---
+pypi/posthog: patch
+---
+
+Match local feature flag string operators using the same ASCII and Unicode lowercasing rules as the flags service.
--- a/.sampo/changesets/standardize-flag-case-folding.md
+++ b/.sampo/changesets/standardize-flag-case-folding.md
@@ -2,4 +2,4 @@
 pypi/posthog: patch
 ---
 
-Match local feature flag string operators using the same ASCII and Unicode lowercasing rules as the flags service.
+Match local feature flag string operators using the flags service's boolean coercion, JSON stringification, and casing rules.
--- a/posthog/feature_flags.py
+++ b/posthog/feature_flags.py
@@ -1,9 +1,12 @@
 import calendar
 import datetime
 import hashlib
+import json
 import logging
+import math
 import re
 import warnings
+from decimal import Decimal
 from enum import Enum
 from typing import Optional
 
@@ -512,8 +515,82 @@ _ASCII_LOWER_TRANSLATION = str.maketrans(
 )
 
 
+def _format_json_float(value: float) -> str:
+    if not math.isfinite(value):
+        raise InconclusiveMatchError(
+            "Non-finite property values cannot be represented by the flags service"
+        )
+
+    representation = repr(value)
+    if "e" not in representation:
+        return representation
+
+    mantissa, exponent_text = representation.split("e")
+    exponent = int(exponent_text)
+    if -6 < exponent < 0:
+        return format(Decimal(representation), "f")
+
+    sign = "+" if exponent >= 0 else ""
+    return f"{mantissa}e{sign}{exponent}"
+
+
+def _json_value_to_string(value) -> str:
+    if value is None:
+        return "null"
+    if isinstance(value, bool):
+        return "true" if value else "false"
+    if isinstance(value, str):
+        return json.dumps(value, ensure_ascii=False, separators=(",", ":"))
+    if isinstance(value, int):
+        return str(value)
+    if isinstance(value, float):
+        return _format_json_float(value)
+    if isinstance(value, list):
+        return "[" + ",".join(_json_value_to_string(item) for item in value) + "]"
+    if isinstance(value, dict):
+        if not all(isinstance(key, str) for key in value):
+            raise InconclusiveMatchError(
+                "Property object keys must be strings for local evaluation"
+            )
+        items = (
+            f"{json.dumps(key, ensure_ascii=False)}:{_json_value_to_string(value[key])}"
+            for key in sorted(value)
+        )
+        return "{" + ",".join(items) + "}"
+
+    raise InconclusiveMatchError(
+        f"Property value of type {type(value).__name__} is not JSON-compatible"
+    )
+
+
+def _value_to_string(value) -> str:
+    if isinstance(value, str):
+        return value
+    return _json_value_to_string(value)
+
+
 def _ascii_lower(value) -> str:
-    return str(value).translate(_ASCII_LOWER_TRANSLATION)
+    return _value_to_string(value).translate(_ASCII_LOWER_TRANSLATION)
+
+
+def _is_truthy_or_falsy_property_value(value) -> bool:
+    if isinstance(value, bool):
+        return True
+    if isinstance(value, str):
+        return value.lower() in ("true", "false")
+    if isinstance(value, list):
+        return all(_is_truthy_or_falsy_property_value(item) for item in value)
+    return False
+
+
+def _is_truthy_property_value(value) -> bool:
+    if isinstance(value, bool):
+        return value
+    if isinstance(value, str):
+        return value.lower() == "true"
+    if isinstance(value, list):
+        return all(_is_truthy_property_value(item) for item in value)
+    return False
 
 
 def match_property(property, property_values) -> bool:
@@ -535,16 +612,27 @@ def match_property(property, property_values) -> bool:
 
     override_value = property_values[key]
 
-    if (operator not in NONE_VALUES_ALLOWED_OPERATORS) and override_value is None:
+    if (
+        operator not in NONE_VALUES_ALLOWED_OPERATORS
+        and operator != "exact"
+        and override_value is None
+    ):
         return False
 
     if operator in ("exact", "is_not"):
 
         def compute_exact_match(value, override_value):
-            override_value = str(override_value).lower()
+            override_string = _value_to_string(override_value).lower()
+            if _is_truthy_or_falsy_property_value(value):
+                return _is_truthy_property_value(value) == _is_truthy_property_value(
+                    override_value
+                )
+
             if isinstance(value, list):
-                return override_value in [str(val).lower() for val in value]
-            return str(value).lower() == override_value
+                return override_string in [
+                    _value_to_string(candidate).lower() for candidate in value
+                ]
+            return _value_to_string(value).lower() == override_string
 
         if operator == "exact":
             return compute_exact_match(value, override_value)
--- a/posthog/feature_flags.py
+++ b/posthog/feature_flags.py
@@ -7,7 +7,6 @@ import warnings
 from enum import Enum
 from typing import Optional
 
-from posthog import utils
 from posthog.types import FlagValue
 from posthog.utils import convert_to_datetime_aware, is_valid_regex
 
@@ -507,6 +506,13 @@ def is_condition_match(
 # branch in match_property. Distinct from the unknown-operator rejection at the top
 # of the function so the dispatch-completeness test can tell the two apart.
 _UNHANDLED_OPERATOR_MESSAGE = "has no match_property branch"
+_ASCII_LOWER_TRANSLATION = str.maketrans(
+    "ABCDEFGHIJKLMNOPQRSTUVWXYZ", "abcdefghijklmnopqrstuvwxyz"
+)
+
+
+def _ascii_lower(value) -> str:
+    return str(value).translate(_ASCII_LOWER_TRANSLATION)
 
 
 def match_property(property, property_values) -> bool:
@@ -534,34 +540,37 @@ def match_property(property, property_values) -> bool:
     if operator in ("exact", "is_not"):
 
         def compute_exact_match(value, override_value):
+            override_value = str(override_value).lower()
             if isinstance(value, list):
-                return str(override_value).casefold() in [
-                    str(val).casefold() for val in value
-                ]
-            return utils.str_iequals(value, override_value)
+                return override_value in [str(val).lower() for val in value]
+            return str(value).lower() == override_value
 
         if operator == "exact":
             return compute_exact_match(value, override_value)
         else:
             return not compute_exact_match(value, override_value)
 
-    if operator == "icontains":
-        return utils.str_icontains(override_value, value)
+    if operator in (
+        "icontains",
+        "not_icontains",
+        "starts_with",
+        "not_starts_with",
+        "ends_with",
+        "not_ends_with",
+    ):
+        property_string = _ascii_lower(override_value)
+        filter_string = _ascii_lower(value)
 
-    if operator == "not_icontains":
-        return not utils.str_icontains(override_value, value)
+        if operator in ("icontains", "not_icontains"):
+            matched = filter_string in property_string
+        elif operator in ("starts_with", "not_starts_with"):
+            matched = property_string.startswith(filter_string)
+        else:
+            matched = property_string.endswith(filter_string)
 
-    if operator == "starts_with":
-        return utils.str_istartswith(override_value, value)
-
-    if operator == "not_starts_with":
-        return not utils.str_istartswith(override_value, value)
-
-    if operator == "ends_with":
-        return utils.str_iendswith(override_value, value)
-
-    if operator == "not_ends_with":
-        return not utils.str_iendswith(override_value, value)
+        if operator.startswith("not_"):
+            return not matched
+        return matched
 
     if operator == "regex":
         return (
--- a/posthog/feature_flags.py
+++ b/posthog/feature_flags.py
@@ -7,6 +7,7 @@ import warnings
 from enum import Enum
 from typing import Optional
 
+from posthog import utils as utils
 from posthog.types import FlagValue
 from posthog.utils import convert_to_datetime_aware, is_valid_regex
 
--- a/posthog/test/test_feature_flags.py
+++ b/posthog/test/test_feature_flags.py
@@ -217,7 +217,7 @@ class TestLocalEvaluation(unittest.TestCase):
         self.assertEqual(person_properties, {"region": "USA"})
         self.assertEqual(patch_flags.call_count, 0)
 
-    def test_case_insensitive_matching(self):
+    def test_exact_matching_uses_unicode_lowercase(self):
         self.client.feature_flags = [
             {
                 "id": 1,
@@ -259,6 +259,16 @@ class TestLocalEvaluation(unittest.TestCase):
                 "person-flag",
                 "some-distinct-id",
                 person_properties={"location": "straße"},
+                only_evaluate_locally=True,
+            )
+        )
+
+        self.assertFalse(
+            self.client.get_feature_flag(
+                "person-flag",
+                "some-distinct-id",
+                person_properties={"location": "strasse"},
+                only_evaluate_locally=True,
             )
         )
 
@@ -266,19 +276,17 @@ class TestLocalEvaluation(unittest.TestCase):
             self.client.get_feature_flag(
                 "person-flag",
                 "some-distinct-id",
-                person_properties={"location": "strasse"},
+                person_properties={"star": "ſun"},
+                only_evaluate_locally=True,
             )
         )
 
-        self.assertTrue(
+        self.assertFalse(
             self.client.get_feature_flag(
-                "person-flag", "some-distinct-id", person_properties={"star": "ſun"}
-            )
-        )
-
-        self.assertTrue(
-            self.client.get_feature_flag(
-                "person-flag", "some-distinct-id", person_properties={"star": "sun"}
+                "person-flag",
+                "some-distinct-id",
+                person_properties={"star": "sun"},
+                only_evaluate_locally=True,
             )
         )
 
--- a/posthog/test/test_feature_flags.py
+++ b/posthog/test/test_feature_flags.py
@@ -5102,22 +5102,23 @@ class TestMatchProperties(unittest.TestCase):
         with self.assertRaises(InconclusiveMatchError):
             match_property(property_c, {"key2": "value"})
 
-    def test_match_properties_exact_uses_unicode_lowercase(self):
-        matching_cases = [("Ä", "ä"), ("323.0", 323.0)]
-        non_matching_cases = [("ß", "ss"), ("Σ", "ς")]
+    @parameterized.expand(
+        [
+            ("non_ascii_case_variant", "Ä", "ä", True),
+            ("float_stringification", "323.0", 323.0, True),
+            ("casefold_expansion", "ß", "ss", False),
+            ("final_sigma", "Σ", "ς", False),
+        ]
+    )
+    def test_match_properties_exact_uses_unicode_lowercase(
+        self, _name, filter_value, property_value, expected
+    ):
+        exact = self.property("key", filter_value, "exact")
+        is_not = self.property("key", filter_value, "is_not")
+        self.assertEqual(expected, match_property(exact, {"key": property_value}))
+        self.assertEqual(not expected, match_property(is_not, {"key": property_value}))
 
-        for filter_value, property_value in matching_cases:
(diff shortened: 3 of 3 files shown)


posthog/test/test_feature_flags.py
```diff
@@ -5094,6 +5102,29 @@ def test_match_properties_exact(self):
         with self.assertRaises(InconclusiveMatchError):
             match_property(property_c, {"key2": "value"})
 
+    def test_match_properties_exact_uses_unicode_lowercase(self):
+        matching_cases = [("Ä", "ä"), ("323.0", 323.0)]
```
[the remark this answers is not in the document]
marandaneto: Good catch! Fixed in ac6fd92 — the exact-match and float-stringification cases now use the repository’s parameterized-test pattern, and the array-specific assertions remain in a focused test.

posthog/feature_flags.py
```diff
@@ -534,34 +541,37 @@ def match_property(property, property_values) -> bool:
...
+    ):
+        property_string = _ascii_lower(override_value)
+        filter_string = _ascii_lower(value)
+
+        if operator in ("icontains", "not_icontains"):
+            matched = filter_string in property_string
+        elif operator in ("starts_with", "not_starts_with"):
+            matched = property_string.startswith(filter_string)
+        else:
+            matched = property_string.endswith(filter_string)
```
ioannisj: [nit] Should the last arm be explicit? A new operator added to the tuple but not to the chain would silently evaluate as `ends_with`, and this is the one branch that returns a wrong boolean instead of raising, which is what `_UNHANDLED_OPERATOR_MESSAGE` and the completeness test exist to catch.

```suggestion
        elif operator in ("ends_with", "not_ends_with"):
            matched = property_string.endswith(filter_string)
        else:
            raise InconclusiveMatchError(f"Operator {operator} {_UNHANDLED_OPERATOR_MESSAGE}")
```

ioannisj (approved):

arnohillen:
`compute_exact_match` in `posthog/feature_flags.py:543-547` (pre-existing, but inside the block this PR rewrites) has no equivalent of the service's leading boolean-coercion branch (`is_truthy_or_falsy_property_value`, `rust/feature-flags/src/properties/property_matching.rs:212-220`), so an `exact` filter of `false` or `"false"` against property `"banana"` returns `False` locally while the service matches, and filter `[]` against property `true` returns `False` locally vs `true` on the service.

marandaneto:
Addressed the current-backend parity concerns in signed commit `83e1c27`. The focused regression matrix reproduced `false`/`"banana"` and `[]`/`true` as local `False` before the fix; both now match the released backend as `True`. It also covers boolean-array precedence and the `is_not` complement. The same commit adds compact recursively sorted JSON serialization, serde-compatible formatting for representable floats, stronger final-sigma/dotted-I tests, and inconclusive fallback for non-JSON values. The full Python suite, Ruff, Mypy, public API check, import check, and autoreview pass.

posthog/feature_flags.py
```diff
@@ -507,6 +510,87 @@ def is_condition_match(
...
+            raise InconclusiveMatchError(
+                "Property object keys must be strings for local evaluation"
+            )
+        items = (
+            f"{json.dumps(key, ensure_ascii=False)}:{_json_value_to_string(value[key])}"
+            for key in sorted(value)
+        )
+        return "{" + ",".join(items) + "}"
+
+    raise InconclusiveMatchError(
```
haacked: `blocking`: A person property holding a `UUID` or `Decimal` stops resolving flags after this change. Today it matches; after this change, it doesn't. `_json_value_to_string` raises `InconclusiveMatchError` for anything outside `None`, `bool`, `str`, `int`, `float`, `list`, and `dict`. Both the `exact`/`is_not` path and the `icontains`/`starts_with`/`ends_with` path now route through it.

The remote fallback doesn't rescue them. `json.dumps(body, cls=DatetimeSerializer)` raises `TypeError` on those types before any HTTP call. The catch-all turns that into the stale-cache path. That `TypeError` predates this PR; what is new is that local evaluation sends these values down it instead of matching them.

I ran this against the PR base (`5e80659`) and this head, with the flags request mocked:

| property value | base | head, `only_evaluate_locally=True` | head, default |
|---|---|---|---|
| `UUID(...)` | `True` | `None` | `None` (no HTTP) |
| `Decimal("9.99")` | `True` | `None` | `None` (no HTTP) |
| `datetime(2024, 1, 1)` | `True` | `None` | `True` (HTTP round trip) |

`person_properties` is typed `Optional[Dict[str, Any]]`. The date operators in this same function already dispatch on `isinstance(override_value, datetime.datetime)`, so these are values the matcher is built to accept. This is a patch release.

Coercing them before stringification keeps local matching on the spelling the service receives.

```python
def _coerce_to_json_value(value):
    if isinstance(value, Decimal):
        return float(value)
    if isinstance(value, UUID):
        return str(value)
    if isinstance(value, (datetime.datetime, datetime.date)):
        return value.isoformat()
    if isinstance(value, tuple):
        return list(value)
    return value
```

Call it at the top of `_json_value_to_string` and `_value_to_string`. Two caveats: this matches what the service receives rather than restoring the old local result (base compared `str(datetime)`'s space-separated spelling, and `Decimal("9.90")` matched the filter `"9.90"` where `float` coercion gives `"9.9"`), and `set` should stay inconclusive since its iteration order is not stable. Nothing in the suite covers these types today, so CI will not catch this either way.

posthog/test/test_feature_flags.py
```diff
@@ -5094,6 +5102,115 @@ def test_match_properties_exact(self):
...
+
+    @parameterized.expand(
+        [
+            ("integral_float", "323.0", 323.0, True),
+            ("negative_zero", "-0.0", -0.0, True),
+            ("small_exponent", "1e-7", 1e-7, True),
+            ("large_exponent", "1e+16", 1e16, True),
+            ("fixed_small_decimal", "0.00001", 1e-5, True),
+            ("fixed_small_decimal_fraction", "0.000099", 9.9e-5, True),
+            ("python_padded_exponent", "1e-07", 1e-7, False),
```
haacked: `suggestion`: The spec rule that keeps integers and floats apart has no test. sdk-specs#51 requires an `exact` filter of `"323"` to match integer property `323`, and states that `"323"` must not match floating-point property `323.0` ("Integer and preserved floating-point values remain distinct"). That distinction is the reason `_json_value_to_string` has a separate integer branch, and no `exact` test in the file uses an integer property value at all, so nothing reaches that branch.

Both rows belong in this table:

```suggestion
            ("python_padded_exponent", "1e-07", 1e-7, False),
            ("integer_property", "323", 323, True),
            ("integer_filter_against_float", "323", 323.0, False),
```

Two smaller spec rows are also unrepresented, though both behave correctly today: the empty-array scenario lists property values `"true"`, `[true]`, `"us"`, `null`, and `1`, and only `True`, `[]`, and `False` appear in `test_match_properties_exact_uses_backend_boolean_precedence`.

posthog/feature_flags.py
```diff
@@ -528,40 +612,54 @@ def match_property(property, property_values) -> bool:
 
     override_value = property_values[key]
 
-    if (operator not in NONE_VALUES_ALLOWED_OPERATORS) and override_value is None:
+    if (
```
haacked: `suggestion`: This guard disagrees with the flags service in two directions. Both land on the same six lines, so they are worth settling together.

A property that is present but holds `None` returns `False` here for `icontains`, `starts_with`, `ends_with`, and their `not_` forms. The service stringifies JSON null to `"null"` and runs the substring check on that. `to_string_representation` in `rust/feature-flags/src/properties/property_matching.rs` falls through to `Value::to_string()` for non-strings. `matching_property_values.get(key)` hands the operator branch `Some(Value::Null)` rather than treating the property as absent. So `email not_icontains "@competitor.com"` against `person_properties={"email": None}` matches on the service and doesn't match locally. The guard returns a plain `False`, not `InconclusiveMatchError`, so nothing falls back to `/flags` and the two answers disagree.

In the other direction, the new `and operator != "exact"` clause lets a filter that has no `value` through. The service short-circuits that case with `let value = match &property.value { Some(v) => v, None => return Ok(false) }`. Locally both sides now stringify to `"null"` and `exact` returns `True`. The `is_not` version of that disagreement predates this PR, so only the `exact` half is new.

Exempting the search operators and short-circuiting a null filter value covers both.

```diff
+_JSON_NULL_STRINGIFIED_OPERATORS = (
+    "exact",
+    "icontains",
+    "not_icontains",
+    "starts_with",
+    "not_starts_with",
+    "ends_with",
+    "not_ends_with",
+)
+
     if (
         operator not in NONE_VALUES_ALLOWED_OPERATORS
-        and operator != "exact"
+        and operator not in _JSON_NULL_STRINGIFIED_OPERATORS
         and override_value is None
     ):
         return False
+
+    if operator in ("exact", "is_not") and value is None:
+        return False
```

`regex`, `gt`/`gte`/`lt`/`lte`, and the date operators still need the guard, so it stays for them. The `exact_null` and `false_array_matches_null` cases both pass a non-null filter value, so neither is affected. Whichever shape you land on, `NONE_VALUES_ALLOWED_OPERATORS` reads as the list of operators that tolerate a `None` property value. After this change it is no longer that list, so it is worth updating to match.

posthog/test/test_feature_flags.py
```diff
@@ -259,26 +259,34 @@ def test_case_insensitive_matching(self):
                 "person-flag",
                 "some-distinct-id",
                 person_properties={"location": "straße"},
+                only_evaluate_locally=True,
             )
         )
 
-        self.assertTrue(
+        self.assertFalse(
```
haacked: `suggestion`: These two `assertFalse` calls never observe a local `False`. They observe `None`, so they pass without testing the thing they name.

The `person-flag` flag has two condition groups, one filtering on `location` and one on `star`. Each call supplies only one of those properties. The group whose property is missing raises `InconclusiveMatchError`, so `match_feature_flag_properties` gives up on the whole flag. `only_evaluate_locally=True` returns `None`. `assertFalse(None)` passes. I ran it with one property supplied and got `None`; with both supplied I got `False`.

Neither case records that `strasse` was rejected as a match for `Straße`, only that the flag did not come back true.

Supplying the second property makes the evaluation conclusive, and `assertIs` then tells the two outcomes apart.

```python
self.assertIs(
    False,
    self.client.get_feature_flag(
        "person-flag",
        "some-distinct-id",
        person_properties={"location": "strasse", "star": "zzz"},
        only_evaluate_locally=True,
    ),
)
```

The same change applies to the `star`/`sun` case below, using `person_properties={"location": "zzz", "star": "sun"}`.

posthog/feature_flags.py
```diff
@@ -528,40 +612,54 @@ def match_property(property, property_values) -> bool:
...
+        operator not in NONE_VALUES_ALLOWED_OPERATORS
+        and operator != "exact"
+        and override_value is None
+    ):
         return False
 
     if operator in ("exact", "is_not"):
 
         def compute_exact_match(value, override_value):
+            override_string = _value_to_string(override_value).lower()
```
haacked: `suggestion`: Moving this line below the boolean gate would turn an `exact` filter of `False` against a property of `float("nan")` from an inconclusive fallback into a match. It looks movable because the branch on the next line never reads `override_string`.

What the call actually does there is run the finiteness check inside `_format_json_float`. `_is_truthy_property_value` has no float case, so without that raise both sides come out non-truthy and compare equal.

The `boolean_filter_with_nan` case in `test_match_properties_exact_non_json_numbers_are_inconclusive` fails if the line moves, so the regression does get caught. A reader tidying an apparently unused local has to run the suite to find out why it is there. This file already comments its traps at [lines 20-24](https://github.com/PostHog/posthog-python/blob/83e1c27f39d554d1e9c6764b6c5424077fff653f/posthog/feature_flags.py#L20-L24) and [509-511](https://github.com/PostHog/posthog-python/blob/83e1c27f39d554d1e9c6764b6c5424077fff653f/posthog/feature_flags.py#L509-L511).

```suggestion
            # Stringify before the boolean gate so a non-finite property value raises
            # instead of comparing as falsey.
            override_string = _value_to_string(override_value).lower()
```

posthog/feature_flags.py
```diff
@@ -1,13 +1,16 @@
...
 import logging
+import math
 import re
 import warnings
+from decimal import Decimal
 from enum import Enum
 from typing import Optional
 
-from posthog import utils
+from posthog import utils as utils
```
haacked: `suggestion`: The next person who tidies imports will delete this line and break the public API. Nothing in the file references `utils.` anymore. The import keeps the `posthog.feature_flags.utils` alias in [`references/public_api_snapshot.txt`](https://github.com/PostHog/posthog-python/blob/83e1c27f39d554d1e9c6764b6c5424077fff653f/references/public_api_snapshot.txt#L317); commit `6b6f16c` already had to restore it once. Deleting it and running `make public_api_snapshot` silently regenerates the snapshot without the alias, so the break becomes the new baseline.

```suggestion
# Re-exported so the posthog.feature_flags.utils alias stays in the public API snapshot.
from posthog import utils as utils
```

.sampo/changesets/standardize-flag-case-folding.md
```diff
@@ -0,0 +1,5 @@
+---
+pypi/posthog: patch
+---
+
+Match local feature flag string operators using the flags service's boolean coercion, JSON stringification, and casing rules.
```
haacked: `suggestion`: The changeset says the SDK now matches the service's rules, but not that specific flags will start evaluating to different values after a patch upgrade. Three changes are visible to anyone running local evaluation: an `exact` or `is_not` filter whose value is entirely boolean-like now compares truthiness instead of string equality, so a filter of `false` matches `"banana"`, `0`, and `null`, and a filter of `["true", "false"]` stops matching `"true"`; `exact` against a property of `None` now matches the filter `"null"` and no longer matches the filter `"none"`; and property values the SDK cannot spell the way the service does now defer to `/flags`.

The patch bump itself looks right, since 7.44.1 shipped the `is_set`/`is_not_set` parity fix the same way. This is about the text. The repo already writes changesets like this: the 7.43.0 entry closes with "This changes the previous behavior where the key was omitted without a request."

```suggestion
Match local feature flag string operators using the flags service's boolean coercion, JSON stringification, and casing rules. This changes local evaluation results in three cases: an `exact` or `is_not` filter whose value is entirely boolean-like (`false`, `"false"`, `[]`, `["true", "false"]`) now compares truthiness rather than string equality, so `false` matches falsey values such as `"banana"`, `0`, and `null`; `exact` against a `None` property now matches the filter `"null"` and no longer matches the filter `"none"`; and property values that cannot be represented the way the service represents them fall back to `/flags`.
```

posthog/test/test_feature_flags.py
```diff
@@ -5094,6 +5102,115 @@ def test_match_properties_exact(self):
...
+
+    @parameterized.expand(
+        [
+            ("nan", "value", float("nan")),
+            ("positive_infinity", "value", float("inf")),
+            ("negative_infinity", "value", float("-inf")),
+            ("boolean_filter_with_nan", False, float("nan")),
+        ]
+    )
+    def test_match_properties_exact_non_json_numbers_are_inconclusive(
```
haacked: `suggestion`: This test covers the inconclusive path for `exact` only, not for `icontains`, `starts_with`, or `ends_with`. Those operators now route both operands through `_value_to_string`, so a property holding a `datetime`, `UUID`, `Decimal`, or non-finite float stops evaluating locally where the old `utils.str_icontains` called `str(...)` and always returned a bool. All four types raise today.

The search-operator path for non-JSON types needs coverage. Right now the behavior is unasserted, so a change either way is invisible to CI.

This test also asserts only `exact` and never `is_not`, unlike every other new parameterized test in this group.

posthog/feature_flags.py
```diff
@@ -507,6 +510,87 @@ def is_condition_match(
...
+
+
+def _value_to_string(value) -> str:
+    if isinstance(value, str):
+        return value
+    return _json_value_to_string(value)
+
+
+def _ascii_lower(value) -> str:
+    return _value_to_string(value).translate(_ASCII_LOWER_TRANSLATION)
```
haacked: `suggestion`: `icontains`, `starts_with`, and `ends_with` got about 4x slower: `str.translate` walks the string doing a table lookup per character, where main's `.casefold()` was one C call. Measured, 1.062us per `match_property` call now against 0.256us, and this runs per property per condition per flag with no cap on flags. `.lower()` is byte-identical for an ASCII-only string and `isascii()` is O(1) in CPython. Run the check on the output of `_value_to_string`, not the raw argument, because nested content in a list or dict can serialize to non-ASCII.

```suggestion
    text = _value_to_string(value)
    return text.lower() if text.isascii() else text.translate(_ASCII_LOWER_TRANSLATION)
```

posthog/feature_flags.py
```diff
@@ -507,6 +510,87 @@ def is_condition_match(
...
+            "Non-finite property values cannot be represented by the flags service"
+        )
+
+    representation = repr(value)
+    if "e" not in representation:
+        return representation
+
+    mantissa, exponent_text = representation.split("e")
+    exponent = int(exponent_text)
+    if -6 < exponent < 0:
```
haacked: `suggestion`: This `-6` bound is the whole difference between matching the flags service's float spelling and diverging from it. Nothing nearby says where the number comes from. Checking it took four repositories: the spec's "same plain-decimal and exponent cutovers as the flags service" rule, then `serde_json`'s float writer, then the formatter crate it delegates to.

The condition is also narrower than it looks. Python's `repr` only produces exponent form at exponents at or below -5 and at or above 16, so -5 is the only exponent that can reach this branch. Spelling out the backend's rule rather than the reachable case is the right call, but only if a reader knows that is what it is doing.

```suggestion
    # The flags service prints plain decimal down to exponent -5 and exponent form below
    # it, while Python's repr switches at -5, so re-expand that one case.
    if -6 < exponent < 0:
```

posthog/test/test_feature_flags.py
```diff
@@ -5094,6 +5102,115 @@ def test_match_properties_exact(self):
         with self.assertRaises(InconclusiveMatchError):
             match_property(property_c, {"key2": "value"})
 
+    @parameterized.expand(
+        [
+            ("non_ascii_case_variant", "Ä", "ä", True),
+            ("float_stringification", "323.0", 323.0, True),
```
haacked: `nit`: This row duplicates `("integral_float", "323.0", 323.0, True)` in the float-stringification table exactly, and it exercises no Unicode lowercasing despite living in `test_match_properties_exact_uses_unicode_lowercase`. Dropping it here loses nothing.

posthog/feature_flags.py
```diff
@@ -507,6 +510,87 @@ def is_condition_match(
...
+    return f"{mantissa}e{sign}{exponent}"
+
+
+def _json_value_to_string(value) -> str:
+    if value is None:
+        return "null"
+    if isinstance(value, bool):
+        return "true" if value else "false"
+    if isinstance(value, str):
+        return json.dumps(value, ensure_ascii=False, separators=(",", ":"))
```
haacked: `nit`: `json.dumps` with non-default keyword arguments skips the module's cached encoder and builds a fresh `JSONEncoder` per call, costing 0.30us against 0.065us for a hoisted one. That hits this line for each string nested in a list or dict and the dict-key line below for each key, so a 50-key object goes from 17.4us to 6.0us; one module-level `_JSON_STRING_ENCODER = json.JSONEncoder(ensure_ascii=False, separators=(",", ":")).encode` covers both, since `separators` has no effect when encoding a bare string.

PR #906, opened 2026-08-31
fix(ai): omit token counts the provider never reported

bernatixer:

An LLM call whose provider never reported usage shows up as `$0.00` in LLM analytics. Every integration sent `$ai_input_tokens: 0` and `$ai_output_tokens: 0` for it, and ingestion prices a call by multiplying a rate with that count. The common case is a cancelled stream: it consumed its prompt, but usage only arrives at the end.

Token counts are now written only when they trace back to a provider report:

- A count the provider never reported is omitted, on every event writer: the shared capture paths in `posthog/ai/utils.py`, the OpenAI and Gemini embedding events, the LangChain callback, and the OpenAI Agents and Claude Agent SDK processors.
- The stream accumulators (OpenAI, Anthropic, Gemini) and the response extractors no longer seed counts at zero, so an interrupted stream carries only what actually arrived.
- A zero the provider reported is still sent as `0`. Zero keeps meaning a real report of nothing.
- LangChain cache and reasoning counts follow the same rule. They previously coalesced unreported values to `0`.
- `$ai_total_tokens` on the Agents processor is the sum of the reported sides, omitted when neither side reported.

This applies the PostHog/posthog-js#4664 semantic to this SDK, following the cross-SDK assessment on that PR. Ingestion distinguishes absent from zero since PostHog/posthog#90211. Older ingestion coalesces absent to 0 at read time, so this ships independently.

> [!NOTE]
> `$ai_input_tokens` and `$ai_output_tokens` were previously always present on wrapper events. Anything reading them downstream has to handle absence.

- A new parameterized test at the shared capture chokepoint covers three cases: no usage omits both counts, explicit zeros still send `0`, one-sided usage sends one side. It fails when the `.get(..., 0)` default is restored, verified by reverting it.
- A new Gemini streaming test covers a stream whose chunks never carry usage. It fails when the zero-seeded accumulator is restored, verified by reverting it.
- The existing tests that pinned the fabricated zeros now assert absence (Gemini embeddings, Agents spans without usage, LangChain cache and reasoning defaults). Assertions backed by an explicit zero in their fixture stay `== 0`, which pins the boundary.
- The `posthog/test/ai` suite passes locally: 648 passed, 5 skipped. The 9 failing tests make live provider calls that this sandbox's gateway rejects with 403, and they fail identically with the change stashed.
- `ruff format`, `ruff check`, and the CI mypy command (through `mypy-baseline filter`) pass.
- The public API snapshot is regenerated: `extract_gemini_embedding_token_count` now returns `Optional[int]`, the one intentional signature change.
- The red MCP SDK matrix is unrelated to this change: the repo's rolling `exclude-newer = "7 days"` quarantine made mcp 2.1.0 eligible between this PR's two CI runs, and mcp 2.1 masks unexpected tool errors, which fails one test on any branch including `main`. Three of the four red jobs are fail-fast cancellations of that one failure. Fixed in #907; re-running the failed check after it lands turns this PR green.

## :pencil: Checklist

### If releasing new changes

- [x] Added a changeset under `.sampo/changesets/` (hand-written in the `sampo add` format)

## 🤖 Agent context

Written by Claude Code in a PostHog Desktop session directed by @bernatixer, as the cross-SDK follow-up agreed on PostHog/posthog-js#4664. An audit found the fabrication concentrated at the shared property-write layer plus the zero-seeded accumulators. Cancellation handling needed no work here: the stream wrappers already capture from a `finally` and close the generator on early exit.

Decisions: provider-reported zeros keep sending `0`, pinned by two existing LangChain fixtures. `$ai_total_tokens` stays a sum of reported sides rather than being omitted on partial usage, matching the lower-bound totals decision on the JS PR. The unused zero defaults inside `_get_tokens_source` stay untouched to keep the diff behavioral.

Skills invoked: writing-tests, writing-pr-descriptions.

---

### DIFF
--- a/.sampo/changesets/omit-unreported-token-counts.md
+++ b/.sampo/changesets/omit-unreported-token-counts.md
@@ -1,0 +1,5 @@
+---
+pypi/posthog: patch
+---
+
+Omit `$ai_input_tokens` and `$ai_output_tokens` when the provider never reported usage, instead of sending `0`, so an interrupted stream no longer looks like a free call. A zero reported by the provider is still sent, and zero keeps meaning a real report of nothing. Covers the OpenAI, Anthropic, Gemini, LangChain, OpenAI Agents and Claude Agent SDK integrations.
--- a/posthog/ai/anthropic/_anthropic_stream.py
+++ b/posthog/ai/anthropic/_anthropic_stream.py
@@ -15,7 +15,7 @@ class _AnthropicStreamAccumulator:
     """Accumulates sync-neutral capture state from Anthropic stream events."""
 
     def __init__(self) -> None:
-        self.usage_stats: TokenUsage = TokenUsage(input_tokens=0, output_tokens=0)
+        self.usage_stats: TokenUsage = TokenUsage()
         self.accumulated_content = ""
         self.content_blocks: List[StreamingContentBlock] = []
         self.tools_in_progress: Dict[str, ToolInProgress] = {}
--- a/posthog/ai/anthropic/anthropic_converter.py
+++ b/posthog/ai/anthropic/anthropic_converter.py
@@ -252,13 +252,16 @@ def extract_anthropic_usage_from_response(response: Any) -> TokenUsage:
     Returns:
         TokenUsage with standardized usage
     """
-    if not hasattr(response, "usage"):
-        return TokenUsage(input_tokens=0, output_tokens=0)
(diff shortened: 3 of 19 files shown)


marandaneto (approved):

marandaneto:
CI is failing, not sure if related

PR #907, opened 2026-09-01
fix(mcp): report the exception a tool raised when the SDK masks it

bernatixer:

The MCP SDK CI matrix went red on Aug 31 with no repo change. The cause is the repo's rolling `exclude-newer = "7 days"` quarantine: mcp 2.1.0 (published Aug 24, 19:04 UTC) aged into eligibility at 19:04 UTC, between two runs of the same commit, so the unpinned `mcp>=2,<3` leg resolved 2.0.0 in the morning and 2.1.0 in the evening. Three of the four red jobs are fail-fast cancellations of the one real failure.

mcp 2.1 masks unexpected tool exceptions: the dispatch wrapper's message becomes `Error executing tool <name>` with the original kept only on `__cause__` (documented in the SDK at `tools/base.py`). `$mcp_error_message` and `$mcp_error_type` read `$exception_list[0]`, so every unexpected tool failure reported the same masked string and the failures view lost the reason a call failed, which is the one thing these properties exist to carry.

The scalars now read the entry behind the SDK's dispatch wrapper. `exceptions_from_error_tuple` already records the `__cause__` chain, so the original exception is sitting right behind the wrapper; the match requires the type name (`ToolError` / `UnexpectedToolError`), the `Error executing tool` message prefix, and a module from an SDK namespace (`mcp.`, `fastmcp.`), and it walks every consecutive wrapper because a tool invoking a failing tool is wrapped once per dispatch. A wrapper without a chained cause is kept as is, an application's own exception that merely shares the wrapper's name is kept, and the `$exception` sibling still carries the full chain, so error-tracking grouping is unchanged.

On mcp 1.x the wrapper includes the original text, so the scalars there change from `Error executing tool boom: explode` / `ToolError` to `explode` / `ValueError`. The tool name is already its own property on the event.

- The previously failing gate test (`test_instrumented_server_failure_carries_them`) passes untouched under every mcp version CI can resolve: 2.0.0 (yesterday's resolution), 2.1.0 (today's), and 2.1.1 (ages in Sep 1, verified with an `--exclude-newer` override).
- Four tests pin the unwrap through the public paths: a real SDK wrapper with a chained cause yields the cause's message and type while the `$exception` sibling keeps the wrapper; a wrapper without a cause is kept; an application exception sharing the wrapper's name is kept; and a nested-server test asserts both the inner and outer events report the root cause. The unit tests import the real per-major `ToolError` classes.
- Revert-proven at each round: against `main`, the gate test and the unwrap test fail; against the first revision of this PR, both review findings reproduce as failing tests (nested wrappers and the same-named application exception); with the final change, all pass.
- Full `posthog/test/mcp` suite: 235 passed on the lockfile env (mcp 1.28.1 + fastmcp 3.2.0), 215 passed + 13 skipped on each of mcp 2.0.0, 2.1.0, and 2.1.1.
- `ruff format`, `ruff check`, and the CI mypy command (through `mypy-baseline filter`) pass.
- Redaction covers the unwrapped entry: `_sanitize_exception_values` maps over every chain entry, not just the first.

## :pencil: Checklist

### If releasing new changes

- [x] Added a changeset under `.sampo/changesets/` (hand-written in the `sampo add` format)

## 🤖 Agent context

Written by Claude Code in a PostHog Desktop session directed by @bernatixer, after tracing why the MCP SDK matrix went red on PostHog/posthog-python#906 with a snapshot-only diff. Rejected alternatives: pinning the v2 leg below 2.1 (defeats the gate's purpose and ships masked messages to customers anyway), version-gating the test assertion (accepts the masked string as the product), and raising `ToolError` in the test (stops exercising the unexpected-exception path customers actually hit). The SDK documents `__cause__` as the channel for the original error, so the instrumentation reads it.

The advisory review's two blocking findings (nested double-wrapping, and type-name matching catching application exceptions) were both reproduced first and are fixed in the second commit: the match is gated on the entry's recorded SDK module and traverses consecutive wrappers.

Skills invoked: writing-tests, writing-pr-descriptions.

---

### DIFF
--- a/posthog/mcp/_posthog_events.py
+++ b/posthog/mcp/_posthog_events.py
@@ -157,6 +157,36 @@ def _add_common_properties(event: Event, properties: Dict[str, Any]) -> None:
         properties["$set"] = {**identify_actor_data}
 
 
+_TOOL_DISPATCH_WRAPPERS = ("ToolError", "UnexpectedToolError")
+
+
+def _primary_exception(error: Any) -> Dict[str, Any]:
+    """Pick the ``$exception_list`` entry that carries the failure reason.
+
+    The MCP SDK's tool dispatch re-raises whatever a tool raised as a
+    ``ToolError`` whose message starts ``Error executing tool <name>``, and
+    mcp >= 2.1 masks the original text out of that message entirely, keeping
+    it only on ``__cause__`` — the next entry of the chain here. The wrapper
+    says nothing the event's tool name does not already say, so the scalars
+    read the entry behind it. Matched by type name and message prefix because
+    each SDK major ships its own ``ToolError`` class.
+    """
+    if not isinstance(error, dict):
+        return {}
+    exception_list = error.get("$exception_list")
+    if not isinstance(exception_list, list) or not exception_list:
+        return {}
+    first = exception_list[0] if isinstance(exception_list[0], dict) else {}
+    if (
+        len(exception_list) > 1
+        and isinstance(exception_list[1], dict)
+        and first.get("type") in _TOOL_DISPATCH_WRAPPERS
+        and str(first.get("value", "")).startswith("Error executing tool")
+    ):
+        return exception_list[1]
+    return first
+
+
 def _add_error_details(event: Event, properties: Dict[str, Any]) -> None:
     """Surface the failure reason on the primary event itself.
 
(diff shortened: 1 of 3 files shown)


posthog/mcp/_posthog_events.py
```diff
@@ -157,24 +157,48 @@ def _add_common_properties(event: Event, properties: Dict[str, Any]) -> None:
...
+    if not isinstance(exception_list, list) or not exception_list:
+        return {}
+    first = exception_list[0] if isinstance(exception_list[0], dict) else {}
+    if (
+        len(exception_list) > 1
+        and isinstance(exception_list[1], dict)
+        and first.get("type") in _TOOL_DISPATCH_WRAPPERS
+        and str(first.get("value", "")).startswith("Error executing tool")
+    ):
+        return exception_list[1]
```
marandaneto: blocking: Unwrap consecutive dispatch wrappers — MCP 2.1 wraps an `UnexpectedToolError` again when an outer tool invokes a failing inner tool, so selecting only `$exception_list[1]` leaves the outer event reporting `Error executing tool inner` instead of the underlying failure. Reproduction: reproduced — a nested `MCPServer` test under MCP 2.1.0 observed that masked message, while traversing consecutive wrappers made it report `root failure`.

posthog/mcp/_posthog_events.py
```diff
@@ -157,24 +157,48 @@ def _add_common_properties(event: Event, properties: Dict[str, Any]) -> None:
...
+        return {}
+    exception_list = error.get("$exception_list")
+    if not isinstance(exception_list, list) or not exception_list:
+        return {}
+    first = exception_list[0] if isinstance(exception_list[0], dict) else {}
+    if (
+        len(exception_list) > 1
+        and isinstance(exception_list[1], dict)
+        and first.get("type") in _TOOL_DISPATCH_WRAPPERS
+        and str(first.get("value", "")).startswith("Error executing tool")
```
marandaneto: blocking: Avoid classifying wrappers solely by type and message text — `capture_tool_call` accepts arbitrary exceptions, so an unrelated application exception named `ToolError` with this prefix is replaced by its cause, corrupting the reported message and type. Reproduction: reproduced — an application-defined `ToolError("Error executing tool application task") from ValueError("inner detail")` reported `inner detail`/`ValueError` on this head but preserved the outer error on the merge base.

marandaneto (commented):

marandaneto (approved):

PR #914, opened 2026-09-02
feat: opt-in runtime metrics autocapture

frankh:
## Summary

Adds an opt-in `metrics_autocapture` mode that periodically samples a small, low-cardinality set of process/host runtime metrics and reports them to PostHog's Metrics product. **Off by default**, no new required dependencies, and no change for users who don't enable it.

```python
posthog = Posthog("phc_...", metrics_autocapture=True, metrics={"service_name": "billing-worker"})
```

## Approach: reuse the existing metrics pipe

The original proposal weighed two dependency strategies — (A) an OpenTelemetry SDK extra, or (B) a hand-rolled OTLP-JSON payload. The SDK already ships (B): `client.metrics` (`posthog/metrics_capture.py`) pre-aggregates samples and exports OTLP/JSON to `POST /i/v1/metrics` with correct delta temporality, gzip, fork-safety, retry/backoff, and a series cap.

So this change adds **neither** — the sampler feeds the existing, battle-tested pipe. No new required dependency, and none of the wire semantics (temporality, encoding) get re-implemented.

## What's collected

Stdlib-first, so the base SDK gains no required dependencies:

- `resource.getrusage(RUSAGE_SELF)` → user/system CPU seconds (cumulative → per-interval **count** deltas), peak RSS (gauge, normalized to bytes across platforms)
- `threading.active_count()` (gauge)
- `gc.get_stats()` collections per generation (cumulative → **count** deltas)
- `os.getloadavg()` (gauge; skipped on Windows)
- **Only when `psutil` is installed** — never required: CPU %, RSS/VMS, open FDs (gauges)

Cumulative sources are primed at `start()` and emitted as per-interval deltas via `metrics.count()` (delta temporality), so each data point stands alone and a restart needs no cross-run state. Every collector is individually guarded — a telemetry timer never raises into the host app.

## Design notes

- **Fork safety**: the sampler thread doesn't survive `fork()`, so it restarts per worker via the client's existing `os.register_at_fork` hook (gunicorn/celery each get a live sampler).
- **Per-process identity**: a fresh `service.instance.id` of `hostname-pid` is stamped per process (only when the user hasn't set one), so preforked workers keep separate series instead of clobbering each other's gauges.
- **Low cardinality**: only service/instance resource attributes — no per-request labels.
- **Config**: `metrics_autocapture` (bool, default `False`); cadence via `metrics={"autocapture_interval": 30}` (seconds, default 60); `service.instance.id` overridable through the metrics resource attributes.

The module is named `posthog/runtime_metrics.py` rather than `metrics_autocapture.py` to avoid shadowing the `posthog.metrics_autocapture` module-level config global.

## Tests & validation

- New `posthog/test/test_runtime_metrics.py` (24 tests): counter deltas, instance-id defaulting/override, interval validation, per-platform RSS normalization, collector isolation, psutil path, lifecycle, and end-to-end flow onto the metrics wire.
- New real-`fork()` test in `test_client_fork.py` proving per-worker sampler restart and instance-id repinning.
- `ruff format` + `ruff check` clean; `mypy` clean across the tree; public API snapshot regenerated.

---


posthog/runtime_metrics.py
```diff
@@ -0,0 +1,307 @@
...
+    def stop(self, final_sample: bool = True) -> None:
+        """Stop the sampler. By default takes one last sample so the final
+        window (flushed by the client on shutdown) isn't a blind spot."""
+        with self._lock:
+            self._active = False
+            self._stop_event.set()
+            thread = self._thread
+            self._thread = None
+        if thread is not None and thread is not threading.current_thread():
+            thread.join(timeout=_THREAD_JOIN_TIMEOUT_SECONDS)
```
marandaneto, line 153: blocking: Prevent in-flight samples after shutdown — The timed join allows `stop()` to return while a collector or metrics `before_send` hook is still running; `Client.shutdown()` can then flush and reset the metrics window, after which the old sampler resumes, records new metrics, and arms another network flush despite shutdown having completed. Reproduction: reproduced — `uv run --extra test pytest -q posthog/test/test_runtime_metrics.py::test_stop_does_not_allow_sampler_to_emit_after_it_returns` fails because a blocked sampler emits after `stop()` returns and the modeled reset; replacing the timed join with a blocking join made the control pass.

marandaneto (commented):

posthog/runtime_metrics.py
```diff
@@ -0,0 +1,307 @@
...
+        if final_sample:
+            self._sample_guarded()
```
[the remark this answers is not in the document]
frankh: Fixed in 0bf856b (guard) + 9757f26 (test). The timed join in `MetricsAutocapture.stop()` can indeed return while a collector or `before_send` hook is still running, so gating in the sampler alone cannot help once the thread is blocked inside `metrics.gauge()`/`count()`. The terminal guard now lives in `PostHogMetrics`: shutdown calls `_close()` (instead of `reset()`), which cancels the timer, drops the window, and sets a `_closed` flag that `_capture()` checks under the lock before folding or arming a timer — so a straggler sample can neither re-seed the window nor schedule a post-shutdown flush. `_close()` is used only at shutdown; `reset()` stays reusable, and the fork reinit clears `_closed` so a forked child records again.

Kept the timed join rather than a blocking one so a wedged collector or user hook can't hang shutdown indefinitely — the guard makes correctness independent of whether the join completes. Added `test_blocked_sample_after_shutdown_does_not_emit`, which models a sample blocked in a `before_send` hook across shutdown; verified it fails against the old `reset()` teardown (a flush timer is armed post-shutdown) and passes with `_close()`.

posthog/runtime_metrics.py
frankh: Fixed in 0bf856b (guard) + 9757f26 (test). The timed join in `MetricsAutocapture.stop()` can indeed return while a collector or `before_send` hook is still running, so gating in the sampler alone cannot help once the thread is blocked inside `metrics.gauge()`/`count()`. The terminal guard now lives in `PostHogMetrics`: shutdown calls `_close()` (instead of `reset()`), which cancels the timer, drops the window, and sets a `_closed` flag that `_capture()` checks under the lock before folding or arming a timer — so a straggler sample can neither re-seed the window nor schedule a post-shutdown flush. `_close()` is used only at shutdown; `reset()` stays reusable, and the fork reinit clears `_closed` so a forked child records again.

Kept the timed join rather than a blocking one so a wedged collector or user hook can't hang shutdown indefinitely — the guard makes correctness independent of whether the join completes. Added `test_blocked_sample_after_shutdown_does_not_emit`, which models a sample blocked in a `before_send` hook across shutdown; verified it fails against the old `reset()` teardown (a flush timer is armed post-shutdown) and passes with `_close()`.

marandaneto (approved):

PR #920, opened 2026-09-04
feat(ai): capture the served service tier into model parameters

bernatixer:

- An OpenAI call served on the flex tier bills at half the standard token rates, and one served on priority at double — but events captured by this SDK carry no tier, so LLM analytics prices them all at standard (flex overstated 2x, priority understated 2x).
- PostHog ingestion resolves tier pricing from `$ai_model_parameters.service_tier` (PostHog/posthog#94200), and `posthog-js` already records it from the response. This PR is the Python side of the same contract.
- The value is the tier the provider served, read from the response — a requested tier can be refused, so request kwargs never supply it.

- Non-streaming: a response carrying `service_tier: "flex"` lands it in `$ai_model_parameters`; a response without the field leaves the parameters unchanged.
- Streaming: both OpenAI stream states (chat and Responses API) pick the tier up from chunks, mirroring how they already track `model`; the field threads through `StreamingEventData` as `NotRequired`, so the Anthropic/Gemini builders are untouched (they have no tier concept).
- LangChain: the callback merges `llm_output["service_tier"]` — which langchain lifts from the response — next to the existing invocation params.
- Full `posthog/test/ai` suite: 655 passed.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

- Claude Code session directed by the assignee, mirroring the JS SDK's served-tier capture as the companion to PostHog/posthog#94200.
- The changeset file was written by hand following the existing changesets' format.

### DIFF
--- a/.sampo/changesets/capture-served-service-tier.md
+++ b/.sampo/changesets/capture-served-service-tier.md
@@ -1,0 +1,5 @@
+---
+pypi/posthog: minor
+---
+
+OpenAI generations now record the service tier the provider served (`service_tier` inside `$ai_model_parameters`), on non-streaming, streaming, and LangChain capture paths. LLM analytics uses it to price flex and priority calls at their real rates instead of standard; a requested tier can be refused, so the value always comes from the response.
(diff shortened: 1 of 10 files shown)


Radu-Raicea (approved):
We should do this for the JS SDK too.

PR #922, opened 2026-09-05
fix(flags): honor versioned local property matching

marandaneto:

Local feature flag evaluation needs to honor the `property_matching_version` supplied with flag definitions. Otherwise Python can return a different result from the service when a project uses version 2 matching.

This follows the [backend behavior](https://github.com/PostHog/posthog/pull/90694) and [shared SDK contract](https://github.com/PostHog/sdk-specs/pull/59).

- Exactly version 2 uses explicit scalar equality and member equality for nonempty filter arrays. Missing/1 and unsupported selectors keep legacy matching. Empty filters retain recursive truthiness. For known properties, `is_not` complements `exact`.
- The selector travels with person, group, nested cohort, and flag dependency evaluations, including multi-flag evaluations.
- Definition snapshots and synchronous/asynchronous definition cache providers retain the selector. Version-only refreshes invalidate evaluated results even when flags are unchanged.
- The first result after lazy definition loading is cached for fallback during an API outage. Definition publication advances an in-memory generation without waiting for Redis I/O. Separate write locks prevent an old in-flight write from overwriting a newer result within a Client. Locks are replaced after fork, and memory writes prune invalidated flags for reused users.
- Client-managed Redis entries now include a private deterministic fingerprint of flags, cohorts, group mapping, and matching version. Normal and stale reads reject entries from other snapshots, including late writes and entries written by another worker. Matching snapshots can reuse results across workers even when their local generation counters differ. Fingerprint serialization happens outside the publication lock for API and provider loads.
- Every snapshot change advances the local generation, including initial hydration and cohort/group-mapping-only changes. Writes retain the identity of the evaluation that produced them rather than adopting the definitions current at write completion.

### Compatibility

The optional definition-cache field and matcher arguments are additive and included in the public API snapshot. `Client` supports async definition providers here. The separate `AsyncClient` remains remote-only. The existing Sampo changeset is a patch for `pypi/posthog`.

Redis keys, TTLs, public cache signatures, and standalone invalidation/clear behavior are unchanged. The new entry metadata is additive and readable by older SDKs. Older or unverifiable entries become cache misses for upgraded Client-managed caches. Older SDK readers do not gain these protections.

Before definitions load, remote results can still serve outage fallback within the same Client. Their private identity is unique to that Client and renewed after fork, so new Clients and workers cannot reuse unverifiable remote-only entries. Authentication or billing resets disable Redis result caching until valid definitions load again. Remote requests use their request-start generation only as a local invalidation boundary, not as proof of the server's definition snapshot.

Redis result writes remain synchronous. This change does not add Redis timeouts or make single-result cache writes nonblocking. Old entries can remain until their existing TTL expires, but upgraded readers only accept matching provenance.

[Optional harness coverage](https://github.com/PostHog/posthog-sdk-test-harness/pull/53) is separate. This PR does not opt an SDK adapter into that coverage.

Using the existing cached virtualenv with `PYTHONDONTWRITEBYTECODE=1` and `PYTHONPATH=.`:

- 1,036 focused pytest cases passed across versioned matching, publication concurrency, definition providers, single/full/bulk APIs, async feature flags, fork handling, Client behavior, and memory/Redis cache serialization. These runs emitted 127 deprecation warnings.
- The independent Redis restart reproducer failed both ordinary-definition and version-only cases on `ef127fd`. Permanent regression tests failed 24 cases on that head and pass with the fix. They cover restart, different worker snapshots, synchronous/asynchronous providers, cohort/group-mapping-only changes, paused writes, old entries without metadata, and delayed remote responses across hydration or reset.
- Committed-branch autoreview found a further remote-only restart case on `806f735`. Three regression cases failed before replacing shared remote-only provenance with per-Client identities. Same-Client fallback, matching-definition reuse, and fork controls pass.
- Changed-file Ruff formatting and lint checks, public API snapshot verification, warning-as-error import, and whitespace checks passed. Mypy with baseline filtering passed for 230 source files.
- Both new commits are signed. Commit `3dbe088` passed isolated committed-branch autoreview against `origin/main` with no actionable findings.

### Targeted mutation CI repair

The targeted job stopped before mutation testing because its CRAP gate selects only `test_utils.py` and `test_size_limited_dict.py`. Cache publication tests in other files did not cover the new snapshot paths in that selection. The CRAP calculator also counts non-executable lines, and the Redis read method's complexity of 10 could not satisfy the strict below-10 limit.

- Added 14 selected regression cases for generation fences, reused-user pruning, snapshot serialization, cross-worker reuse, wrong or missing fingerprints, disabled snapshots, TTL boundaries, and refresh during serialization. Added assertions that standalone entries have no snapshot metadata.
- Removed repeated conditions in three cache methods. Memory cache initialization and eviction share one new-user check. Redis reads choose snapshot or standalone validation with an `if/else`. Redis writes select the fingerprint once. Publication fences, write locks, Redis I/O ordering, and public signatures are unchanged.
- On Python 3.11.11, all 100 selected tests passed with 100% executable line coverage of `utils.py`. The maximum CRAP score is 8.66, below the unchanged limit of 10. A fresh actual mutation run killed all 535 mutants, with empty `mutmut results` and no survivors, timeouts, or untested mutants.
- All 519 focused cache, snapshot, fork, definition-provider, and versioned-matching tests passed, with 127 deprecation warnings. Repository-wide Ruff format/lint, baseline-filtered Mypy for 230 source files, the public API snapshot check, warning-as-error import, and whitespace checks passed.
- Signed commit `a5a6dfb` passed isolated committed-branch autoreview against `origin/main` with no actionable findings. All GitHub checks for `a5a6dfb8e96c4b5cb8ad4e857ab2a18da3e6209e` completed successfully, with no failed, pending, or skipped checks. The [CI run](https://github.com/PostHog/posthog-python/actions/runs/33981798586) includes Python 3.10 through 3.14 tests. Its [targeted mutation job](https://github.com/PostHog/posthog-python/actions/runs/33981798586/job/101348213413) executed mutation testing rather than skipping on a cache hit and killed all 535 mutants. Both capture protocols also passed [SDK compliance](https://github.com/PostHog/posthog-python/actions/runs/33981798982).

Full repository tests and generated mirror packaging were not run locally for this repair. Redis regressions use the repository's deterministic FakeRedis, not a live Redis service.

## :pencil: Checklist

### If releasing new changes

The existing `.sampo/changesets/versioned-property-matching.md` patch changeset was updated. `sampo add` was not rerun and no duplicate changeset was added.

## 🤖 Agent context

Pi agents implemented and tested this user-directed SDK contract update using repository inspection, Python/pytest, Ruff, Git, GitHub CLI, and the isolated autoreview helper. Human review is required before merging. No public agent session link is available.

Fresh review found that a process-local generation could allow invalidated Redis entries to revive after restart. The repair adds private snapshot provenance without putting Redis I/O under the publication lock. The user-directed review workflow approved the additive metadata and stricter handling of unverifiable remote-only entries. Legacy matching defaults, public cache APIs, and Redis key formats are preserved.

### DIFF
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -2253,6 +2253,9 @@ class Client(object):
         # If using Memory cache, we keep it as-is to benefit from the inherited warm cache.
         if isinstance(self.flag_cache, RedisFlagCache):
             self.flag_cache = self._initialize_flag_cache(self.flag_fallback_cache_url)
+        if self.flag_cache:
+            self.flag_cache._write_lock = threading.Lock()
+            self.flag_cache._advance_generation(self.flag_definition_version)
 
         reset_sessions()
 
@@ -2896,9 +2899,10 @@ class Client(object):
                     and old_flags_by_key != (self.feature_flags_by_key or {})
                 )
             ):
-                old_version = self.flag_definition_version
                 self.flag_definition_version += 1
-                self.flag_cache.invalidate_version(old_version)
+                # Only advance an in-memory fence while publishing. Redis I/O
+                # must not delay snapshots, and late writes remain invalid.
+                self.flag_cache._advance_generation(self.flag_definition_version)
 
     def _local_evaluation_snapshot(self) -> _LocalEvaluationSnapshot:
         # Capture references together. Publication replaces these collections, so
@@ -3078,7 +3082,10 @@ class Client(object):
                     self._flag_definition_cache_generation = fetch_generation
 
                     if self.flag_cache:
-                        self.flag_cache.clear()
+                        self.flag_definition_version += 1
+                        self.flag_cache._advance_generation(
+                            self.flag_definition_version
+                        )
 
                     if self.debug:
                         raise APIError(status=401, message=detail)
@@ -3095,9 +3102,12 @@ class Client(object):
                     self._flag_definition_published_generation = fetch_generation
                     self._flag_definition_cache_generation = fetch_generation
 
-                    # Clear flag cache when quota limited
+                    # Invalidate results without waiting for external cache I/O.
                     if self.flag_cache:
-                        self.flag_cache.clear()
+                        self.flag_definition_version += 1
+                        self.flag_cache._advance_generation(
+                            self.flag_definition_version
+                        )
 
                     if self.debug:
                         raise APIError(
@@ -3403,15 +3413,12 @@ class Client(object):
                 cached_flag_result = FeatureFlagResult.from_value_and_payload(
                     key, flag_value, self._compute_payload_locally(key, flag_value)
                 )
-            with self._flag_definition_publication_lock:
-                if (
-                    self.flag_cache
-                    and cached_flag_result
-                    and local_definition_version == self.flag_definition_version
-                ):
-                    self.flag_cache.set_cached_flag(
-                        distinct_id, key, cached_flag_result, local_definition_version
-                    )
+            if self.flag_cache and cached_flag_result:
+                # The cache rejects invalidated generations, including writes
+                # already in flight when new definitions are published.
+                self.flag_cache.set_cached_flag(
+                    distinct_id, key, cached_flag_result, local_definition_version
+                )
         elif only_evaluate_locally:
             if self.feature_flags is None:
                 self.log.warning(
--- a/posthog/client.py
+++ b/posthog/client.py
@@ -130,6 +130,11 @@ _atexit_deadline: Optional[float] = None
 _atexit_deadline_lock = threading.Lock()
 
 
+class _LocalEvaluationSnapshot(FlagDefinitionCacheData):
+    flags_by_key: Dict[str, Any]
+    flag_definition_version: int
+
+
 def _supports_lane_synchronization(queue) -> bool:
     return all(
         hasattr(queue, attribute)
@@ -879,6 +884,7 @@ class Client(object):
         self.feature_flags_by_key: Optional[dict[str, Any]] = None
         self.group_type_mapping: Optional[dict[str, str]] = None
         self.cohorts: Optional[dict[str, Any]] = None
+        self._property_matching_version = 1
         self.poll_interval = poll_interval
         self.feature_flags_request_timeout_seconds = (
             feature_flags_request_timeout_seconds
@@ -895,7 +901,7 @@ class Client(object):
         self._flag_definition_fetch_generation = 0
         self._flag_definition_published_generation = 0
         self._flag_definition_cache_generation = 0
-        self._flag_definition_publication_lock = threading.Lock()
+        self._flag_definition_publication_lock = threading.RLock()
         self._flag_definition_cache_write_lock = threading.RLock()
         self._flag_definition_cache_provider = flag_definition_cache_provider
         self._flag_definition_cache_provider_async_runner: Optional[
@@ -2234,7 +2240,7 @@ class Client(object):
 
         # A parent thread may have been publishing or caching flag definitions at
         # fork time.
-        self._flag_definition_publication_lock = threading.Lock()
+        self._flag_definition_publication_lock = threading.RLock()
         self._flag_definition_cache_write_lock = threading.RLock()
 
         # Metrics locks may have been held by a parent thread at fork time; replace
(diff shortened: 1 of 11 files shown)


posthog/client.py
```diff
@@ -4337,6 +4434,7 @@ def _get_all_flags_and_payloads_locally(
                         group_properties=group_properties,
                         warn_on_unknown_groups=warn_on_unknown_groups,
                         device_id=device_id,
+                        _definition_snapshot=snapshot,
```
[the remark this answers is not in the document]
marandaneto: Fixed in d2cde4673a1bb2e8f001d4a57f9b5b72dd0869ff. Local value and payload lookups now share the captured definition snapshot, including payload overrides and the actual-result cache write; bulk APIs pass their existing snapshot through. Cache generation fences, lazy loading, remote fallback and public API signatures are unchanged.

Reproduced deterministically by replacing/removing definitions immediately after matching: `uv run --no-sync pytest posthog/test/test_property_matching_version.py -k in_flight_payload --timeout=30 -q` failed all 12 cases before the fix (replacement/missing payload), and passes all 12 afterward. Retained regression coverage for single results, payload lookups including boolean/variant overrides, legacy bulk results, evaluate_flags, and rejection of stale generation cache writes. Surrounding flags/client/cache suites: 1,070 passed plus 29 subtests; Ruff format/lint, baseline-filtered Mypy, public API snapshot and warning-as-error import passed locally.

posthog/client.py
```diff
@@ -891,11 +898,17 @@ def __init__(
         self.flag_fallback_cache_url = flag_fallback_cache_url
         self.flag_cache = self._initialize_flag_cache(flag_fallback_cache_url)
         self.flag_definition_version = 0
+        # Without definitions, remote results have only process-local provenance.
+        self._flag_definition_fingerprint = f"remote-only:{uuid4().hex}"
```
turnipdabeets, line 902: A worker that starts while `/flags` and definitions are both down gets a `remote-only:` fingerprint, so it can't read Redis entries other workers wrote — that includes workers with a secret key, not just remote-only clients. Main serves those entries (confirmed with a two-worker FakeRedis test: passes on main, fails here), and the `_initialize_flag_cache` docstring promises this setup works during an outage. I think we need to keep cross-process fallback for remote results, or move the provenance rework into its own PR so it can be weighed separately from the matching fix.

posthog/client.py
```diff
@@ -2868,28 +2891,63 @@ def _shutdown_flag_definition_cache_provider(self):
...
+            separators=(",", ":"),
+            ensure_ascii=True,
+        )
+        return _hashlib.sha256(serialized.encode("utf-8")).hexdigest()
+
     def _update_flag_state(
-        self, data: FlagDefinitionCacheData, old_flags_by_key: Optional[dict] = None
+        self,
+        data: FlagDefinitionCacheData,
+        old_flags_by_key: Optional[dict] = None,
```
turnipdabeets: [nit] We no longer read `old_flags_by_key` in `_update_flag_state`. Should we drop it along with the two call sites that build it?

posthog/client.py
marandaneto: Fixed in dd488a5dccb681bf840bddc30ac533ad4a367aca. Redis entries now explicitly identify remote evaluations: a worker without definitions can reuse those entries during an outage, including with a secret key. Locally evaluated entries still require matching snapshots; workers with loaded definitions still reject differing snapshots. Generation fences, 401/402 reset denial, stale TTL and the prior value/payload snapshot fix remain intact.

Reproduced using two clients sharing FakeRedis with both endpoints unavailable: `uv run --no-sync pytest posthog/test/test_redis_flag_cache_snapshot.py -k "cold_worker or loaded_worker or reset_worker" --timeout=30 -q` failed all four cold-start combinations before the fix (with/without secret key, writer with/without definitions); now 8 passed including negative controls. The full Redis snapshot suite passes 40 tests, including remote payload preservation, expiry and rejected late responses. Full repository suite: 3,196 passed, 16 skipped, 38 subtests passed; Ruff, Mypy, API snapshot and warning-as-error import checks passed.

posthog/client.py
marandaneto: Removed in dd488a5dccb681bf840bddc30ac533ad4a367aca: the unused private parameter, both production call-site constructions, and the obsolete positional argument in four regression-test call sites. `rg old_flags_by_key posthog` finds no remaining references. Full suite (3,196 passed plus 38 subtests), Ruff, Mypy and public API snapshot checks pass.

posthog/client.py
```diff
@@ -2868,28 +2891,62 @@ def _shutdown_flag_definition_cache_provider(self):
...
-            old_version = self.flag_definition_version
-            self.flag_definition_version += 1
-            self.flag_cache.invalidate_version(old_version)
+            if fingerprint != self._flag_definition_fingerprint:
+                self._flag_definition_fingerprint = fingerprint
+                if self.flag_cache:
+                    self.flag_definition_version += 1
+                    # No Redis I/O under this lock. The fingerprint remains valid
+                    # across processes; the counter fences this process's writes.
+                    self.flag_cache._advance_generation(
```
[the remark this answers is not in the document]
marandaneto: Addressed in bd6b719ca6153779c1184ba583d98bba1349569f. After a loaded-definition change (API or cache provider), or a 401/402 reset, we now best-effort scan entries belonging to the old fingerprint outside the publication lock. A single-key Lua compare-and-delete removes only the exact observed value, preserving concurrent replacements. Initial hydration still preserves shared remote outage fallback; snapshot/generation fences remain unchanged.

Reproduced first with `uv run --no-sync pytest posthog/test/test_redis_flag_cache_snapshot.py -k "legacy_reader or snapshot_cleanup" --timeout=30 -q`: 9 failures before, 9 passes after. Retained tests cover legacy-reader misses after invalidation/reset, concurrent replacement, Redis errors, nonblocking publication and debug-reset cleanup. Verified the Lua script with redis-py/fakeredis[lua] for both byte and decoded responses. Full suite: 3,224 passed plus 38 subtests; mutation and CRAP gates, lint, types, API snapshot, and exact-head CI/compliance are green: https://github.com/PostHog/posthog-python/actions/runs/35427512099

This is deliberately best effort, not an authorization/revocation guarantee: unavailable Redis or denied EVAL cannot be cleaned, later writes by stale workers can reappear, and untagged entries cannot safely be attributed to the old snapshot. No unsafe GET/DELETE or broad namespace deletion is used.

turnipdabeets (approved):

PR #923, opened 2026-09-07
feat(mcp): redact structured PII from $mcp_intent + stronger prompt

gesh:
Ports [PostHog/posthog-js#4783](https://github.com/PostHog/posthog-js/pull/4783) to the Python SDK.

`$mcp_intent` is agent-narrated free text written into the injected `context` argument, so a model can narrate personal data it read aloud. Two defensive layers:

- **Always-on redaction** of structured identifiers — emails, IPv4/IPv6, US SSNs, Luhn-valid card numbers, and phone numbers — scoped to the intent only. Structured tool `arguments`/responses are left untouched, since the same shapes are often legitimate there. Best-effort for well-defined shapes, not free-form names/addresses. Email quantifiers are bounded to avoid quadratic backtracking; every card in a multi-card span is redacted.
- **Stronger default `context` prompt** telling the agent not to repeat/paraphrase/infer personal data and to refer to people by role.

`redact_pii` lives in `_sanitization.py` and runs on `user_intent` after the existing token/secret scrub — a single choke point covering every adapter and the manual `PostHogMCP` path.

## Tests
New parametrized redaction + intent-scoping tests in `posthog/test/mcp/test_pipeline.py`, mirroring the JS suite (per-identifier cases, multi-identifier, multi-card span, anti-quadratic timing guard, and `sanitize_event` scoping). Full MCP suite passes on both SDK v1 and v2 (Python 3.10 and 3.14).

---

### DIFF
--- a/posthog/mcp/_sanitization.py
+++ b/posthog/mcp/_sanitization.py
@@ -192,7 +192,7 @@ def _redact_card_in_match(match: re.Match[str]) -> str:
     return output + text[cursor:]
 
 
-def redact_pii(value: str) -> str:
+def redact_pii(value: Any) -> Any:
     """Redact structured personal identifiers (emails, IP addresses, credit-card
     numbers, US SSNs, and phone numbers) from a free-text string. Intended for the
     agent-narrated $mcp_intent value only — not for structured tool parameters or
--- a/posthog/mcp/_sanitization.py
+++ b/posthog/mcp/_sanitization.py
@@ -27,6 +27,75 @@ _SENSITIVE_KEY_PATTERN = re.compile(
     re.IGNORECASE,
 )
 
+# PII redaction for the agent-narrated intent string only. $mcp_intent is free
+# text the calling LLM writes into the injected `context` argument, so it can
+# carry personal data the model read aloud despite being told not to. We redact
+# well-defined *structured identifiers* — the kind regex can match with high
+# precision. Person names and postal addresses are deliberately out of scope:
+# they need an NER model that a client SDK cannot ship, and naive patterns would
+# over-redact ordinary prose. Patterns are ordered so an earlier pass never eats
+# digits a later pass needs (email before phone, IPs before phone, cards before
+# the generic phone pass). See `redact_pii`.
+#
+# The `\d`/`\w`-based patterns are compiled with re.ASCII to match the JS
+# semantics they are ported from: JS `\d`/`\w`/`\b` are ASCII-only, whereas
+# Python's default is Unicode and would over-match (e.g. Unicode digits).
+#
+# Horizontal Unicode spaces (NBSP, narrow NBSP, ideographic space, ...) are what
+# appear when text is copied from web pages or PDFs. `redact_pii` normalizes them
+# to an ASCII space first so the separator-based card/phone/SSN candidates match
+# them instead of leaking the identifier they group.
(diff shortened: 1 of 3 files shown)


gesh (commented):
QA Swarm review complete. See inline comments.

posthog/mcp/_sanitization.py
```diff
@@ -106,9 +244,15 @@ def sanitize_event(event: Dict[str, Any]) -> Dict[str, Any]:
...
     # The intent comes straight from an agent-narrated `context` string, so it
-    # can contain a secret the LLM read aloud. Redact it like any other value.
+    # can contain a secret the LLM read aloud or personal data it narrated about
+    # the user. Redact it like any other captured value, then strip structured
+    # PII (emails, phone numbers, IPs, cards, SSNs) rather than shipping it raw
+    # as $mcp_intent. PII redaction is scoped to the intent only — structured
+    # tool parameters and responses often hold the same shapes as legitimate data.
     if result.get("user_intent") is not None:
-        result["user_intent"] = sanitize_captured_value(result["user_intent"])
+        result["user_intent"] = redact_pii(
```
gesh, line 257: > [!NOTE]

**[router]** 🟢 LOW

`sanitize_event` calls `redact_pii(sanitize_captured_value(result["user_intent"]))` without checking the value is a string. `user_intent` can be non-string via the custom-event API (`_capture.py:64`, `event_input.get("user_intent")`; `Event = Dict[str, Any]`). Before this PR, `sanitize_captured_value` returned non-str/list/dict values unchanged, so a non-string `user_intent` passed through; now `redact_pii` calls `.sub()` on it and raises `TypeError`.

Verified: `sanitize_event({'user_intent': 123})` raises `TypeError`; previously it returned the event unchanged. The pipeline's `_sink.py` try/except catches it (event dropped + logged — fails closed, no PII leak or crash), so this is a behavior regression rather than a security hole.

Suggest an `isinstance(value, str)` guard at the top of `redact_pii` (return the value unchanged otherwise) to restore parity with the prior non-string handling.

gesh:
> [!NOTE]
>
> Multi-perspective review: router (cheap-first pass) + delegated reviewers as warranted

## Verdict: 💬 APPROVE WITH NITS <sub>(round 1 @ 9c4ec97)</sub>

A well-scoped, well-tested port of posthog-js#4783. The router independently fuzzed every redaction pattern for ReDoS (linear up to ~250k chars), confirmed the Luhn card scan / IPv6 / phone boundary logic, and verified redaction is wired only onto `user_intent` (not `parameters`/`response`). One low-severity robustness nit.

### Key findings

- 🟢 **LOW** — `redact_pii` assumes a `str` argument; a non-string `user_intent` (reachable via the custom-event API, `_capture.py:64`) now raises `TypeError` where it previously passed through. Fails closed via `_sink.py`'s try/except (event dropped + logged), so no leak/crash — a behavior regression, not a security hole. Suggest an `isinstance(value, str)` guard in `redact_pii`.

### Convergence

None (single router pass; no delegation warranted for a LOW-danger, well-tested diff).

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🧭 router | MEDIUM danger / HIGH confidence. Fuzzed all patterns for ReDoS, verified redaction scope and Luhn/IPv6/phone correctness, ran the sync test suite (pass). Delegated nothing — cheap pass sufficient. |

---

lucasheriques (approved):

PR #927, opened 2026-09-07
feat(mcp): capture model identifiers

lucasheriques:

Python MCP server owners can compare tool behavior by calling model, matching [posthog-js#4829](https://github.com/PostHog/posthog-js/pull/4829).

- Model capture is opt-in and prefers recognized Codex metadata over an injected self-report field.
- `$mcp_llm_model_source` distinguishes `client_metadata` from `self_reported` values.
- Application-owned fields and complex schemas remain untouched by model injection.
- Custom dispatchers preserve model-field ownership across repeated listings by copying object tools before injection.
- Analytics schema injection preserves existing restrictions on undeclared arguments.
- Raw low-level and standalone FastMCP adapters keep the model field optional for validation compatibility.
- Both model sources remain unverified and suit product analytics, not billing or access control.

Public API snapshot updates and the shared injection predicate are mechanical.

- Regression coverage checks default-off and explicit opt-out behavior through prepared calls and captured events.
- Object-tool cases cover repeated listings with and without an explicit original tool across both MCP SDK majors.
- Schema coverage checks accepted analytics fields and rejected undeclared arguments, including after a low-level tool-cache rebuild.
- Existing wire coverage checks metadata precedence, argument stripping, capture properties and tool execution.
- The full non-MCP suite was not rerun locally for the review fixes; CI covers the broader matrix.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Codex authored the change and review fixes with terminal and GitHub CLI tools. Skills: debugging MCP analytics, writing tests, writing dataclasses, and writing PR descriptions.

Model resolution respects opt-in on custom dispatchers. Optional fields remain optional where stripping them before validation would otherwise reject calls.


posthog/mcp/posthog_mcp.py
```diff
@@ -285,10 +314,26 @@ def prepare_tool_call(
...
         )
+        analytics_owns_model = False
+        if is_capture_model_enabled(self._capture_model):
+            if original_tool is not None:
+                analytics_owns_model = can_inject_model_parameter(
+                    _tool_schema(original_tool)
+                )
+            else:
+                analytics_owns_model = self._model_parameter_injected.get(name, False)
+        llm_model, llm_model_source = resolve_model(
```
gesh: > [!NOTE]

**[convergent: router + validation]** 🟠 HIGH

`resolve_model()` is called outside the `capture_model` gate. Line 318 gates only `analytics_owns_model`, and `resolve_model()` checks Codex `x-codex-turn-metadata` before it consults `allow_self_reported` (`_model_parameters.py:123-127`). So the client-metadata branch runs even when `capture_model` is left at its default of `False`.

Reproduced at this head: a `PostHogMCP` built without `capture_model` and called as `prepare_tool_call("search", {"q": "hello"}, request_meta={"x-codex-turn-metadata": {"model": "..."}})` returns a populated `llm_model` with source `client_metadata`, where both should be `None`.

There is no downstream gate to catch it. `_apply_model()` writes the fields onto the event whenever they are non-empty, and `capture_tool_call()` / `capture_missing_capability()` take them as plain parameters without re-checking the option. A host that plumbs `PreparedToolCall` through therefore emits the model identifier with no opt-in anywhere in the chain.

This is the documented custom-dispatcher path, and it contradicts the PR's own "off by default" claim. The `instrument()` adapters get this right — both `resolve_model()` call sites in `_instrumentation.py` sit inside `if is_capture_model_enabled(...)`.

Fix: gate the call itself, matching the adapters.

```python
llm_model, llm_model_source = None, None
if is_capture_model_enabled(self._capture_model):
    llm_model, llm_model_source = resolve_model(
        request_meta, args, allow_self_reported=analytics_owns_model
    )
```

Worth a test too: every `capture_model` test in the suite constructs the client with `capture_model=True`, so the default-off case is unpinned.

gesh (commented):
> [!NOTE]

QA Swarm review complete. See inline comments.

posthog/mcp/posthog_mcp.py
```diff
@@ -356,6 +401,45 @@ def _inject_context(self, tool: Any, description: Optional[str]) -> Any:
             pass
         return tool
 
+    def _inject_models(self, tools: List[Any]) -> List[Any]:
```
gesh, line 410: > [!NOTE]

**[router]** 🟢 LOW

`_inject_models()` clears `self._model_parameter_injected` and then rebuilds it, while `prepare_tool_call()` reads the same dict at line 324. Between the `clear()` and the `update()`, a reader sees an empty map, concludes the SDK does not own `llm_model` for that tool, and skips stripping it from `prepared_args` — so the injected argument reaches the real tool.

Reachability is limited: there is no `await` between the two statements, so asyncio alone cannot interleave here. It needs genuine OS threads sharing one `PostHogMCP` and re-listing tools while a call is in flight. That is a narrow window, but the class is explicitly built for reuse across concurrent requests, and `context` never had a comparable shared ownership cache.

Cheapest fix is to swap rather than mutate: build the new dict locally and assign it in one statement, so a concurrent reader sees either the old map or the new one, never an empty one. If you would rather not spend anything here, a comment stating that `prepare_tool_list()` must not run concurrently with `prepare_tool_call()` would at least make the assumption explicit.

posthog/mcp/_model_parameters.py
```diff
@@ -0,0 +1,142 @@
...
+        required_list = schema.get("required")
+        if isinstance(required_list, list):
+            if "llm_model" not in required_list:
+                required_list.append("llm_model")
+        else:
+            schema["required"] = ["llm_model"]
+    return schema
+
+
+def can_inject_model_parameter(input_schema: Any) -> bool:
```
gesh, line 71: > [!NOTE]

**[router]** 🟢 LOW

`can_inject_model_parameter()` re-derives the same three conditions that `add_model_parameter_to_schema()` already implements at lines 32-49: does the schema own `llm_model`, and is it a `$ref` / `oneOf` / `allOf` / `anyOf` schema. The two are kept in sync by hand.

Since those conditions are exactly what makes the feature fail closed, an edit to one copy that misses the other reintroduces the ownership bug this design exists to prevent. Consider extracting one predicate and having `add_model_parameter_to_schema()` call it.

posthog/mcp/posthog_mcp.py
```diff
@@ -260,22 +279,32 @@ def prepare_tool_list(
...
     def prepare_tool_call(
-        self, name: str, args: Optional[JsonRecord] = None
+        self,
+        name: str,
+        args: Optional[JsonRecord] = None,
+        *,
+        request_meta: Optional[JsonRecord] = None,
+        original_tool: Any = None,
     ) -> PreparedToolCall:
         """Pull the agent's intent off the injected ``context`` argument, strip
```
gesh, line 311: > [!NOTE]

**[router]** ⚪ NIT

The docstring still describes only the `context` behavior. This method now also resolves the model, records the source, and strips `llm_model` from the arguments — worth a line, since this is the documented entry point for the new feature on the custom-dispatcher path.

posthog/mcp/_model_parameters.py
```diff
@@ -0,0 +1,142 @@
...
+        return schema
+
+    if not schema:
+        schema = {"type": "object", "properties": {}, "required": []}
+
+    schema = copy.deepcopy(schema)
+    if not isinstance(schema.get("properties"), dict):
+        schema["properties"] = {}
+    if schema.get("additionalProperties") is False:
+        schema.pop("additionalProperties", None)
```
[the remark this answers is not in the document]
lucasheriques: Fixed in [11c48e9](https://github.com/PostHog/posthog-python/commit/11c48e9dbf850bd881f0a4125e4a50270a2464c4). Model injection now preserves `additionalProperties: false`. Context and conversation-ID injection also preserve it, so later stages cannot relax the constraint. Regression coverage verifies that declared analytics fields remain valid and undeclared arguments are rejected, including after a low-level tool-cache rebuild.

posthog/mcp/posthog_mcp.py
```diff
@@ -285,10 +314,26 @@ def prepare_tool_call(
             if isinstance(raw_context, str) and raw_context.strip()
             else None
         )
+        analytics_owns_model = False
+        if is_capture_model_enabled(self._capture_model):
+            if original_tool is not None:
+                analytics_owns_model = can_inject_model_parameter(
+                    _tool_schema(original_tool)
+                )
```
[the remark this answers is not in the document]
lucasheriques: Fixed in [11c48e9](https://github.com/PostHog/posthog-python/commit/11c48e9dbf850bd881f0a4125e4a50270a2464c4). Model injection now copies object tools before changing their schemas, leaving the original application tool available for ownership checks. Regression cases cover repeated listings with and without `original_tool`, using SDK tool objects and custom descriptors across both MCP SDK majors. Self-reported values are captured and stripped before dispatch.

posthog/mcp/_instrumentation.py
```diff
@@ -630,19 +668,38 @@ def mutate_tool_schema(
...
+        model_was_injected = data.tool_model_parameter_injected.get(tool.name, False)
+        app_owns_model = (
+            schema_has_param(schema, "llm_model") and not model_was_injected
+        )
+        if not app_owns_model and not schema_has_param(schema, "llm_model"):
+            schema = add_model_parameter_to_schema(
+                schema,
+                tool.name,
+                get_model_description(data.options.capture_model),
+                required=context_required,
```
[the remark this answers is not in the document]
lucasheriques: Addressed through the documented-contract option in [11c48e9](https://github.com/PostHog/posthog-python/commit/11c48e9dbf850bd881f0a4125e4a50270a2464c4). The README now distinguishes required fields on custom dispatchers and official high-level adapters from optional fields on raw low-level servers and standalone FastMCP. The standalone adapter strips injected fields before input validation, so making `llm_model` required there would reject calls after stripping. Requiredness is unchanged; the low-level regression also verifies that a call omitting the field succeeds.

posthog/mcp/posthog_mcp.py
lucasheriques: Confirmed and fixed in [11c48e9](https://github.com/PostHog/posthog-python/commit/11c48e9dbf850bd881f0a4125e4a50270a2464c4). The `resolve_model()` call itself now sits inside the `capture_model` gate, with both prepared fields initialized to `None`. The capture test now covers enabled, explicit `False`, and default settings with Codex metadata present, checking both prepared values and emitted event properties.

posthog/mcp/posthog_mcp.py
lucasheriques: Fixed in [11c48e9](https://github.com/PostHog/posthog-python/commit/11c48e9dbf850bd881f0a4125e4a50270a2464c4). Ownership is now built in a local dictionary, including injection failures, and assigned to the instance only after preparation completes. The existing map remains available during rebuilding; the `clear()`/`update()` window is gone.

posthog/mcp/_model_parameters.py
lucasheriques: Updated in [11c48e9](https://github.com/PostHog/posthog-python/commit/11c48e9dbf850bd881f0a4125e4a50270a2464c4). `add_model_parameter_to_schema()` now calls `can_inject_model_parameter()` rather than duplicating the ownership and complex-schema checks. Existing coverage for application-owned fields and complex schemas passes.

posthog/mcp/posthog_mcp.py
lucasheriques: Updated in [11c48e9](https://github.com/PostHog/posthog-python/commit/11c48e9dbf850bd881f0a4125e4a50270a2464c4). The `prepare_tool_call()` docstring now describes opt-in model resolution, source attribution, and stripping the SDK-owned `llm_model` argument. The list helper and README also explain object copying and original-tool ownership.

posthog/mcp/_context_parameters.py
```diff
@@ -79,10 +79,7 @@ def add_context_parameter_to_schema(
     if not isinstance(schema.get("properties"), dict):
         schema["properties"] = {}
 
-    # additionalProperties: false would reject the injected context — remove it
-    # (the SDK adds this when converting Pydantic models to JSON Schema).
-    if schema.get("additionalProperties") is False:
-        schema.pop("additionalProperties", None)
+    # The declared context property is allowed even under additionalProperties: false.
```
gesh, line 82: > [!NOTE]

**[router]** 🟡 MEDIUM

Removing the `additionalProperties: false` stripping here is correct, and I verified it rather than assuming: a property added to `properties` stays valid under `additionalProperties: false`, and the new `Draft202012Validator` test plus the extended low-level cache-rebuild test both pin it. The SDK should not be silently weakening a schema constraint the tool author declared.

The concern is scope, not correctness. Neither this file nor `_conversation_id.py` was touched by this PR's first commit, and `context` injection is on by default — so this reaches every `instrument()` user, not only those who opt into `capture_model`.

It is also a change to released behavior. The stripping came in with #888 and shipped in v7.43.1; the package is now on 7.47.1. So a tool author who declared `additionalProperties: false` and has had extra arguments accepted since v7.43.1 will start seeing those calls rejected once this ships.

The changeset covers it in one clause — "keeps strict input validation when adding analytics fields" — which reads as a property of the new opt-in feature rather than something every existing user gets. Worth splitting into its own commit with its own changeset entry, or at least rewording the changeset to say plainly what existing users will notice.

posthog/test/mcp/test_model_parameters.py
```diff
@@ -0,0 +1,104 @@
+from types import SimpleNamespace
+
+import pytest
+from jsonschema import Draft202012Validator
```
gesh, line 4: > [!NOTE]

**[router]** 🟢 LOW

`jsonschema` is now imported at module level here, but it is not a declared test dependency — it only resolves transitively through `mcp`, which is pinned `mcp>=1.28.1,<2` in the `test` extra.

The PR description says the suite was also run against MCP Python SDK 2.x. If that environment does not pull `jsonschema` along, this import fails and takes the whole module with it, not just the new assertion — so the failure would look much broader than its cause.

Adding `jsonschema` to the `test` extra in `pyproject.toml` makes the dependency real rather than incidental.

gesh (commented):
> [!NOTE]

QA Swarm round 2 @ `11c48e9`. All four round 1 findings verified fixed. See inline comments.

gesh (approved):
Approving to unblock, the robot comments look meaningful

posthog/mcp/_context_parameters.py
lucasheriques: Updated the changeset in [49ca1a4](https://github.com/PostHog/posthog-python/commit/49ca1a44b755debfdc9bc1d31a89342833424820). It now explicitly says that context and conversation-ID injection preserve `additionalProperties: false` even when model capture is disabled, and that servers validating those schemas will reject undeclared arguments accepted by earlier SDK versions. Declared analytics fields remain valid. Kept the schema fix in this PR and used the suggested release-note clarification. The MCP suite passes locally (267 tests).

posthog/test/mcp/test_model_parameters.py
lucasheriques: Added `jsonschema>=4.0` directly to the `test` extra and refreshed its dependency entries in `uv.lock` in [49ca1a4](https://github.com/PostHog/posthog-python/commit/49ca1a44b755debfdc9bc1d31a89342833424820). The tests no longer rely on MCP installing it transitively. The lockfile check and all 267 MCP tests pass locally with MCP SDK 1.29.0; MCP v2 will be checked by the PR's CI matrix.

PR #928, opened 2026-09-07
feat(mcp): capture resource discovery and reads

lucasheriques:

MCP server authors cannot tell whether clients discover or read their resources.

This leaves resource-based integrations absent from MCP Analytics even when tool tracking works.

Resource bodies stay out of analytics because resources may contain private documents or credentials.

## :hammer: Changes

- `resources/list`, `resources/templates/list` and `resources/read` are instrumented on every adapter (low-level v1, FastMCP, jlowin FastMCP, v2). Listings emit `$mcp_resources_list` with the listing metadata as `$mcp_response`; reads emit `$mcp_resource_read` with the URI, duration and error state and never capture the body.
- Captured URLs redact credentials: userinfo, credential-named query and fragment fields (matched per `-`/`_`/`.`/`/`/`;` segment, plus exact signed-URL names), URLs nested one level inside a retained value, authority-less URIs such as `resource:guide?token=x`, hash-routed fragments, and adjacent addresses run together without whitespace. Whenever a credential's tail is ambiguous the sanitizer fails closed. **This applies to every captured string, so URLs already flowing through `$mcp_tool_call` data will show `%5Bredacted%5D` values after upgrading.**
- Credential and PII redaction run before URL rewriting so percent-encoding cannot hide a token or an email; the narrated intent is gated for binary blobs first. Work is linear in the input: no per-address backward scans, no recursion per fragment or address.
- `$identify` from a resource read is named by its (redacted) URI, and `resource_name` is sanitized with the URL passes only so tool names are never eaten by the entropy detector.
- On MCP SDK v2 a failing read reports what the handler actually raised instead of the SDK's `MCPError` wrapper.

The same behavior ships in PostHog/posthog-js#4830; both sanitizers produce byte-identical output on 78 shared vectors.

- The MCP 1.x suite verifies resource discovery, reads, and preserved responses.
- The MCP 2.x suite verifies the same behavior under the 2026-07-28 protocol.
- The late-registration test verifies dynamically added resource handlers receive instrumentation.
- Mypy passes against the repository baseline.
- 78 URL vectors are asserted in `test_pipeline.py`, including three timed pathological inputs; a cross-model review pass ran to completion with no findings on the final branch.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Codex implemented this change with the `debugging-mcp-analytics`, `writing-tests`, and `writing-pr-descriptions` skills. The paired JavaScript implementation uses the same event contract to prevent SDK drift.


posthog/mcp/_instrumentation.py
```diff
@@ -834,3 +833,45 @@ async def record_tools_list(
...
+        params = request.get("params")
+        uri = params.get("uri") if isinstance(params, dict) else None
+        event: Dict[str, Any] = {
+            "event_type": event_type,
+            "session_id": session_id,
+            "resource_name": uri
+            if event_type == MCPAnalyticsEventType.MCP_RESOURCES_READ
+            else None,
+            "parameters": build_captured_mcp_parameters(request),
+            "response": _wrap_response(response) if response is not None else None,
```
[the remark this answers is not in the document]
lucasheriques: Fixed in a341587. Resource events now omit responses entirely, and both MCP-major tests assert that `$mcp_response` is absent.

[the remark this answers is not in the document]
lucasheriques: Fixed in a341587: resource bodies are no longer included in analytics. We also fixed credential redaction for resource addresses in 62caaf1. Tests cover successful and failed reads and check that the token is absent from every captured event.

posthog/mcp/_sanitization.py
```diff
@@ -27,6 +30,44 @@
...
+            else (key, item)
+            for key, item in query
+        ]
+        netloc = url.netloc
+        if "@" in netloc:
+            netloc = "%5Bredacted%5D@" + netloc.rsplit("@", 1)[1]
+        if netloc == url.netloc and sanitized_query == query:
+            return value
+        return urlunsplit(
+            (url.scheme, netloc, url.path, urlencode(sanitized_query), url.fragment)
```
gesh: **URL fragments are never redacted.** `_sanitize_url` examines only `netloc` and `query`. `url.fragment` goes to `urlunsplit` unchanged, and the unchanged-URL early return also emits it verbatim. A URI such as `https://app.example.com/cb#access_token=eyJ...` ships the token to PostHog in `$mcp_resource_name`, `$mcp_parameters`, and error messages. The same token in a query string is redacted. This is the RFC 6749 implicit-grant shape, so it is a standard pattern. The entropy fallback does not catch it, because `_looks_like_path_or_url` bails on strings that contain `://`. Suggestion: apply the sensitive-key redaction to fragments that parse as `key=value` pairs, and add a test with a credential-bearing fragment (the only fragment test, `test_pipeline.py`, uses a benign fragment).

posthog/mcp/_sanitization.py
```diff
@@ -27,6 +30,44 @@
...
+)
+
+
+def _sanitize_url(match: re.Match[str]) -> str:
+    if match.end() - match.start() > _MAX_URL_LENGTH:
+        return _REDACTED_VALUE
+    value = match.group(0)
+    try:
+        url = urlsplit(value)
+        query = parse_qsl(
```
gesh: **Semicolon-delimited query credentials bypass redaction.** `parse_qsl` splits on `&` only (Python >= 3.10). For `https://example.com/x?a=1;token=sekret123`, the whole query becomes one pair `('a', '1;token=sekret123')`. No key matches the sensitive patterns, so the secret ships unredacted. Verified by execution on this branch. Real-world likelihood is low (legacy convention), but it contradicts the README claim. Suggestion: also split on `;`, or redact URLs whose query contains `;`.

posthog/mcp/_sanitization.py
```diff
@@ -27,6 +30,44 @@
     re.IGNORECASE,
 )
 
+_URL_PATTERN = re.compile(r"\b[a-z][a-z0-9+.-]{0,63}://[^\s<>\"']+", re.IGNORECASE)
+_MAX_URL_LENGTH = 8192
+_MAX_URL_QUERY_FIELDS = 128
+_SENSITIVE_QUERY_KEY_PATTERN = re.compile(
```
gesh, line 79: **Some standard credential query keys are missing from the redaction list.** `code` (the OAuth authorization code), `jwt`, and `sessionid` pass through verbatim, while `token` and `sig` are redacted. Verified by execution: `?jwt=...&sessionid=abc123&code=authcode123` is captured unchanged (short values dodge the entropy fallback). `code` in a callback URL is as standardized as `access_token`. Suggestion: add `code`, `jwt`, and `session`/`sessionid` to `_SENSITIVE_QUERY_KEY_PATTERN`.

posthog/mcp/_sanitization.py
```diff
@@ -27,6 +30,44 @@
     re.IGNORECASE,
 )
 
+_URL_PATTERN = re.compile(r"\b[a-z][a-z0-9+.-]{0,63}://[^\s<>\"']+", re.IGNORECASE)
```
gesh: **The URL regex absorbs trailing prose punctuation into the rewrite.** The terminal class `[^\s<>"']+` matches `,`, `)`, and `.`. When redaction rewrites a URL inside a message, adjacent punctuation is parsed into the query and dropped. Verified by execution: `See https://example.com/x?sig=abc, then retry.` becomes `See https://example.com/x?sig=%5Bredacted%5D then retry.` and `Failed (https://example.com/x?sig=abc).` loses `).` entirely. Impact is only on captured analytics text, and only when a rewrite triggers, but it degrades `$mcp_error_message` fidelity. Suggestion: strip trailing punctuation from the match before parsing, and re-append it after the rewrite.

posthog/mcp/_instrument_lowlevel.py
```diff
@@ -73,6 +77,87 @@ def instrument_fastmcp_v2(server: Any, data: MCPAnalyticsData) -> None:
     # sees: under `FastMCP(strict_input_validation=True)` every call fails with
     # "'context' is a required property".
     _wrap_list_tools(low_level, data, context_required=False)
+    _wrap_resource_requests(low_level, data)
+
+
+def _wrap_resource_requests(server: Any, data: MCPAnalyticsData) -> None:
+    for request_type, event_type in (
+        (mcp_types.ListResourcesRequest, MCPAnalyticsEventType.MCP_RESOURCES_LIST),
```
gesh, line 86: **`resources/templates/list` is not instrumented in any adapter.** MCP resource discovery has two methods: `resources/list` and `resources/templates/list`. This wrapper covers only `ListResourcesRequest`/`ReadResourceRequest`, and `_RESOURCE_METHODS` in `_instrument_v2.py` lists only `resources/list` and `resources/read`. FastMCP registers a `ListResourceTemplatesRequest` handler, and templated resources (`@mcp.resource("users://{id}")`) are advertised only through it. A template-only server therefore shows reads with no matching discovery events, while the README and changeset claim resource discovery is captured. No test registers a templated resource either. Suggestion: wrap `ListResourceTemplatesRequest` / `resources/templates/list` in both adapters, or scope the README claim.

posthog/mcp/_instrumentation.py
```diff
@@ -904,3 +903,43 @@ async def record_tools_list(
...
+    client_name: Optional[str] = None,
+    client_version: Optional[str] = None,
+    protocol_version: Optional[str] = None,
+    extra: Optional[Dict[str, Any]] = None,
+) -> None:
+    """Record resources/list or resources/read without affecting dispatch."""
+    try:
+        params = request.get("params")
+        uri = params.get("uri") if isinstance(params, dict) else None
+        event: Dict[str, Any] = {
```
gesh, line 936: **`$mcp_resources_list` records nothing about the listing outcome.** Both wrappers await the handler result and then discard it; `record_resource_request` receives only the request. The event carries no resource names or URIs, no count, no `nextCursor`, and no empty-list flag — it can only say a listing happened, in N ms. The sibling `record_tools_list` captures `$mcp_listed_tool_names`, the full response, and flags an empty listing as an error. Listing metadata (name/uri/mimeType) is not a resource body, so the README exclusion does not apply. Suggestion: pass the result through and capture listed URIs (sanitized) and a count, for parity with `$mcp_tools_list`.

posthog/test/mcp/test_resources.py
```diff
@@ -0,0 +1,88 @@
...
+    flush_background,
+)
+
+
+@pytest.fixture(params=["official", "fastmcp"])
+def server(request):
+    if request.param == "fastmcp":
+        if MCP_MAJOR >= 2:
+            pytest.skip("jlowin FastMCP requires MCP SDK v1")
+        from fastmcp import FastMCP as Server
```
gesh, line 24: **This bare `from fastmcp import ...` errors instead of skipping when the optional package is absent.** The fixture guards only `MCP_MAJOR >= 2`. On MCP SDK v1 without jlowin `fastmcp` installed (a supported configuration), the four `[fastmcp-*]` parametrizations report ERROR at fixture setup with `ModuleNotFoundError`. Reproduced by running the file in such an environment: 4 passed, 4 errors. The repo convention is `pytest.importorskip("fastmcp")` (see `test_fastmcp_v2.py`). Suggestion: add `pytest.importorskip("fastmcp")` on the fastmcp path.

posthog/test/mcp/test_lowlevel.py
```diff
@@ -62,6 +97,110 @@ async def test_list_tools_injects_optional_context_and_captures():
...
+        with pytest.raises(ValueError) as caught:
+            await read
+        assert str(caught.value) == f"Cannot read {uri}"
+    else:
+        result = await read
+        assert result.root.contents[0].text == "# Guide"
+        assert str(result.root.contents[0].uri) == uri
+    await _flush()
+
+    assert len(_events(client, "$mcp_resources_list")) == 1
```
gesh, line 211: **`$mcp_resources_list` is only asserted by count.** All three test files assert `len(events(..., "$mcp_resources_list")) == 1` and check no property of the event (here, `test_v2_lowlevel.py`, and `test_resources.py`). The `resource_error` parametrization only makes reads fail, so the wrapper error branch for `ListResourcesRequest` and the empty-listing case are never exercised. Suggestion: assert the list event's properties (`$mcp_is_error`, duration, absence of `$mcp_resource_name`) and add a failing-list and an empty-list case.

.sampo/changesets/calm-resource-atlas.md
```diff
@@ -0,0 +1,5 @@
+---
+pypi/posthog: minor
+---
+
+Capture MCP resource discovery and reads from instrumented servers.
```
gesh: **The changeset omits a cross-cutting behavior change to existing events.** The new `_URL_PATTERN.sub(_sanitize_url, ...)` in `_sanitize_string` runs on every captured string of every `$mcp_*` event, including pre-existing `$mcp_tool_call` parameters, responses, and error messages, and `_SENSITIVE_QUERY_KEY_PATTERN` redacts generic keys such as `key`, `sig`, and `auth`. After an upgrade, users will see URLs in existing captured tool-call data rewritten with `%5Bredacted%5D` values, with no changelog note. Suggestion: add a line that describes the new URL-credential redaction and that it also applies to tool-call events.

gesh (commented):
QA swarm review (automated multi-agent pass, adversarially verified). The suite passes both CI legs (SDK v1 and v2), the wrapping topology and failure isolation are sound, and events follow the existing tool-call conventions. The nine comments below are the confirmed findings, all minor severity. The two most useful before merge: fragment redaction in the URL sanitizer, and `importorskip` in `test_resources.py`.

gesh (approved):
Approving to unblock, have a look at the inline comments

posthog/mcp/_sanitization.py
lucasheriques: Fixed in df1c814. A fragment that parses as `key=value` fields gets the same key rules as the query; a plain anchor like `#section-2` stays byte-for-byte. Covered by the `#access_token=...` and `#section-2` rows in `test_sanitize_url_credentials`. Same change shipped to the JS SDK.

posthog/mcp/_sanitization.py
lucasheriques: Fixed in df1c814. `;` is normalized to `&` before splitting query and fragment fields, so `?a=1;token=x` redacts. When nothing is redacted the original string (with its `;`) is returned unchanged; only a rewrite re-serializes with `&`.

posthog/mcp/_sanitization.py
lucasheriques: Fixed in df1c814, and widened: query keys are now matched per `-`/`_`/`.` segment (`auth|token|secret|password|passwd|pwd|credential|signature|sig|key|hmac|sas|bearer|jwt|session|sessionid`), so `jwt`, `sessionid`, `session_id`, `private_token`, `oauth_signature`, `id_token`, `subscription-key` all redact. `code` is matched exactly rather than as a segment so `country_code` and `zip_code` survive. Over-redacting a benign `sort_key` is the accepted trade.

posthog/mcp/_sanitization.py
lucasheriques: Fixed in df1c814 (and made linear in 21f9f55). Trailing `.,;:!?)]}'` is split off before parsing and re-appended, so `See https://x?a=b, then retry.` and `Failed (https://x?a=b).` keep their punctuation. One deliberate exception after Codex review (2b11efc): when the URL's last field was itself redacted, the suffix is dropped, because `?password=hunter2!!!` would otherwise come back as `[redacted]!!!`. So `See https://x?sig=abc, then retry.` now loses the comma while `See https://x?sig=abc&page=2, then retry.` keeps it. Structured fields like `$mcp_resource_name` never strip anything, since there is no prose there.

posthog/mcp/_instrument_lowlevel.py
lucasheriques: Fixed in df1c814. `resources/templates/list` is wrapped on every adapter (low-level v1, FastMCP, v2) and emitted as `$mcp_resources_list`; `$mcp_parameters.request.method` separates it from `resources/list`. Tests register a templated `users://{user_id}/profile` resource on each adapter.

posthog/mcp/_instrumentation.py
lucasheriques: Fixed in df1c814. Listing events now carry the listing result as `$mcp_response` (names, uris, uriTemplates, mimeTypes, nextCursor), which is metadata rather than a body, so it lines up with `$mcp_tools_list`. Reads still capture no response. An empty listing is intentionally not flagged as an error: a template-only server legitimately lists zero static resources.

posthog/test/mcp/test_resources.py
lucasheriques: Fixed in df1c814 with `pytest.importorskip("fastmcp")` on that fixture path.

posthog/test/mcp/test_lowlevel.py
lucasheriques: Fixed in df1c814. `test_lowlevel.py` and `test_v2_lowlevel.py` now assert the list event's method, `$mcp_response` uris, `$mcp_is_error`, `$mcp_duration_ms`, and absence of `$mcp_resource_name`, with parametrized static / empty / templates cases plus a failing-listing case that checks the `$exception` sibling. `test_resources.py` does the same through the high-level adapters.

.sampo/changesets/calm-resource-atlas.md
lucasheriques: Fixed in df1c814. The changeset now says URL credential redaction applies to every captured string, including existing `$mcp_tool_call` parameters, responses and error messages, and that URLs in existing tool-call data will show `%5Bredacted%5D` values after upgrading.

lucasheriques:
Addressed every inline finding above (replies on each thread), then ran an additional cross-model review pass on the result, which surfaced more URL-sanitizer gaps. All fixed on this branch, and every change ships identically in posthog-js#4830 so both SDKs redact the same bytes (verified by diffing 46 shared vectors through both sanitizers):

- Apostrophes are valid in URIs, so `https://example.com/o'reilly?token=x` no longer stops the match before the credential (e5892c9).
- Punctuation split off the end of a URL is dropped when the last field was redacted, so `?password=hunter2!!!` cannot come back as `[redacted]!!!`; a URL that is the whole captured string is never stripped (2b11efc).
- A URL nested more than one level deep in a retained query value is dropped rather than trusted (2b11efc).
- PostHog tokens are redacted before URLs are re-serialized, since percent-encoding hid the token boundary (2b11efc).
- Intent PII redaction runs before the generic pass for the same reason (1d37981).
- `resource_name` is sanitized with the URL and PostHog-token passes only, so tool names like `Get_Organization_Memberships` are not eaten by the entropy detector (65ccc76).
- Authority-less resource URIs (`resource:guide?token=x`, `file:/guide.md?token=x`) are covered; over-matching prose is harmless because an unchanged match is returned byte-for-byte (5f5a65e, 36e175e).
- A match holding a prose word before an address (`URL:https://u:p@host`) or two addresses run together (`doc,https://u:p@host`) is split at the second address (98abc7b, b50cd56, 1999e40).

Two things deliberately left for separate PRs: the low-level v1 adapter does not wrap resource handlers registered after `instrument()`, which is the same pre-existing limitation `tools/call` has on that adapter; and the entropy detector already redacts identifier-shaped tool names inside `$mcp_parameters` on main.

lucasheriques:
Two more from the same review pass, pushed on top:

- The known-secret detector now runs before the URL pass. Rewriting a URL could grow a string past the detector's scan window, so a GitHub token that main redacted was surviving (f0c6c57). Side effect: rewritten authority-less URIs like `resource:guide?token=%5Bredacted%5D` keep their readable form instead of collapsing to `[redacted]`, so Python and JS now agree byte-for-byte on every shared vector.
- Hash-routed fragments (`#/callback?token=x`) parse their fields after the route instead of treating `/callback?token` as the key (3419fe3).

lucasheriques:
Two more, pushed on top (same in posthog-js#4830):

- A `?` inside a fragment value (`#access_token=x&next=https://other/?page=1`) is no longer taken as a route boundary; a route is only split off when the `?` precedes the first `=` (757aa3c).
- The narrated intent runs the binary gate before PII redaction, so a large base64 blob containing a Luhn-valid digit run still collapses to the binary marker instead of being captured (e9e82dd).

lucasheriques:
Two more, pushed on top:

- A plain fragment (no `=`) is now run through the URL text pass, so two Markdown links run together as `#intro)[b](https://user:pw@...` no longer hide the second address (916eadf). Same in posthog-js#4830.
- On MCP SDK v2 a failing read is wrapped in `MCPError` / `UnexpectedResourceError` with a generic message; the failure details now unwrap that chain the same way tool dispatch does, so `$mcp_error_type` and `$mcp_error_message` carry what the handler actually raised. The caller still receives the SDK wrapper. v1's `ResourceError` already keeps the handler message, so it is left as is (19a696c). The TypeScript SDK observes the raw throw on both majors, so it needed no change.

lucasheriques:
One more hardening commit, pushed on top (same in posthog-js#4830): a fragment route prefix that is itself an address (`#https://user:pw@host/doc?page=1`) now gets the URL text pass; the fragment text pass respects the one-level nesting budget so `resource:x#` repeated 10,000 times finishes in 10 ms instead of raising RecursionError; and the address split is a single pass with the 8 KB bound applied per piece, so 10,000 comma-joined URLs are sanitized individually (6c980c6).

lucasheriques:
One more, pushed on top (same in posthog-js#4830): a hash route that carries its own `=` before its `?` (`#/docs/id=1?token=x`) is now recognized as a route, and `/` counts as a credential-name segment separator so `#/token=x` redacts (84655d0). Over-redacting `sort/key` is the same accepted trade as `sort_key`.

lucasheriques:
Two more, pushed on top (same in posthog-js#4830): an adjacent address is now split off wherever it sits unless it directly follows `=` (a field value), which closes the case where a second Markdown link was absorbed into a query key (67a76c6); and the fragment route heuristics are gone: a fragment is split at its first `?` and each half gets the field pass when it contains `=` and the text pass otherwise, so `#/token=x&next=https://other/?page=1` no longer masquerades as a route (2502fb8).

lucasheriques:
Two more, pushed on top (same in posthog-js#4830): an adjacent address is only kept with its field when it sits in value position inside the query or fragment (nearest structural character before it is `=`, and only past the first `?`/`#`), so a `=` in the path no longer hides a credential and a credential value containing a URL after `%20` is redacted whole (d867631, b7179b9); and a `?` inside a fragment credential fails closed: when the head's last field was rewritten the whole tail is replaced with `[redacted]` (b7179b9).

lucasheriques:
Four more, pushed on top (same in posthog-js#4830): value position is decided in one forward pass instead of a backward scan per address, which was quadratic on thousands of joined URLs (d3ec6c8); only a URL's real delimiters count as structural, so a second `?` inside a credential value no longer splits it, and the fail-closed rules key off a field being sensitive rather than only changed (b760dc6); `;` is no longer normalized to `&` before parsing, since that split a credential containing a semicolon, and a value that hides a `;`-separated credential is redacted whole instead (7371ec7); and the narrated intent redacts credentials before PII, so a token containing a phone-shaped run is removed whole (d05b16e).

lucasheriques:
One more, pushed on top (same in posthog-js#4830): `;` now counts as a credential-name segment separator so `?download;token=x` redacts, and a credential's suffix after `;` or after a fragment's `?` stays attached to its field so the fail-closed rules see all of it (f70061d).

PR #941, opened 2026-09-10
fix: preserve redirect URLs with async capture host prefixes

nickita-khylkouski:

When the configured capture host contains a path prefix, passing relative redirect paths through an HTTPX client can reapply or retain that prefix incorrectly. Resolve each redirect against the current absolute URL, enforce the original origin, and pass absolute URLs through the redirect chain.

- Full test suite: 2,502 passed, 15 skipped, 38 subtests passed
- Full mypy check: 230 source files, no issues
- Import smoke test and package build
- Added 307/308 coverage for absolute, root-relative, relative, and query-only redirects with prefixed hosts

The repository-wide Ruff command currently reports existing formatting and legacy-lint findings in untouched files; the modified Python files pass formatting.

## :pencil: Checklist

### If releasing new changes

An equivalent patch changeset is included at `.sampo/changesets/async-capture-redirect-path-prefix.md`.

## 🤖 Agent context

The contributor directed the scope and redirect behavior. Codex helped inspect the repository, apply the bounded change, and run the repository checks; same-origin enforcement remains in place and fragments are not forwarded.

### DIFF
--- a/.sampo/changesets/async-capture-redirect-path-prefix.md
+++ b/.sampo/changesets/async-capture-redirect-path-prefix.md
@@ -1,0 +1,5 @@
+---
+pypi/posthog: patch
+---
+
+Follow async capture redirects without duplicating or incorrectly retaining a configured host path prefix.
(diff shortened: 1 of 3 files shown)


marandaneto:
thanks @nickita-khylkouski

marandaneto (approved):

PR #942, opened 2026-09-11
fix(ai): harden Prompts.get_all against old servers and malformed rows

jurajmajerik:

Review of the posthog-js port (PostHog/posthog-js#4903) found two gaps that also exist in the merged `Prompts.get_all` (#938):

- On an old server that ignores the `label` param, prompts whose label happens to point at the latest version still pass the row check. `get_all` then returns that subset as a complete result, silently missing the rest. The old guard only fired when no row passed.
- A malformed row (for example a non-string prompt) was skipped like a moved label, instead of raising the invalid-response error the per-name path raises.

Best to land before `get_all` ships to PyPI, so the first published version has the hardened behavior.

## Changes

- The row check now has three states. A row whose `all_labels` does not contain the label at all can only come from a server that did not filter, so even one such row makes `get_all` raise. A label present at a different version stays the benign mid-request race: skip and warn.
- A malformed row raises instead of being skipped.
- Rows are validated before anything is cached, so a rejected batch leaves the cache untouched.
- Both new failures go through `_maybe_capture_error`, like the other fetch failures.

- Updated the old-server test to the partial case: one row that looks resolved next to one unlabeled row now raises, and nothing is cached.
- New test: a malformed row raises and leaves the cache empty, even when a valid row came first.
- Ran the full prompts test file, ruff, mypy (baseline clean), and the public API snapshot check.

## :pencil: Checklist

- [ ] I updated the docs if needed. (No docs exist for get_all yet; coming with the release.)

### If releasing new changes

- [x] Changeset added in `.sampo/changesets/` (written by hand, same format as `sampo add`)

## 🤖 Agent context

- Written with Claude Code, directed by the assignee. Ports posthog-js commits 3d8197f56 and 25a7bd5cd from PostHog/posthog-js#4903 review findings back to Python.
- The `__proto__` fix from that review does not apply here: Python dicts have no prototype.

### DIFF
--- a/posthog/ai/prompts.py
+++ b/posthog/ai/prompts.py
@@ -425,6 +425,18 @@ class Prompts:
                 continue
             resolved_rows.append(row)
 
+        if rows and not resolved_rows:
+            # Every returned row was skipped as moved. One moved label is a
+            # mid-request race, but all of them means the server most likely
+            # ignored the label param and served latest versions.
+            compat_error = Exception(
+                f'[PostHog Prompts] The server returned prompts, but none resolve label "{label}". '
+                "It may not support fetching prompts by label on the list endpoint yet. "
+                "Upgrade PostHog, or fetch prompts one by one with get()."
+            )
+            self._maybe_capture_error(compat_error, name="*", version=None, label=label)
+            raise compat_error
+
         now = time.time()
         results: Dict[str, PromptResult] = {}
         for row in resolved_rows:
--- a/posthog/ai/prompts.py
+++ b/posthog/ai/prompts.py
@@ -385,8 +385,16 @@ class Prompts:
         results: Dict[str, PromptResult] = {}
         skipped: List[str] = []
         for row in rows:
-            if not _is_prompt_api_response(row) or not _row_resolves_label(row, label):
-                skipped.append(str(row.get("name")) if isinstance(row, dict) else "?")
+            if not _is_prompt_api_response(row):
+                invalid_error = Exception(
+                    f'[PostHog Prompts] Invalid response format for prompts with label "{label}"'
+                )
+                self._maybe_capture_error(
+                    invalid_error, name="*", version=None, label=label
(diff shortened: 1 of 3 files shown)


marandaneto (approved):
found an issue, see comment, approving to unblock

posthog/ai/prompts.py
```diff
@@ -381,14 +391,43 @@ def get_all(self, *, label: str) -> Dict[str, PromptResult]:
...
+                compat_error = Exception(
+                    f'[PostHog Prompts] The server returned a prompt that does not carry label "{label}". '
+                    "It may not support fetching prompts by label on the list endpoint yet. "
+                    "Upgrade PostHog, or fetch prompts one by one with get()."
+                )
+                self._maybe_capture_error(
+                    compat_error, name="*", version=None, label=label
+                )
+                raise compat_error
+            if label_state == "moved":
```
marandaneto, line 423: blocking: Preserve rejection when no rows resolve the label — An old server can return latest versions while every prompt’s `production` label points to an earlier version. These rows are now classified as `"moved"` and skipped; removing the final no-results guard makes `get_all()` return `{}` instead of raising the compatibility error. This silently reports no matching prompts despite labeled prompts existing, regressing the previous behavior. Preserve rejection or otherwise verify resolution before accepting an entirely skipped batch. Reproduction: reproduced — `uv run --no-sync python -m pytest -q --timeout=30 test_review_get_all.py` fails on head with `DID NOT RAISE` for a mocked list response containing latest version 2 labeled at version 1, and passes on the merge base.

marandaneto (commented):

posthog/ai/prompts.py
jurajmajerik: fixed in https://github.com/PostHog/posthog-python/pull/942/commits/8d26eb199bfd659ac09a0b59c37dcc958af43ce3

PR #944, opened 2026-09-11
chore(mcp): enable model capture and conversation correlation by default

lucasheriques:
## Motivation and changes

Model capture and conversation correlation currently require separate opt-ins. Enable both by default, preserve explicit opt-outs, and leave missing-capability reporting and feedback collection disabled for Gesh's work. Both SDKs follow [ADR-0013](https://github.com/PostHog/posthog-js/blob/codex/mcp-analytics-defaults/packages/mcp/docs/adr/0013-analytics-capture-is-on-by-default.md), which records the decision and compatibility tradeoffs.

Compatible schemas gain `llm_model` (required on the official high-level adapters, optional elsewhere, never enforced by analytics dispatch) and optional `conversation_id`. An echoed handle takes precedence, but a newly minted handle no longer replaces an existing transport session or PostHog session token. Sessionless calls can still receive a handle to echo. `MCPAnalyticsOptions(capture_model=False, enable_conversation_id=False)` restores the previous defaults. Model capture is also on by default on `PostHogMCP`; the public API snapshot records the new defaults.

Fresh raw low-level instances read analytics arguments when ownership is unknown and strip only with evidence of SDK ownership, following ADR-0011. No host catalog is replayed during calls. Standalone FastMCP reads ownership from its registered schema, accounts for references/composition and middleware, and preserves application-declared keys. Transient registry lookup failures now preserve potentially host-owned arguments; actual not-found results retain middleware fallback behavior.

On MCP SDK v1, standalone FastMCP middleware that overrides listing or dispatch disables model argument injection and invalidates stale ownership. This prevents a fresh replica from rejecting an advertised analytics argument and preserves a replacement tool's own `llm_model` argument. Model capture from client metadata remains enabled; self-report injection is unavailable on these middleware installations, including pass-through logging/auth middleware. No new public option or catalog replay is introduced.

## Validation

- Regression coverage includes fresh instances with pass-through middleware, client metadata capture, replacement tools, listing order, and middleware added after cached discovery.
- MCP v1 suite: 553 passed, 1 version-specific skip.
- MCP v2 suite: 444 passed, 21 version-specific skips.
- Ruff lint/format, mypy baseline (239 files), and public API snapshot check pass.
- CodeScene pre-commit safeguard passes for the middleware fixes, improving both runtime files. Existing large-module and regression-test complexity warnings remain.
- Merged current `main` and verified its virtual-tool collision regressions alongside the new defaults, resolving conflicts without rewriting branch history.

## Release

`.sampo/changesets/mcp-analytics-defaults.md` requests a minor release. Human review and the normal release workflow are still required; no package has been published from this branch.

## Agent context

**Autonomy:** Human-driven (agent-assisted), directed by Lucas Faria.

Initial implementation and latest review fixes with Codex. Ownership rework and earlier review response with Claude Code. Human review required, particularly for `_standalone_ownership` and its helpers in `posthog/mcp/_instrument_lowlevel.py`.

https://claude.ai/code/session_012YBJXRzzCizpHFGn2mZaEg

### DIFF
--- a/.sampo/changesets/mcp-analytics-defaults.md
+++ b/.sampo/changesets/mcp-analytics-defaults.md
@@ -1,0 +1,5 @@
+---
+pypi/posthog: minor
+---
+
+Enable MCP model capture and conversation correlation by default, including on fresh low-level servers.
--- a/posthog/mcp/README.md
+++ b/posthog/mcp/README.md
@@ -36,10 +36,24 @@ Request headers use the same identity and package version, so SDK Health can com
 Because `$lib` is a client-level identity, `instrument()` relabels every event sent by the client passed to it.
 Use a client dedicated to MCP analytics if the application also captures unrelated events.
 
+## Defaults and opt-outs
+
+Intent, model capture, conversation correlation, and MCP exception capture are on by default.
+Missing-capability reporting and feedback collection remain off.
+
+```python
+instrument(server, posthog, MCPAnalyticsOptions(capture_model=False, enable_conversation_id=False))
+```
+
+Conversation correlation adds an optional `conversation_id` argument and returns a handle in
+eligible tool results. Clients must echo it to group later calls; calls without it mint new handles.
+Set `enable_conversation_id=False` to retain transport-based session grouping and unchanged
+response content. Custom `PostHogMCP` dispatchers enable model capture by default but still
+supply their own session IDs.
+
 ## Capture the calling model
 
-Model capture is off by default. Enable it for an instrumented MCP Python SDK 1.x or
-2.x server:
+Model capture is on by default for instrumented MCP Python SDK 1.x and 2.x servers:
 
 ```python
 from posthog.mcp import MCPAnalyticsOptions, instrument
@@ -47,7 +61,6 @@ from posthog.mcp import MCPAnalyticsOptions, instrument
 analytics = instrument(
     server,
     posthog,
-    MCPAnalyticsOptions(capture_model=True),
 )
 ```
 
@@ -74,6 +87,12 @@ Model self-reporting only runs when PostHog can prove it owns the injected field
 If a tool already declares `llm_model`, or uses a root `$ref`, `oneOf`, `allOf`, or
 `anyOf` schema, PostHog leaves the schema and argument untouched. Client metadata
 can still be captured in those cases.
+
+On fresh low-level instances, model argument ownership is resolved from the original tool
+listing before dispatch. This internal lookup emits no discovery event and stops after 16 pages
+or 250 ms. If the listing fails or omits the tool, its arguments remain unchanged and self-reported
+model capture stays empty; recognized client metadata can still supply the model. Existing
+listings and high-level registries continue to supply ownership directly.
 
 For a custom dispatcher, use the same option on `PostHogMCP` and pass request
 metadata through explicitly:
@@ -229,7 +248,7 @@ sess = get_mcp_session(request)   # sess.session_id, sess.client_name, ...
 
 ### Or skip the middleware entirely: conversation ids
 
-`MCPAnalyticsOptions(enable_conversation_id=True)` derives `$session_id` from the
+`enable_conversation_id` is on by default and derives `$session_id` from the
 agent's conversation handle, deterministically and identically on every pod. That
 needs no middleware and no ordering discipline, and it is the only thing that
 correlates a session under the 2026-07-28 revision's per-request server instances.
--- a/posthog/mcp/_instrument_lowlevel.py
+++ b/posthog/mcp/_instrument_lowlevel.py
@@ -42,6 +42,7 @@ from ._instrumentation import (
 )
 from ._internal import MCPAnalyticsData
 from ._model_parameters import request_meta_from_context
+from ._tool_schema import resolve_model_ownership
 from ._output_instructions import mirror_instructions_into_structured_content
 from .logger import log
 from .tools import get_more_tools_result_text, resolve_missing_capability_tool_name
@@ -171,6 +172,31 @@ def _wrap_resource_request(
     handlers[request_type] = handler
 
 
+async def _prepare_model_arguments(
+    server: Any, data: MCPAnalyticsData, req: Any, strip_injected: bool
+) -> Tuple[Any, bool]:
+    async def list_page(cursor: Optional[str]) -> Any:
+        listing = server.request_handlers.get(mcp_types.ListToolsRequest)
+        raw = getattr(listing, "__posthog_mcp_original__", listing)
+        return await raw(
+            mcp_types.ListToolsRequest(
+                method="tools/list",
+                params=mcp_types.PaginatedRequestParams(
+                    cursor=cursor, _meta=req.params.meta
+                ),
+            )
+        )
+
+    owns_model = await resolve_model_ownership(data, req.params.name, list_page)
+    if strip_injected or not owns_model:
+        return req, owns_model
+    arguments = dict(req.params.arguments or {})
+    arguments.pop("llm_model", None)
+    return req.model_copy(
+        update={"params": req.params.model_copy(update={"arguments": arguments})}
+    ), owns_model
+
+
 def _wrap_call_tool(
     server: Any, data: MCPAnalyticsData, *, strip_injected: bool, high_level: Any = None
 ) -> None:
@@ -182,6 +208,10 @@ def _wrap_call_tool(
     async def handler(req: Any) -> Any:
         name = req.params.name
         arguments = dict(req.params.arguments or {})
+
+        req, analytics_owns_model = await _prepare_model_arguments(
+            server, data, req, strip_injected
+        )
         client_name, client_version = _client_info(server)
         protocol_version = _protocol_version(server)
         mcp_session_id = _mcp_session_id(server)
@@ -195,9 +225,7 @@ def _wrap_call_tool(
             name=name,
             arguments=arguments,
             request_meta=request_meta_from_context(_request_context(server)),
-            allow_self_reported_model=data.tool_model_parameter_injected.get(
-                name, False
-            ),
+            allow_self_reported_model=analytics_owns_model,
             mcp_session_id=mcp_session_id,
             token=token,
             client_name=client_name,
@@ -240,7 +268,7 @@ def _wrap_call_tool(
         if strip_injected and req.params.arguments:
             owned = await _tool_owned_injected_keys(high_level, name)
             injected_keys = ["context", "conversation_id"]
-            if data.tool_model_parameter_injected.get(name, False):
+            if analytics_owns_model:
                 injected_keys.append("llm_model")
             for key in injected_keys:
                 if key not in owned:
@@ -424,6 +452,7 @@ def _wrap_list_tools(
 
         return result
 
+    setattr(handler, "__posthog_mcp_original__", original)
     setattr(handler, _WRAPPED_FLAG, True)
     handlers[mcp_types.ListToolsRequest] = handler
 
--- a/posthog/mcp/_instrument_v2.py
+++ b/posthog/mcp/_instrument_v2.py
@@ -61,6 +61,7 @@ from ._model_parameters import (
     request_meta_from_context,
 )
 from ._output_instructions import mirror_instructions_into_structured_content
+from ._tool_schema import resolve_model_ownership
 from .logger import log
 from .request_headers import get_request_headers
 from .session_token import read_mcp_session_header
@@ -477,6 +478,40 @@ async def _standalone_injected_parameters(
     return frozenset(key for key in injected if not schema_has_param(schema, key))
 
 
+async def _prepare_raw_v2_arguments(
+    server: Any, data: MCPAnalyticsData, ctx: Any, params: Any
+) -> Tuple[Any, bool]:
+    async def list_page(cursor: Optional[str]) -> Any:
+        listing = server.get_request_handler(_LIST_METHOD)
+        raw = getattr(listing.handler, "__posthog_mcp_original__", listing.handler)
+        return await raw(ctx, mcp_types.PaginatedRequestParams(cursor=cursor))
+
+    owns_model = await resolve_model_ownership(data, params.name, list_page)
+    if not owns_model:
+        return params, False
+    arguments = dict(params.arguments or {})
+    arguments.pop("llm_model", None)
+    return params.model_copy(update={"arguments": arguments}), True
+
+
+async def _prepare_v2_arguments(
+    server: Any, data: MCPAnalyticsData, ctx: Any, params: Any
+) -> Tuple[Any, bool]:
+    standalone = data.standalone_fastmcp() if data.standalone_fastmcp else None
+    if standalone is None:
+        return await _prepare_raw_v2_arguments(server, data, ctx, params)
+    version = _requested_tool_version(ctx)
+    injected = await _standalone_injected_parameters(
+        standalone, data, params.name, version
+    )
+    if injected is None:
+        return params, False
+    arguments = dict(params.arguments or {})
+    for key in injected:
+        arguments.pop(key, None)
+    return params.model_copy(update={"arguments": arguments}), "llm_model" in injected
+
+
 def _wrap_v2_call_tool(server: Any, data: MCPAnalyticsData) -> None:
     entry = server.get_request_handler(_CALL_METHOD)
     if entry is None or getattr(entry.handler, _WRAPPED_FLAG, False):
@@ -486,21 +521,10 @@ def _wrap_v2_call_tool(server: Any, data: MCPAnalyticsData) -> None:
     async def handler(ctx: Any, params: Any) -> Any:
         name = params.name
         arguments = dict(params.arguments or {})
-        analytics_owns_model = data.tool_model_parameter_injected.get(name, False)
-        standalone = data.standalone_fastmcp() if data.standalone_fastmcp else None
-        if standalone is not None:
-            version = _requested_tool_version(ctx)
-            injected = await _standalone_injected_parameters(
-                standalone, data, name, version
-            )
-            analytics_owns_model = injected is not None and "llm_model" in injected
-            if injected is not None:
-                call_arguments = {
-                    key: value
-                    for key, value in arguments.items()
-                    if key not in injected
-                }
-                params = params.model_copy(update={"arguments": call_arguments})
+
+        params, analytics_owns_model = await _prepare_v2_arguments(
+            server, data, ctx, params
+        )
         token, client_name, client_version, protocol_version, mcp_session_id = (
             _resolve_ctx(ctx)
         )
@@ -715,6 +739,7 @@ def _wrap_v2_list_tools(
 
         return result
 
+    setattr(handler, "__posthog_mcp_original__", original)
     setattr(handler, _WRAPPED_FLAG, True)
     _replace_handler(server, _LIST_METHOD, handler, entry.params_type)
 
--- a/posthog/mcp/_tool_schema.py
+++ b/posthog/mcp/_tool_schema.py
@@ -1,0 +1,68 @@
+"""Resolve model argument ownership on low-level servers without a tool registry."""
+
+from __future__ import annotations
+
+import asyncio
+from typing import Any, Awaitable, Callable, Optional
+
+from ._internal import MCPAnalyticsData
+from ._model_parameters import can_inject_model_parameter, is_capture_model_enabled
+from .logger import log
+
+
+async def resolve_model_ownership(
+    data: MCPAnalyticsData,
+    name: str,
+    list_page: Callable[[Optional[str]], Awaitable[Any]],
+) -> bool:
+    if not is_capture_model_enabled(data.options.capture_model):
+        return False
+    if name in data.tool_model_parameter_injected:
+        return data.tool_model_parameter_injected[name]
+    try:
+        return await asyncio.wait_for(
+            _find_model_ownership(name, list_page), timeout=0.25
+        )
+    except Exception:  # noqa: BLE001 - discovery must not prevent tool dispatch
+        log(
+            "Warning: Could not resolve model argument ownership; leaving tool arguments unchanged."
+        )
+        return False
+
+
+async def _find_model_ownership(
+    name: str, list_page: Callable[[Optional[str]], Awaitable[Any]]
+) -> bool:
+    cursor = None
+    seen = set()
+    for _ in range(16):
+        response = await list_page(cursor)
+        result = getattr(response, "root", response)
+        ownership = _model_ownership(result, name)
+        if ownership is not None:
+            return ownership
+        cursor = _next_cursor(result)
+        if cursor is None or cursor in seen:
+            return False
+        seen.add(cursor)
+    return False
+
+
+def _model_ownership(result: Any, name: str) -> Optional[bool]:
+    for tool in getattr(result, "tools", []):
+        if getattr(tool, "name", None) == name:
+            schema = getattr(tool, "input_schema", None)
+            if schema is None:
+                schema = getattr(tool, "inputSchema", None)
+            return can_inject_model_parameter(schema)
+    return None
+
+
+def _next_cursor(result: Any) -> Optional[str]:
+    if hasattr(result, "next_cursor"):
+        cursor = result.next_cursor
+    else:
+        cursor = getattr(result, "nextCursor", None)
+    if not isinstance(cursor, str):
+        return None
+    return cursor or None
--- a/posthog/mcp/posthog_mcp.py
+++ b/posthog/mcp/posthog_mcp.py
@@ -71,7 +71,7 @@ class PostHogMCP(Client):
         api_key: str,
         missing_capability_tool_name: Optional[str] = None,
         mcp_exception_autocapture: bool = True,
-        capture_model: Union[bool, MCPAnalyticsModelOptions] = False,
+        capture_model: Union[bool, MCPAnalyticsModelOptions] = True,
         collect_feedback: Union[bool, CollectFeedbackOptions] = False,
         **kwargs: Any,
     ) -> None:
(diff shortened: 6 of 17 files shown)


.sampo/changesets/mcp-analytics-defaults.md
```diff
@@ -0,0 +1,5 @@
+---
+pypi/posthog: minor
```
gesh, line 2: > [!NOTE]

**[convergent: router + paul]** 🟠 HIGH

Two reviewers landed on this line independently: is `minor` the right tier?

Verified empirically — with `capture_model=True` by default, the advertised `inputSchema` of every tool on the official high-level FastMCP / `MCPServer` adapters now lists `llm_model` in `required` (next to the already-default `context`). On a routine `pip install --upgrade posthog`, every existing deployment's wire-visible tool contract changes with **zero code change by the user**. `enable_conversation_id=True` adds a handle to eligible tool responses on top of that.

The dispatch path does **not** enforce the `required` flag — a call omitting `context` and `llm_model` still dispatches fine — so there is no server-side breakage. The risk is client-side: strict-schema MCP clients that validate before sending, and tooling that generates call templates from a cached schema.

Paul's read: "i can talk myself into minor (`capture_model` only landed in #927 six days ago, so the blast radius is genuinely small, and `context` already set the precedent of injecting a required arg), and i'm not going to block on the tier. but the thing i'd actually want confirmed: `references/public_api_snapshot.txt` changes on four lines here, and CONTRIBUTING.md now says that means 'this touches public API, agree the shape on the issue first'. i'm lazily asking rather than digging — was this one agreed somewhere?"

.sampo/changesets/mcp-analytics-defaults.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+Enable MCP model capture and conversation correlation by default, including on fresh low-level servers.
```
gesh: > [!NOTE]

**[paul]** 🟡 MEDIUM

this line is the *only* thing most upgraders will ever read about this change — especially the dependabot-bump crowd who didn't choose to upgrade. right now it says what we turned on but not what happens to them. i'd want it to name the two observable consequences and the escape hatch in one breath: advertised tool schemas gain a required `llm_model`, eligible tool responses gain a conversation handle, `MCPAnalyticsOptions(capture_model=False, enable_conversation_id=False)` restores the old shape.

also — two independent behaviour flips in one release means anyone who gets bitten has to bisect two flags to find out which one did it. would you consider two changesets over two releases? not blocking, ship as you see fit.

posthog/mcp/types.py
```diff
@@ -197,8 +197,8 @@ class MCPAnalyticsOptions:
     # Extra properties merged onto every auto-captured event.
     event_properties: Optional[EventPropertiesFn] = None
     # Capture the model from recognized client metadata, falling back to an
-    # SDK-injected llm_model argument. Off by default.
-    capture_model: Union[bool, MCPAnalyticsModelOptions] = False
+    # SDK-injected llm_model argument. On by default; False disables capture.
+    capture_model: Union[bool, MCPAnalyticsModelOptions] = True
```
gesh, line 204: > [!NOTE]

**[paul]** 🟠 HIGH

this is the one i keep coming back to. `capture_model` is a single boolean doing two completely unrelated things: (a) read the model out of recognised client `_meta` — invisible, read-only, no wire change at all; and (b) inject `llm_model` into the advertised schema and stick it in `required` — very much a wire change. i'm allergic to flags like that generally, but here it also blocks the rollout i'd want: default-on the half that only *observes*, keep the half that *changes the contract* opt-in, and use a release or two of `$mcp_llm_model_source: client_metadata` data to find out how much self-reporting is even buying us. log before you limit.

`MCPAnalyticsModelOptions` already exists as the object form and currently only carries `description`, so there's a ready-made home for something like `inject_argument: bool = False`. that'd let this ship default-on today with no consumer-visible change, and the schema half flip once we've seen numbers. feel free to disagree — but "the schema we advertise to the world changes on a patch-ish upgrade" is exactly the class of thing i'd want a staged path for.

posthog/mcp/types.py
```diff
@@ -183,7 +183,7 @@ class MCPAnalyticsOptions:
     logger: Optional[LoggerFn] = None
     report_missing: bool = False
     missing_capability_tool_name: Optional[str] = None
-    enable_conversation_id: bool = False
+    enable_conversation_id: bool = True
```
gesh, line 189: > [!NOTE]

**[paul]** 🟡 MEDIUM

this one i'm much more relaxed about — an *optional* argument plus a handle in the response content is a far gentler change than a new `required` field. two thoughts though. first: the response-payload part is the bit someone will actually notice, because anything asserting on exact tool output starts seeing an extra content block (`test_defaults.py`'s `len(result.root.content) == (2 if enabled else 1)` is the tell). second: how does a surprised operator find their way to the off switch? the module already has good form here — the session-middleware misconfiguration logs once to `posthog.mcp` — so i wonder if the first injection could do the same, one line, naming the option. that's the difference between "weird, where did this come from" and a thirty-second fix.

posthog/mcp/posthog_mcp.py
```diff
@@ -71,7 +71,7 @@ def __init__(
         api_key: str,
         missing_capability_tool_name: Optional[str] = None,
         mcp_exception_autocapture: bool = True,
-        capture_model: Union[bool, MCPAnalyticsModelOptions] = False,
+        capture_model: Union[bool, MCPAnalyticsModelOptions] = True,
```
gesh, line 81: > [!NOTE]

**[paul]** 🟢 LOW

not blocking at all, just planting a seed. the only way off this default is an argument in source, which means a code change and a redeploy — and the person upgrading is often not the person who wrote the `instrument()` call. an env-var override (`POSTHOG_MCP_CAPTURE_MODEL=0` or similar) reading through to this default would let someone turn it off at 3am without shipping code. knight capital is my usual sermon here (https://specbranch.com/posts/knight-capital/) and it's admittedly a bit heavy for an analytics field — but "make it configurable and measurable" is cheap now and expensive later. ship as you see fit.

posthog/mcp/README.md
```diff
@@ -36,18 +36,31 @@ Request headers use the same identity and package version, so SDK Health can com
 Because `$lib` is a client-level identity, `instrument()` relabels every event sent by the client passed to it.
 Use a client dedicated to MCP analytics if the application also captures unrelated events.
 
+## Defaults and opt-outs
+
+Intent, model capture, conversation correlation, and MCP exception capture are on by default.
```
gesh, line 44: > [!NOTE]

**[paul]** 🟡 MEDIUM

the section does the right job for conversation correlation — it says plainly that responses change and how to get the old behaviour back. model capture doesn't get the same treatment, and it's the riskier of the two. nothing here tells the reader that their advertised `inputSchema` now lists `llm_model` in `required` on the official high-level adapters.

the person who needs this section is someone whose strict-schema client just started complaining. they won't grep for `capture_model`, they'll grep for `llm_model` or `required`. one sentence with those words in it would land them here. worth saying too that the dispatch path doesn't enforce `required`, so servers keep working — that's genuinely reassuring and currently unsaid.

posthog/mcp/README.md
```diff
@@ -36,18 +36,31 @@ Request headers use the same identity and package version, so SDK Health can com
...
+Missing-capability reporting and feedback collection remain off.
+
+```python
+instrument(server, posthog, MCPAnalyticsOptions(capture_model=False, enable_conversation_id=False))
```
gesh, line 50: > [!NOTE]

**[paul]** ⚪ NIT

nit picking, and i almost didn't mention it — but this is the snippet someone pastes while something is on fire, and it's the only snippet in the file that doesn't import what it uses. `from posthog.mcp import MCPAnalyticsOptions, instrument` above it and it's copy-paste-runnable.

posthog/mcp/README.md
```diff
@@ -36,18 +36,31 @@ Request headers use the same identity and package version, so SDK Health can com
...
 ## Capture the calling model
 
-Model capture is off by default. Enable it for an instrumented MCP Python SDK 1.x or
-2.x server:
+Model capture is on by default for instrumented MCP Python SDK 1.x and 2.x servers:
 
 ```python
 from posthog.mcp import MCPAnalyticsOptions, instrument
 
 analytics = instrument(
```
gesh: > [!NOTE]

**[paul]** ⚪ NIT

now that this is the default, the code block under here demonstrates... the absence of configuration. `instrument(server, posthog)` is already the very first example at the top of the README. i'd either drop the block and let the prose carry it, or make it earn its place by showing the object form (`MCPAnalyticsModelOptions(description=...)`) or the opt-out.

same flavour of thing further down in the custom-dispatcher example, which still passes `capture_model=True` explicitly — harmless, but it teaches the reader that they need it.

posthog/mcp/_instrument_lowlevel.py
```diff
@@ -171,6 +172,31 @@ async def handler(req: Any) -> Any:
     handlers[request_type] = handler
 
 
+async def _prepare_model_arguments(
```
gesh: > [!NOTE]

**[router]** 🟡 MEDIUM

`_prepare_model_arguments` always resolves model ownership via the async, paginated `resolve_model_ownership` walk (up to 16 pages, 250 ms timeout) — including for the `instrument_fastmcp_v2` (jlowin standalone FastMCP 2.0) path, which already has a live tool registry available via `high_level.get_tool(name)` (used a few lines down in `_tool_owned_injected_keys` to synchronously check `context` / `conversation_id` ownership).

Compare `_analytics_owns_model_v2` in `posthog/mcp/_instrument_v2.py:253`, which for the analogous v2 high-level `MCPServer` case computes model ownership synchronously and cheaply from `high_level._tool_manager.get_tool(name)`, falling back to the cached dict on failure — no discovery walk needed. The standalone path could likely do the same, avoiding up to 250 ms of latency on the first call to every tool even when a live registry answers the question instantly.

Worth confirming this is an intentional inconsistency (e.g. because standalone-fastmcp schemas can contain nested Pydantic-model refs that only the raw-schema discovery detects correctly) rather than an oversight.

gesh (commented):
> [!NOTE]

QA Swarm review complete — see inline comments. Summary posted separately.

posthog/mcp/_tool_schema.py
```diff
@@ -0,0 +1,71 @@
...
+    name: str,
+    list_page: Callable[[Optional[str]], Awaitable[Any]],
+) -> bool:
+    if not is_capture_model_enabled(data.options.capture_model):
+        return False
+    if name in data.tool_model_parameter_injected:
+        return data.tool_model_parameter_injected[name]
+    try:
+        ownership = await asyncio.wait_for(
+            _find_model_ownership(name, list_page), timeout=0.25
```
gesh: > [!NOTE]

**[router]** 🟡 MEDIUM

`resolve_model_ownership` caches into `data.tool_model_parameter_injected` only when the tool was actually found in a listing page (`ownership is not None`). A timeout, an exception, or a "tool never appears in any page" outcome is deliberately left uncached — the new tests `test_bounded_catalog_failures` and `test_slow_listing_does_not_block_dispatch` assert exactly that (`"echo" not in state.tool_model_parameter_injected`).

So a tool name that is permanently mistyped, permanently paginated past 16 pages, or whose `tools/list` handler is consistently slow or broken re-runs the full (up to 250 ms) discovery walk on **every single call** to that tool for the lifetime of the server instance — not just the first. For a moderately busy tool that is a persistent latency tax on the hot call-tool path rather than a one-time cost.

A short-TTL negative cache (even a few seconds) would bound this without meaningfully hurting the "tool just appeared" case the current design protects (`test_missing_tool_is_retried_when_it_appears`).

posthog/mcp/_instrument_lowlevel.py
```diff
@@ -171,6 +172,31 @@ async def handler(req: Any) -> Any:
...
+    server: Any, data: MCPAnalyticsData, req: Any, strip_injected: bool
+) -> Tuple[Any, bool]:
+    async def list_page(cursor: Optional[str]) -> Any:
```
gesh: > [!NOTE]

**[router]** 🟡 MEDIUM

Model-ownership discovery calls the real, un-instrumented `tools/list` handler (via `__posthog_mcp_original__`) out-of-band, **inside a `tools/call` dispatch**, wrapped in `asyncio.wait_for(timeout=0.25)`.

If the application's listing handler has side effects — DB queries, external auth or rate-limited calls, audit logging, a shared generator/cursor object — a timeout injects `CancelledError` into it mid-flight. The SDK cannot in general guarantee an arbitrary handler cleans up correctly from an arbitrary cancellation point. The existing test proves the *coroutine* gets cancelled, not that user handlers tolerate it.

This is inherent to the design and partly disclosed in the README, but it is a new code path that runs during ordinary tool dispatch rather than only at listing time. Worth an explicit call-out that a listing handler must be safely cancellable.

posthog/mcp/_instrument_v2.py
```diff
@@ -477,6 +478,40 @@ async def _standalone_injected_parameters(
     return frozenset(key for key in injected if not schema_has_param(schema, key))
 
 
+async def _prepare_raw_v2_arguments(
+    server: Any, data: MCPAnalyticsData, ctx: Any, params: Any
+) -> Tuple[Any, bool]:
+    async def list_page(cursor: Optional[str]) -> Any:
```
gesh: > [!NOTE]

**[router]** 🟢 LOW

Minor asymmetry between the two SDK-major adapters: the v1 raw low-level `list_page` (`posthog/mcp/_instrument_lowlevel.py:178-189`) forwards the originating call's `_meta` into the synthetic `tools/list` request (`_meta=req.params.meta`), but this v2 raw low-level `list_page` builds `mcp_types.PaginatedRequestParams(cursor=cursor)` with no `_meta` at all.

Probably harmless for a bounded internal discovery call, but a listing handler that inspects `_meta` (progress tokens, auth context threaded through metadata) would see different inputs depending on which SDK major is in use. Intentional?

posthog/test/mcp/test_defaults.py
```diff
@@ -0,0 +1,79 @@
...
+    await flush_background()
+    calls = events_named(client, "$mcp_tool_call")
+    assert len(calls) == 1
+    assert not calls[0]["properties"]["$mcp_is_error"]
+    assert calls[0]["properties"].get("$mcp_llm_model") == (
+        "model-a" if enabled else None
+    )
+    assert len(result.root.content) == (2 if enabled else 1)
+    assert len(events_named(client, "$mcp_tools_list")) == 0
+    assert request.params.arguments == {"msg": "ok", "llm_model": "model-a"}
```
gesh: > [!NOTE]

**[router]** 🟢 LOW

`test_v1_cold_capture_and_opt_out` asserts the *original* `request.params.arguments` object is left unmodified and that `$mcp_llm_model` is captured correctly, but never asserts what arguments are actually delivered to the real tool handler (`original(req)`) after `_prepare_model_arguments` runs.

Verified manually that on a cold low-level v1 server `llm_model` is correctly popped before the real handler sees it — this is new behaviour, since pre-PR the raw low-level path (`strip_injected=False`) never stripped `llm_model` at all. But nothing asserts it: the fixture's `echo` handler ignores unknown keys either way, so a regression that leaked `llm_model` through to a strict handler would pass. Consider asserting on the arguments dict the real handler receives.

gesh:
> [!NOTE]
>
> Multi-perspective review: router (cheap-first pass) + delegated reviewers (qa-team, paul-reviewer, xp-reviewer, security-audit as warranted)

## Verdict: ⚠️ REQUEST CHANGES <sub>(round 1 @ 80a8dd2)</sub>

No correctness, security, or data-loss bug was found. The code works. Both reviewers instead stopped at the same place: this change alters the wire-visible tool contract of every existing deployment on a routine `pip install --upgrade posthog`, and the release note does not say so.

### Key findings

**🟠 HIGH**
- **Is `minor` the right semver tier?** (`.sampo/changesets/…`) — With `capture_model=True` by default, the advertised `inputSchema` of every tool on the official high-level FastMCP / `MCPServer` adapters now lists `llm_model` in `required`. `enable_conversation_id=True` adds a handle to eligible tool responses. Both happen with zero code change by the user. Dispatch does **not** enforce `required`, so servers keep working — the risk is client-side (strict-schema clients, cached call templates). `references/public_api_snapshot.txt` also changes on four lines, which `CONTRIBUTING.md` says means the shape should be agreed on an issue first.
- **`capture_model` is one boolean doing two unrelated jobs** (`posthog/mcp/types.py:201`) — (a) read the model from recognised client `_meta`, which is invisible and changes no wire contract; (b) inject `llm_model` into the advertised schema as `required`, which very much does. Because it is one flag, the safe half cannot ship first. An `inject_argument` field on the already-existing `MCPAnalyticsModelOptions` would allow default-on observation now, and the schema flip after the data is in.

**🟡 MEDIUM**
- Changeset body names what was turned on, but not what the upgrader will see, nor the opt-out one-liner.
- README "Defaults and opt-outs" covers conversation correlation well, but never says `llm_model` joins `required` — the words a surprised operator would grep for.
- Discovery walk runs on the standalone FastMCP path even though a live tool registry is available there, unlike the equivalent v2 path (`_instrument_v2.py:253`).
- Negative lookups are never cached, so a permanently missing or slow-listing tool re-pays the full walk on every call, not just the first.
- Discovery calls the application's real `tools/list` handler out-of-band during `tools/call`, and a 250 ms timeout cancels it mid-flight. Arbitrary handlers may not be safely cancellable.

**🟢 LOW / ⚪ NIT** — `_meta` not forwarded on the v2 discovery path (v1 forwards it); no test asserts that `llm_model` is stripped before the real handler runs; no env-var escape hatch; two README copy nits.

### Convergence

One finding was raised independently by both reviewers: **the semver tier and the silent default flip** (`.sampo/changesets/mcp-analytics-defaults.md:2`). The router reached it from the empirical side (it installed mcp 1.30.0 and 2.2.0, instrumented live servers, and confirmed the `required` schema change and the response-payload change). Paul reached it from the rollout side. Convergent findings carry the highest confidence — treat this one as the review's main ask.

### Reviewer summaries

| Reviewer | Assessment |
| --- | --- |
| 🧭 router | Danger MEDIUM, confidence HIGH. Read every touched file with full context, installed both MCP SDK majors in real virtualenvs, and confirmed the pagination duck-typing, the argument stripping on cold low-level servers, and that the full suite plus the new tests pass under both majors. Found no correctness bug. Delegated one item: the rollout/semver judgment call. |
| 👤 paul-reviewer | Approves with caveats. Likes the direction — "you get the data without reading the docs" is usually right for an analytics SDK — and can live with `minor`, since `capture_model` is about a week old and `context` set the precedent. One real ask: split the `capture_model` boolean so the observing half can ship before the contract-changing half. |

> [!WARNING]
> `qa-team` was unavailable in this environment (skill not on disk). Its correctness, concurrency, and database lenses were not applied. `xp-reviewer` and `security-audit` were available but the router did not delegate to them.

---

posthog/mcp/_tool_schema.py
```diff
@@ -0,0 +1,68 @@
...
+    list_page: Callable[[Optional[str]], Awaitable[Any]],
+) -> bool:
+    if not is_capture_model_enabled(data.options.capture_model):
+        return False
+    if name in data.tool_model_parameter_injected:
+        return data.tool_model_parameter_injected[name]
+    try:
+        return await asyncio.wait_for(
+            _find_model_ownership(name, list_page), timeout=0.25
+        )
```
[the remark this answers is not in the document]
lucasheriques: Fixed for the earlier design in 80a8dd2, and since then the catalog lookup was removed entirely in ea63cf1. A cold raw instance now reads the self-reported model under the posthog-js ADR-0011 rule (reads fail open, strips fail closed), and nothing consults the listing on the call path.

.sampo/changesets/mcp-analytics-defaults.md
lucasheriques: Keeping `minor`: `context` shipped the same required-argument precedent as a minor, and dispatch never enforces `required`. On the CONTRIBUTING.md point: this repo has no "agree the shape on an issue first" rule; AGENTS.md only asks that the snapshot be regenerated (`make public_api_snapshot`), which it is. posthog-js does have that rule for new or changed option shapes, and this change alters no shape: same options, same types, different defaults. The decision and the rejected alternatives are now recorded in posthog-js ADR-0013, which both SDKs cite.

.sampo/changesets/mcp-analytics-defaults.md
lucasheriques: Done in ea63cf1: the changeset names the schema change, the response handle, and the one-line opt-out. Keeping one changeset, since both flags share the same opt-out line and ship as one behaviour change.

posthog/mcp/types.py
lucasheriques: Considered and rejected in posthog-js ADR-0013. Only Codex exposes model metadata today, so a default-on observe-only half yields near-zero data for every other harness and only delays what the flip exists to collect. The object form (`MCPAnalyticsModelOptions`) leaves room for an `inject=False` knob later without a breaking change, if the data says self-report is not earning its field.

posthog/mcp/types.py
lucasheriques: The README's Defaults section now says responses change and how to get the old shape back (ea63cf1). Not adding a first-injection log line: the module logs misconfiguration, not enabled defaults, and a per-instance log would fire on every request in the per-request topology this targets.

posthog/mcp/posthog_mcp.py
lucasheriques: Not taking this one: nothing in the package reads the environment, and the opt-out is a one-line option. Recorded as rejected in posthog-js ADR-0013.

posthog/mcp/README.md
lucasheriques: Done in ea63cf1: the Defaults section now says `llm_model` joins the schema (required on the official high-level adapters, optional elsewhere), that dispatch never enforces it, and names the opt-out.

posthog/mcp/README.md
lucasheriques: Done in ea63cf1.

posthog/mcp/README.md
lucasheriques: Done in ea63cf1: the redundant block is gone and the custom-dispatcher example no longer passes `capture_model=True`.

posthog/mcp/_instrument_lowlevel.py
lucasheriques: Agreed, and the walk is gone. Standalone FastMCP now reads ownership of all three injected keys from the registered tool schema (`_tool_owned_injected_keys`), never through a listing or middleware, so a cold replica strips `llm_model` and records it without any `tools/list`. The rule that makes this safe: FastMCP rejects undeclared arguments anyway, so stripping a key the tool does not declare can never hurt, while a declared key (top-level or inside a composed schema) always stays. A listing-based resolver was tried in between; six `codex review` rounds showed it diverging from the advertised listing across FastMCP 2.5-4.0, dereferencing, and middleware, which is why the schema won.

posthog/mcp/_tool_schema.py
lucasheriques: Moot: no lookup remains to cache (ea63cf1).

posthog/mcp/_instrument_lowlevel.py
lucasheriques: Moot: the SDK no longer calls the listing handler during `tools/call` (ea63cf1).

posthog/mcp/_instrument_v2.py
lucasheriques: Moot with the lookup removed. For the record it was not an asymmetry: SDK v2 handlers read request meta from `ctx`, and the call's `ctx` was passed through.

posthog/test/mcp/test_defaults.py
lucasheriques: Done in ea63cf1: the cold v1 test now records what the real handler received. The assertion is the inverse of the one suggested: a cold raw instance strips nothing, because it cannot prove it owns the argument and raw handlers ignore extra keys.

.sampo/changesets/mcp-analytics-defaults.md
lucasheriques: Verified the existing disposition against the current branch: the changeset remains minor, the README/changelog text describes the wire-visible defaults and opt-out, and the public API snapshot check passes. [ADR-0013](https://github.com/PostHog/posthog-js/blob/ee1941ec64c312f8f52728e25596e789670b4944/packages/mcp/docs/adr/0013-analytics-capture-is-on-by-default.md) records the decision and its compatibility tradeoffs.

The additional carried-session fix in https://github.com/PostHog/posthog-python/commit/47520743473079f753953149a58c83f5fab01f3a prevents a newly minted handle from replacing an existing transport/token session. The required-model-schema default remains intentional; this thread is being closed as the previously answered release-policy decision, not as removal of that contract change.

posthog/mcp/types.py
lucasheriques: Verified and retained the existing decision in [ADR-0013](https://github.com/PostHog/posthog-js/blob/ee1941ec64c312f8f52728e25596e789670b4944/packages/mcp/docs/adr/0013-analytics-capture-is-on-by-default.md): this release enables the existing model-capture option as a whole, with an explicit opt-out. Splitting metadata observation from argument injection is deferred, so no new option shape is introduced here.

The merged branch passes both MCP SDK major suites and the public API snapshot check. Closing the answered design thread on that basis; this does not claim the suggested split was implemented.

posthog/mcp/posthog_mcp.py
lucasheriques: Verified the existing scope decision in [ADR-0013](https://github.com/PostHog/posthog-js/blob/ee1941ec64c312f8f52728e25596e789670b4944/packages/mcp/docs/adr/0013-analytics-capture-is-on-by-default.md). This release keeps the documented `capture_model=False` / `enable_conversation_id=False` options as the opt-out and does not introduce an environment-variable override. Explicit opt-out regressions pass. Closing this answered, non-blocking suggestion.

DanielVisca (approved):

PR #947, opened 2026-09-14
fix: Return None for malformed feature flag payloads

marandaneto:
Malformed JSON payloads currently return the raw string through `evaluate_flags(...).get_flag_payload(...)`, while the deprecated `get_feature_flag_payload(...)` can raise `JSONDecodeError`.

Both APIs now use a shared parser that returns `None` for malformed or empty JSON strings, for local and server-provided payloads. Valid JSON and pre-decoded values are preserved.

- Reproduced the inconsistent behavior with failing regression tests before the fix.
- All 327 focused payload, evaluation, feature flag, and type tests pass.
- Ruff formatting and lint checks, the public API snapshot check, and import with warnings treated as errors pass.
- Autoreview of commit `3781203` against `origin/main` reported no actionable findings.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with Pi using file editing, shell tests, GitHub CLI, and the isolated autoreview helper. The requested behavior is to treat malformed JSON payloads as absent rather than return invalid strings or raise decoding errors. The change shares parsing between the newer and deprecated APIs and includes local and remote regression coverage. Human review is required.


marandaneto:
idk if this one should be fixed or if the other sdks should be fixed
a malformed json should return as is or `none`
cc @PostHog/team-feature-flags ?

posthog/types.py
```diff
@@ -4,6 +4,16 @@
 
 FlagValue = Union[bool, str]
 
+
+def _parse_flag_payload(raw_payload: Any) -> Optional[Any]:
```
JamesPatrickGill, line 13: **Blocking:**

_compute_payload_locally at returns the stored payload as a raw JSON string and is called in two places, on the single flag path and the bluk flag path. On the bulk path the raw string goes straight into the dict but on the single-flag path we go through this function and is decoded.

These two used to accidentally agree on what the result of "" would be. Removing the `and payload` guard removed the coincidence so this path returns None whilst the bulk return "" still. I think this is the failure the CI is showing

"" reading as absent makes sense, and we are tackling "malformed or empty" here. So we just need both paths to agree:

We can either:
* keep the parser as is and make the bulk drop the "" fields, or
* add back `if not raw_payload...` return statement.

Whichever we go down we need to make sure the "" edgecase is called out somewhere

posthog/types.py
```diff
@@ -4,6 +4,16 @@
...
+    if isinstance(raw_payload, str):
+        try:
+            return json.loads(raw_payload)
+        except json.JSONDecodeError:
+            return None
```
JamesPatrickGill, line 21: _Suggestion:_

Might be worth logging here, think it aligns with how we should be logging in parse failure territory.

posthog/types.py
```diff
@@ -231,9 +241,7 @@ def from_value_and_payload(
             key=key,
             enabled=enabled,
             variant=variant,
-            payload=json.loads(payload)
-            if isinstance(payload, str) and payload
-            else payload,
+            payload=_parse_flag_payload(payload),
```
JamesPatrickGill, line 256: **Blocking:**

This is wider than just payload getters, feature_enabled, get_feature_flag and get_feature_flag_result all reach it through _get_feature_flag_result. 

This is good but these other pathways remain untested despite behaviour changing. We should add that

JamesPatrickGill (commented):
Just a couple of small blocking points below. Looks good otherwise. 

To answer your question I think None is the correct direction in this regard. We already have the API rejecting malformed payloads and the raw string is just a bit ambiguous.

posthog/types.py
marandaneto: Fixed in eae7347. Added local/remote bulk cases to `test_flag_payload_parsing.py`: 22 failed before the fix because bulk returned serialized values unchanged, and all now pass. Bulk payloads are decoded at the public return boundary so internal values are not parsed twice; malformed, empty, and whitespace-only JSON returns None while valid JSON strings (including `""`), false, zero, and healthy sibling payloads are preserved. Updated the outdated empty-input expectations and called out the distinction in the changeset, consistent with https://github.com/PostHog/sdk-specs/pull/71. Full suite: 2,786 passed, 15 skipped.

posthog/types.py
marandaneto: Added in eae7347: parse failures now emit a warning through the PostHog logger, without logging payload contents or exception details. Three diagnostic tests failed before the change and pass now.

posthog/types.py
marandaneto: Added coverage in eae7347 for get_feature_flag_result, get_feature_flag, and feature_enabled across local and remote evaluation, true/false/multivariate results, and malformed/empty/whitespace-only payloads. The 18 cases verify that key, value, enabled state, variant, and reason are preserved while payload is None and no extra fallback request is introduced. These already passed on the prior implementation and are retained as regression coverage. Full suite: 2,786 passed, 15 skipped.

posthog/client.py
```diff
@@ -4410,12 +4406,17 @@ def get_all_flags_and_payloads(
...
                     device_id=device_id,
                 )
-                return to_flags_and_payloads(decide_response)
+                response = to_flags_and_payloads(decide_response)
             except Exception as e:
                 self.log.exception(
                     f"[FEATURE FLAGS] Unable to get feature flags and payloads: {e}"
                 )
 
+        payloads = response.get("featureFlagPayloads")
```
ioannisj: Heads up, the bulk getters (`get_all_flags_and_payloads`, `get_feature_flags_and_payloads`, `get_feature_payloads`) used to return payloads as raw JSON strings and now return them decoded. So anyone doing `json.loads(result["featureFlagPayloads"][key])` gets a `TypeError` after what looks like a patch bump.

Decoding is the right call, it matches `normalizeFlagsResponse` in posthog-js. Should the changeset be a minor, or at least call out that callers need to drop their own `json.loads`?

ioannisj (approved):

posthog/client.py
marandaneto: good point, returning decoded would be a breaking change, will return as it was and `none` if malformed only

marandaneto: Agreed: changing valid bulk payloads from serialized strings to decoded values would break existing callers, regardless of a minor bump. Fixed in 89fab0f by preserving valid serialized payloads in all three existing bulk getters and validating them only to reject malformed payloads. Existing json.loads() calls therefore keep working for valid payloads. Single-flag and evaluate_flags() snapshot payload getters retain their existing decoded-value behavior; changing legacy bulk return types is deferred to a major release. Added the compatibility rationale in code and corrected the changeset. The expanded compatibility tests failed before this change and all 224 focused cases now pass; the full suite passes with 4,095 tests. The merged bulk spec currently mandates decoded values, so it needs a follow-up compatibility allowance; this commit intentionally preserves the existing Python API rather than introducing that breaking change.

posthog/types.py
```diff
@@ -1,9 +1,27 @@
...
+
+
+def _parse_flag_payload(raw_payload: Any) -> Optional[Any]:
+    if isinstance(raw_payload, str):
+        try:
+            return json.loads(raw_payload, parse_constant=_reject_json_constant)
+        except (ValueError, RecursionError):
+            logging.getLogger("posthog").warning(
+                "[FEATURE FLAGS] Unable to parse flag payload as JSON"
+            )
```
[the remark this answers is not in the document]
marandaneto: Fixed in 84728a9 by lowering payload parse diagnostics to DEBUG. Repeated malformed, empty, whitespace-only, or non-JSON-constant inputs no longer emit warnings under the default logger configuration. Debug mode still provides diagnostics, including for empty inputs, to retain the merged spec’s parse-failure logging requirement without exposing payload contents. Twelve regression cases (decoded and legacy serialized modes) failed before the change and pass now; all 230 payload tests and the full suite of 4,101 tests pass.

PR #955, opened 2026-09-14
feat(traces): per-span limits and exception stacktraces

turnipdabeets:
Bounds what one span can hold, per the traces spec.

- At most `max_attributes_per_span` user attributes and `max_events_per_span` events (128 each, earliest-set wins); each event at most 128 attributes. Overflow is reported as `droppedAttributesCount` / `droppedEventsCount`, clamped to uint32. The `posthogDistinctId` and `sessionId` join keys are exempt.
- `max_attribute_value_length` (8192) bounds every string an attribute holds, nested ones included, plus span and event names, status messages and resource attributes, so one large value cannot get a span dropped as too large.
- Recorded exceptions now carry `exception.stacktrace`, keeping the tail of the traceback where Python puts the raising frame.

Not reachable from the client yet.

**Stack** (PR 7 of 9, based on `traces/06-export`):
1. #949 `traces/01-ids-traceparent`
2. #950 `traces/02-otlp-encoding`
3. #951 `traces/03-span-handles`
4. #952 `traces/04-transport`
5. #953 `traces/05-pipeline`
6. #954 `traces/06-export`
7. #955 `traces/07-span-limits` ← this PR
8. #956 `traces/08-before-span-send`
9. #957 `traces/09-client-wiring`

Unit tests in `posthog/test/tracing/test_limits.py`, plus additions to `test_span.py`, `test_otlp.py`, `test_pipeline.py` and `test_export.py`.

## :pencil: Checklist

### If releasing new changes

## 🤖 Agent context

Implemented with Claude Code (Claude Opus 5) against the traces spec, one commit per slice so each PR reviews on its own. Rebased onto main and opened as a stacked draft in a later Claude Code session (Claude Fable 5.1).

https://claude.ai/code/session_012o7CtHLfcypjmXL7g9ZGRC


posthog/tracing/_limits.py
```diff
@@ -0,0 +1,150 @@
...
+        key_str = attribute_key(key)
+        if not key_str:
+            continue
+        try:
+            item = _truncate(value[key], max_length, state, depth + 1)
+        except Exception:
+            item = UNSERIALIZABLE_VALUE
+        if item is not None:
+            emittable += 1
+        bounded[key] = item
```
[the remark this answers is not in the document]
turnipdabeets: Keeping as is: the spec bounds attribute values, not keys, and this SDK treats keys the same at every depth, so a top-level key is equally unbounded. Truncating only nested keys would make the two disagree. An oversized span is still caught: the transport measures the body against the 10 MiB limit, and the 413 path halves down to that one span and drops it with a reason.

posthog/tracing/_limits.py
```diff
@@ -0,0 +1,150 @@
...
+        keys = list(source.keys())
+    except Exception:
+        return {}, 0
+    attributes: Dict[str, Any] = {}
+    dropped = 0
+    for key in keys:
+        key_str = attribute_key(key)
+        if key_str is None:
+            continue
+        if len(attributes) >= max_count and key_str not in attributes:
```
dustinbyrne: Non-blocking: with 128 real event attributes followed by an optional `None`, this capacity check increments `droppedAttributesCount` before checking whether the value would be omitted. The 128 real attributes still export, but the loss count becomes 1 instead of 0.

Consider skipping null values before charging overflow, with coverage for nulls before and after the cap and a non-null overflow control.

posthog/tracing/_limits.py
```diff
@@ -0,0 +1,150 @@
...
+        if state.remaining_nodes <= 0:
+            return value
+        state.remaining_nodes -= 1
+        if isinstance(value, str):
+            return truncate_string(value, max_length)
+        if isinstance(value, (bool, int, float, date)):
+            return value
+        # The encoder stringifies anything else, so bound that text.
+        try:
+            return truncate_string(str(value), max_length)
```
dustinbyrne, line 65: Non-blocking: the OTLP encoder already represents callables with the stable `[Function]` marker, but this eager `str()` means a callable now reaches it as an address-bearing string instead. This affects span/event/resource values, including nested callables.

Consider preserving the bounded `[Function]` marker here and adding a span/event-to-encoder regression case alongside the existing direct encoder test.

dustinbyrne (approved):
Looks good overall. Two non-blocking inline notes on null-attribute drop counts and callable encoding; neither blocks this tracing slice.

Validation was static inspection and existing CI. The proposed regression cases were not executed.

Agent-led review, human-reviewed before posting.

posthog/tracing/_limits.py
turnipdabeets: Good catch, fixed in bbc0e2c. We now read the value and skip a None before checking the cap, so it never counts as a drop. Added cases for a None before the cap, past the cap, and a real value past the cap as the control. Node's boundAttributes has the same ordering, so we should mirror this there too.

posthog/tracing/_limits.py
turnipdabeets: Good catch, fixed in bbc0e2c. The walk now leaves a callable alone so the encoder emits [Function], matching Node. Added a span and event to encoder test with a nested callable alongside the direct encoder case.

posthog/tracing/_limits.py
```diff
@@ -0,0 +1,150 @@
...
+    finally:
+        state.ancestors.discard(marker)
+
+
+def _truncate_mapping(
+    value: Mapping, max_length: int, state: _WalkState, depth: int
+) -> dict:
+    bounded: dict = {}
+    emittable = 0
+    for key in list(value.keys()):
```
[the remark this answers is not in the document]
turnipdabeets: Keeping as is. The mapping already lives in the caller's memory, and the list holds one reference per key on top of that, so it cannot exhaust anything the caller has not exhausted already. The snapshot is deliberate: iterating lazily over a mapping another thread mutates raises mid-walk, and the walk then gives up and ships the value unbounded. The OTLP encoder takes the same snapshot, and Node's truncateValue uses Object.keys the same way.

jzhu13 (approved):
Reviewed against `traces/06-export`. Tests pass at the head. Limits are applied at write time, overwrites and `None` are accounted correctly, auto join keys are exempt, and the truncation walk mirrors the encoder's budgets in the safe direction. No blocking issues. Approving with non-blocking notes.

**Non-blocking**

1. `posthog/tracing/_span.py:281` an empty-string key spends a span (or event) attribute slot that the encoder then discards. Reproduced with `max_attributes=1`: `set_attribute("", 1).set_attribute("real", 2)` stores `{"": 1}`, counts one drop, and ships zero attributes. `_truncate_mapping` already skips falsy keys for nested values, so the top level is the odd one out. Treat `not key_str` like `None`.
2. `posthog/tracing/_span.py:190` tail truncation of a chained traceback keeps the outer wrapper and discards the root cause, since `format_exception` prints the cause first. Reproduced: `raise RuntimeError("wrapper") from ValueError("ROOT CAUSE")` at `max_length=120` keeps `wrapper`, not `ROOT CAUSE`. The test `test_bounds_the_stack_keeping_the_crash_site` actually asserts the wrapper survives while the recursive crash site is cut off, contradicting its name. Either format with `chain=False` and prepend the innermost cause's tail, or rename the test and document that the tail is the outermost exception.
3. `posthog/tracing/_limits.py:34` `truncate_attribute_value` allocates a `_WalkState` with a `set` and enters a `try` for every scalar. Measured: `set_attribute("k", "value")` spends 0.32 µs of 0.40 µs there versus 0.03 µs for a direct slice. A fast path for `str`, numbers, `bool`, and `None` before the walk removes it.
4. `posthog/tracing/_span.py:317` `add_event` with non-mapping `attributes` no longer logs. The base went through `copy_user_attributes`, which debug-logs "expected a mapping"; `bound_attributes` returns `({}, 0)` silently. Route through `copy_user_attributes` (which also removes the duplicated key walk in `bound_attributes`).
5. `posthog/tracing/_span.py:281` same-key check-then-write across threads can permanently consume a slot. Not reproduced in 300 trials, but the `_Activatable` docstring invites multi-thread use. Guarding with the existing lock is cheap.
6. Nits: two tests (`test_the_per_event_attribute_cap_is_opentelemetrys_default`, the `== 8192` literal in `test_config.py`) only fail on an intentional constant edit; `test_a_self_referencing_attribute_ends_the_span` asserts only the record count and would pass if the value were stored unbounded.

Two notes for #956: `apply_span_limits` inherits item 1 verbatim and re-walks every already-bounded value whenever a hook is configured, and #956 adds `SpanRecord.auto_attribute_keys` that this PR's `end()` does not populate, so the tip's `end()` must.

Reviewed with Claude Code (Claude Fable 5.1). Behaviors above were reproduced against this branch head.

turnipdabeets:
Thanks. Fixed in 27b9229:
- 1: an empty key spends no slot, at the span and in the post-hook re-cap.
- 2: test renamed and docstring says the outermost exception survives the cut.
- 3: scalars skip the walk.
- 4: event attributes go through the shared copier, so a non-mapping logs.
- 6: the self-reference test checks the stored marker.
- 5 (locking writes): still open.

turnipdabeets:
Follow-up, 1975532: item 5 is done. The cap check, the event count and the `end()` snapshot run under the handle's lock, with the value walk still outside it. Tests race eight threads on one key and on the event cap. Ready for another look.

jzhu13 (approved):

PR #966, opened 2026-09-16
fix(prompts): support unlabeled batch fetch in get_all

jurajmajerik:
## Problem

`get_all` requires a label, so there is no way to batch-fetch the latest version of every prompt. Passing `label=None` anyway urlencodes to the literal string `label=None`, and the server filters by a label named "None", returning an empty result. Reported by a customer.

## Changes

- `label` on `get_all` is now optional. Without it, the request omits the `label` param and returns the latest version of every prompt, matching plain `get(name)`.
- Unlabeled results are cached under the same keys plain `get(name)` reads, so those calls become cache hits.
- The label-resolution checks (absent/moved rows) only run when a label was requested.
- Updated the public API snapshot for the new signature.
- Added a sampo changeset (minor), so merging releases the change.

Added one test to `TestPromptsGetAll`: the unlabeled fetch must omit the `label` param from the URL, accept rows without labels, and seed the unlabeled `get()` cache. All 78 prompt tests pass locally.


marandaneto (approved):
The code looks good; no qualifying findings. Verdict: correct.

PR #967, opened 2026-09-17
feat(metrics): record requests and httpx durations as a histogram

pauldambra:
🤖 This description was written by an agent on Paul's behalf.

posthog-js can now record a duration histogram for every `fetch` and `XMLHttpRequest` the page makes (`metrics.network`, PostHog/posthog-js#4918). Python services have no equivalent, so a Django app that wants timings for the HTTP calls it makes has to wrap each call by hand.

This adds the same option to the Python SDK:

```python
posthog.setup(metrics={"network": True})
# or
Client("phc_...", metrics={"network": {"name": ..., "attributes": ...}})
```

When set, the SDK wraps `requests.Session.send` and, when httpx is installed, `httpx.Client.send` and `httpx.AsyncClient.send`. Every HTTP or HTTPS request is recorded as the `http.client.request.duration` histogram (unit `ms`) with `method`, `host`, templated `path` and `status_class` attributes. `name` accepts a string or a function of the request (return `None` to skip it). `attributes` accepts a function of the request and response whose result is merged over the defaults.

Design points, kept the same as posthog-js where they apply:

- The wrappers only observe. They call the original with the same arguments, return its result or re-raise its error, and a failure inside recording is logged once and never reaches the caller.
- The SDK marks its own `requests` sessions and its async httpx client, so PostHog's own uploads (including the metrics flush) are never recorded.
- `requests` sends each redirect hop through `Session.send` again. A context variable folds those into the outer request, so a redirected call is recorded once with its final status.
- The wrappers install when the client is built and are removed on `shutdown()`. If another wrapper was layered on top of ours and it cannot be spliced out, ours stays in place as a pass-through.
- `path` replaces all-digit and uuid-like segments with `:id`. Ids with a prefix or suffix (`order-123`, `38217.pdf`) are kept, so return your own `path` from `attributes` for those routes.

Nothing Django-specific is needed: the middleware already uses the global or configured client, so setting `metrics={"network": True}` on that client times every outbound call a view makes.

This is a public API addition (`metrics["network"]`). The shape copies the agreed posthog-js option rather than a new design, but per CONTRIBUTING it should still get a maintainer nod before merge.

- New `posthog/test/test_network_metrics.py` (37 tests, parametrized where the cases are a table): default attributes, status classes, failures recorded as `missing` and re-raised, redirects recorded once, non-HTTP schemes skipped, the SDK's own sessions skipped, httpx sync and async, `name` and `attributes` config, invalid config warning and falling back to defaults, install on construction and removal on shutdown, layered wrapper pass-through, module-level `posthog.setup()`.
- `uv run ruff format --check .`, `uv run ruff check .`, mypy with the baseline filter, `python -W error -c "import posthog"`, `make public_api_check`, and the full suite (2795 passed, 15 skipped).

## :pencil: Checklist

### If releasing new changes

- [x] Ran `sampo add` to generate a changeset file (added `.sampo/changesets/network-metrics.md` by hand, `minor`)

## 🤖 Agent context

Written with PostHog Desktop (Claude Code). Paul asked for the Python SDK equivalent of the posthog-js network metrics wrapper. Decisions: patch `Session.send` rather than `HTTPAdapter.send` so plain `requests.get()` and non-https mounts are covered; mark the SDK's own sessions instead of matching PostHog hosts by URL, as posthog-js ended up doing; use a context variable rather than a depth counter so the redirect guard works for both threads and asyncio tasks; keep the new module's helpers private so the only new public surface is the `network` option and the default metric name.

---


pauldambra:
🤖 Closed by an agent on Paul's behalf. We decided this duplicates what OpenTelemetry instrumentation already provides for server-side HTTP clients, and PostHog accepts OTLP metrics directly. Closing rather than merging.

PR #975, opened 2026-09-21
feat: send $release_id from POSTHOG_RELEASE_ID on every event

ablaszkiewicz:

A Python app cannot link its error tracking issues to a release. `posthog-cli` injects the release id into JavaScript bundles, and Python has no bundle.

- The SDK now reads `POSTHOG_RELEASE_ID` (the id that `posthog-cli release resolve` prints) and sends it as `$release_id` on every event.
- On `$exception` events the server resolves the release by a direct id lookup.
- Both clients read the variable once at init. The value is trimmed, and a blank value counts as unset.
- An explicit `$release_id` in event `properties` or in `super_properties` wins over the env var.
- Minimal `$feature_flag_called` events keep their cross-SDK property allowlist, so they do not carry the id.

- 25 new tests in `posthog/test/test_release_id.py`. The full suite passes (3467 passed, 16 skipped), as do ruff, mypy, and the public API check.
- End to end against a local PostHog stack: a wheel built from this branch, a release from a locally built `posthog-cli release resolve`, one handled exception, and one plain event. Both events carried `$release_id`, and the server resolved `python-release-env@1.0.0` from it. Not tested against production.

The issue page resolves the release that the CLI created:

python-release-issue

The event JSON shows the `$release_id` that the SDK sent and the `$exception_release` that the server resolved from it (same id):

python-release-json

## :pencil: Checklist

- [ ] I reviewed the submitted code.

### If releasing new changes

## 🤖 Agent context

Written with Claude Code (Claude Fable 5.1), directed by the PR assignee.

- The env var was chosen over a CLI-written file (fragile file discovery) and over git or version auto-detection (that gives a version, not a release id).
- The first iteration sent the id on `$exception` events only. The assignee then asked for all events.
- The changeset file was written by hand in the `.sampo/changesets` format, not with the `sampo add` command.


ablaszkiewicz:
## End-to-end test against a local PostHog stack

Tested with a new `python-release-env` app in `error-tracking-examples` (modelled on `rust-release-env`, not committed yet). Its `run` script does the full flow:

1. `uv build --wheel` on this branch, then installs **the built wheel** into a fresh venv (not an editable install, so the run also proves `posthog/release_id.py` ships in the package). The app printed `posthog 7.58.0` loaded from the venv's `site-packages`.
2. `posthog-cli release resolve --release-name python-release-env --release-version 1.0.0` (locally built CLI) created the release and printed its id.
3. The app started with that id in `POSTHOG_RELEASE_ID`, captured one handled `RuntimeError` through a three-module call chain, and captured one plain event.

Result, read back from the local instance:

| Event | `$release_id` | Server-resolved release |
|---|---|---|
| `$exception` | `01a0c42a-f8df-0000-5141-a9287ac358b0` | `python-release-env@1.0.0` (`$exception_release`, with git commit, branch, and repo) |
| `python-release-env started` | `01a0c42a-f8df-0000-5141-a9287ac358b0` | n/a (not an exception) |

So the server resolved the release by a direct id lookup from `$release_id`, and the plain event carries the id too.

**Issue page: the release chip resolves to the release the CLI created**

python-release-issue

**Event JSON: the raw `$release_id` the SDK sent, and the `$exception_release` the server resolved from it (same id)**

python-release-json

One UI note, unrelated to this PR: the curated Properties tab does not list `$release_id` (its filter reports "No matching properties"), so the raw value is only visible with the JSON toggle.

🤖 Test run and comment by Claude Code, directed by the PR assignee.

marandaneto (approved):

